Issue #116, opened 2019-03-29
Enumerate/Identify angles, dihedrals, etc


ahy3nz:
Thinking in a workflow manner (specifically for foyer), a user will probably start with an mb.Compound -> topology.Topology. Bonds will get transferred, but the resultant Topology will not be aware of angles, dihedrals, etc.

We need a function that will search sites and bonded neighbors to find angles and dihedrals. I'll note that we can try to be fancy with connected graphs, but that would require a good assessment of different graph packages. 

This is necessary for reworking `ff.apply`. I'll tackle this issue. I will probably be basing a lot of logic off [this code that uses omm objects](https://github.com/mosdef-hub/foyer/blob/8f3019db8554fbe176981cb6a26466424c29aa1d/foyer/forcefield.py#L616)

daico007:
solved by `identify_connection()` method

PR #119, opened 2019-03-29
Enumerate connectivity

ahy3nz:
#116 Inferring angles so far, dihedrals might get messier and might not be easily extended from the current infer_angles implementation. See the notes I wrote in `infer_angles`

Also updated adding sites from connections for n-membered Connections, not just 2-membered Connections


daico007:
Close by #356

PR #140, opened 2019-05-05
Use NetworkX to identify bonds, angles, dihedral, impropers

ahy3nz:
While messing around with using networkx to identify angles/dihedrals/impropers, some comparison functions are needed. This PR includes definitions of those comparison functions - I don't think these comparison funcs need to be super sophisticated, we just need to have some sort of way to compare and sort sites so the `nx.graph` can do something
I've attached an example jupyternotebook that does this. I think the idea of using some graph-matching techniques could be a robust way to identify angles/dihedrals/impropers, with some clear patterns that we can extend to other types of connections. As @volgy said, we may want to implement these graphs and matching algorithms in our own way #138 
[example.gz](https://github.com/mattwthompson/topology/files/3146081/example.gz)


uppittu11:
I done messed this up

PR #151, opened 2019-05-30
Hoomd metadata

ahy3nz:
First pass at printing to a hoomd metadata json (#151 ). Goal is to have this working for LJ systems

Some important features:
* The use of a dictionary of `top.PotentialTemplate : hoomd.class` that outlines the supported functions while also relating the top.Potential to a particular hoomd object in the metadata
* Checking for supported functions is written right inside the metadata-writer, but a very similar ideas might be needed for gmx (mapping top.Potential to gmx.functiontype) or lammps (mapping top.Potential to lmp keyword)
* To use the dictionary, I needed to override the `PotentialTemplate.__hash__` because it was inheriting from `Potential.__hash__` which will throw errors because PotentialTemplate lacks the information a Potential has
* Some crude warnings/exceptions are thrown for un-supported NB functions, or for NB functions that won't mix well (two atom types with different NB functions)
* Some basic combining rule (this just looks at `top.combining_rule` string to perform a function) - I'm wondering if, down the line, `top.combining_rule` shouldn't be a string, but rather a class that has functions like `top.combinig_rule.mix_sigma` or `top.combining_rule.mix_epsilon`
* Since HOOMD works in reduced units, printing sigma and epsilon is not so trivial - we need to take the `unyt.quantities`, identify their fundamental dimensions, and then reduce their units according to a unitsystem

### DIFF
--- a/topology/formats/gsd.py
+++ b/topology/formats/gsd.py
@@ -92,8 +92,8 @@ def write_gsd(top,
     #    _write_pair_information(gsd_file, top)
     if top.n_bonds > 0:
         _write_bond_information(gsd_file, top)
-    #if structure.angles:
-    #    _write_angle_information(gsd_file, top)
+    if top.n_angles > 0:
+        _write_angle_information(gsd_file, top)
     #if structure.rb_torsions:
     #    _write_dihedral_information(gsd_file, top)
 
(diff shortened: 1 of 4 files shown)


daico007:
replaced by https://github.com/mosdef-hub/gmso/pull/696

PR #197, opened 2020-01-24
Add virtual templates

rmatsum836:
Really primitive `VirtualSite` class.  My plan was originally to inherit from the `Connection` class.  However virtual sites are not technically connected to the other members, so I decided to create a new class.  I could be doing this wrong though so feel free to provide any criticism/ feedback.

In this class, I've separated out the attributes so that `virtual_site` and `virtual_site_members` are separate attributes.  The underlying `virtual_site_type` needs work but hoping we can discuss this sometime.

### DIFF
--- a/topology/core/virtualsite.py
+++ b/topology/core/virtualsite.py
@@ -1,5 +1,6 @@
 import warnings
 from topology.core.site import Site
+from topology.core.potential import Potential
 
 
 class VirtualSite(object):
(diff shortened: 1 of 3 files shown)


rmatsum836:
Based on recent discussion regarding the GMSO API, this PR will get refactored based on the following design decisions:
- `VirtualSite` will be a subclass of the `Site` abstract base class
- We will either have a site groups in the `VirtualSite` class itself or a `VirtualSitePosition` class that handles calculating the virtual site position from the surrounding sites.

chrisjonesBSU:
Closing as we are working on this in #876

PR #267, opened 2020-02-22
XML Writer For a GMSO ForceField

umesh-timalsina:
I know writing XMLs can sometimes be a pain, especially with a verbose and strict schema like ours. So, in an attempt to make life easier for everyone and especially to the user once this package gets public, I present an initial idea for writing XMLs. Since our schema is grouped in such a way that the `ForceField` XML will have `CoreTypes` grouped together by expression, This PR adds this feature in ForceField class. Apart from few quirks like string representation of a `unyt_quantity` for example, this should be good to go at this point if and only if accompanied by couple of good unit tests.


umesh-timalsina:
See #360

PR #306, opened 2020-03-02
TOP writer: "moleculename" tag doesn't correspond to the "molecule" tag 

uppittu11:
This is a hotfix to get GROMACS sims working with the current format (one big molecule). When/if we unique subtops/molecules set up this can be reverted. 

This change ensures that the name in "[moleculename]" is the same as the name in "[molecules]".

### DIFF
--- a/gmso/formats/top.py
+++ b/gmso/formats/top.py
@@ -100,7 +100,7 @@ def write_top(top, filename):
             out_file.write(
                 '{0}\t\t\t'
                 '{1}\n\n'.format(
-                    top.subtops[0].name,
+                    top.name,
                     3
                 )
             )


uppittu11:
Here a water box test that works with this feature branch. 

https://gist.github.com/uppittu11/c421280913edee7e3a56ce09a7d3c2d3

mattwthompson:
Is there any way this can be tested? Will the malformed TOP file be read in by i.e. ParmEd?

mattwthompson:
Just adding this block

```python3
import parmed as pmd

pmd.load_file('system.gro')
pmd.load_file('system.top')
```
to your notebook works on this patch but not master

PR #316, opened 2020-03-02
Update status checks

mattwthompson:
Just seeing which status checks trigger

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,12 +25,18 @@ jobs:
         Python37Ubuntu:
           imageName: 'ubuntu-latest'
           python.version: 3.7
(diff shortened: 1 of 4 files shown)


mattwthompson:
@mosdef-hub/mosdef-maintainers

* AZP is updated; I decided to make a new pipeline

* There is a duplicate pipeline in my name on Azure, I will delete this soon

* Codecov isn't listed among the status checks, maybe it will take a day or two to find it: https://github.com/mosdef-hub/gmso/settings/branch_protection_rules/14953364

* Do we want to add travis and Appveyor? My preference is nah

* We should test on other OS and Python versions

* Any apps I've missed installing?

PR #322, opened 2020-03-03
docs namechange topology to gmso

rwsmith7531:
Update url for pull requests, cloning url, and directory name for installation of gmso to reflect change from 'topology' to 'gmso'


mattwthompson:
Thanks @rwsmith7531!

PR #326, opened 2020-03-03
Update AZP to only run codecov on one build

mattwthompson:
I goofed in #316 and codecov was running on the entire matrix. This fixes that


azure-pipelines.yml
```diff
@@ -76,6 +76,7 @@ jobs:
           coverage xml
           codecov --file ./coverage.xml --token 4020cb2c-5787-485a-92f2-5d106dff85a1
           coverage html
+        condition: and( eq( variables['Agent.OS'], 'Linux' ), eq( variables['python.version'], '3.7' ) )
         displayName: Upload Coverage Report
       
       - task: PublishCodeCoverageResults@1
```
uppittu11: You can probably add the conditional statement to the "Publish Coverage Results" task too so that we don't get a codecov warning

image

PR #329, opened 2020-03-03
Update setup.py to fix packages, requirements, urls.

bdice:
This updates the `setup.py` script to use `packages=find_packages()`, which is necessary for installation on conda-forge.

I also included the requirements in `setup.py` and fixed outdated URLs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup
 
 #####################################
 VERSION = "0.0.0"
-ISRELEASED = False 
+ISRELEASED = False
 if ISRELEASED:
     __version__ = VERSION
 else:
@@ -10,13 +10,23 @@ else:
 #####################################
 
 
+requirements = [
+    'numpy',
+    'scipy',
+    'mbuild>=0.10.4',
+    'unyt>=2.4',
+    'boltons',
+    'lxml',
+]
+
+
 setup(
     name='gmso',
     version=__version__,
     author='Matthew W Thompson, Justin Gilmer',
     author_email='matt.thompson@vanderbilt.edu, justin.b.gilmer@vanderbilt.edu',
-    url='https://github.com/mattwthompson/topology',
-    download_url='https://github.com/mattwthompson/topology/tarball/{}'.format(__version__),
+    url='https://github.com/mosdef-hub/gmso',
+    download_url='https://github.com/mosdef-hub/gmso/tarball/{}'.format(__version__),
     package_dir={'gmso': 'gmso'},
     license="MIT",
     zip_safe=False,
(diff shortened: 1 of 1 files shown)


bdice:
@mattwthompson This PR will need to be merged and a release will need to be made before `gmso` can be put onto conda-forge. Ping me when you make a release and I will update https://github.com/conda-forge/staged-recipes/pull/10978/.

setup.py
```diff
@@ -1,25 +1,35 @@
...
 
+requirements = [
+    'numpy',
+    'scipy',
+    'mbuild>=0.10.4',
+    'unyt>=2.4',
+    'boltons',
+    'lxml',
+]
+
```
mattwthompson: General question - does listing this array replace the use of a `requirements.txt` file?

setup.py
```diff
@@ -1,25 +1,35 @@
...
+
 setup(
     name='gmso',
     version=__version__,
+    packages=find_packages(),
+    zip_safe=True,
     author='Matthew W Thompson, Justin Gilmer',
     author_email='matt.thompson@vanderbilt.edu, justin.b.gilmer@vanderbilt.edu',
```
mattwthompson: Should this author list be expanded to include the major contributors (I don't want to leave Umesh and others out) or should this be modified to be credited to the team? I also assume there is no copyright implication here

setup.py
```diff
@@ -1,25 +1,35 @@
-from setuptools import setup
+from setuptools import setup, find_packages
 
 #####################################
 VERSION = "0.0.0"
```
mattwthompson:
```suggestion
VERSION = "0.2.0"
```

setup.py
bdice: @mattwthompson I concur there is no copyright implication. You may be interested in the way we've handled this in signac, which attributes the developer group as a whole as well as the "original" author. This [pull request comment](https://github.com/glotzerlab/signac-flow/pull/210#issuecomment-568912170) explains what I did, and you can look at the [file changes](https://github.com/glotzerlab/signac-flow/pull/210/files#diff-2eeaed663bd0d25b7e608891384b7298R40-R43) in that PR as well.

setup.py
bdice: No, and I recommend the use of `install_requires` if you had to pick one. The reasons for using each are slightly different. The actual dependencies (things that should be installed when the package is installed) should go in `setup.py` in the `install_requires`. See these links for more info:
1. https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/
2. https://packaging.python.org/discussions/install-requires-vs-requirements/

PR #330, opened 2020-03-03
More tests for GROMACS topology writer

mattwthompson:
This is built off of #306, I tried to include the typed water box as a fixture so that we can run a round-trip with `ParmEd`, which handles the case that @uppittu11's PR fixed.

I could use some help turning `test_pmd_loop` into something that more elegantly loops over pytest fixtures, ideally with `pytest.mark.parametrize`.

I like the idea of using `ParmEd` as a dummy check for our writers as we build them up, and maybe remove this in the future if we are more confident in our glue.

### DIFF
--- a/gmso/tests/base_test.py
+++ b/gmso/tests/base_test.py
@@ -5,8 +5,9 @@ import unyt as u
 
 from gmso.core.box import Box
 from gmso.core.topology import Topology
-from gmso.core.element import Hydrogen
+from gmso.core.element import Hydrogen, Oxygen
 from gmso.core.site import Site
+from gmso.core.angle import Angle
 from gmso.core.atom_type import AtomType
 from gmso.core.forcefield import ForceField
 from gmso.external.convert_mbuild import from_mbuild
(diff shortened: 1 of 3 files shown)


mattwthompson:
Tests are failing by not finding some bond information

gmso/tests/base_test.py
```diff
@@ -97,8 +98,34 @@ def water_system(self):
...
+
+        ff = ForceField(get_path('tip3p.xml'))
+
+        element_map = {"O": "opls_111", "H": "opls_112"}
+
+        for atom in top.sites:
+            atom.atom_type = ff.atom_types[atom.name]
+
+        for bond in top.bonds:
+            bond.bond_type = ff.bond_types["opls_111~opls_112"]
```
uppittu11: It doesn't look like `Bond` doesn't have a `bondtype` attribute. It only has `connection_type`. 
```suggestion
            bond.connection_type = ff.bond_types["opls_111~opls_112"]
```

https://github.com/mosdef-hub/gmso/blob/aa1d2a43719dae6e1c1a21ec5c9e2b4a1a44ffbf/gmso/core/bond.py#L33

gmso/tests/test_top.py
```diff
@@ -15,10 +15,10 @@ def test_write_top(self, typed_ar_system):
         write_top(top, 'ar.top')
 
 
-    def test_pmd_loop(self, typed_ar_system):
-        top = typed_ar_system
-        write_top(top, 'ar.top')
-        pmd.load_file('ar.top')
+    def test_pmd_loop(self, typed_ar_system, typed_water_system):
```
rmatsum836: Would it be more appropriate to use here: 

```
pytest.mark.parametrize('top', [typed_ar_system, typed_water_system])
def test_pmd_loop(self, top):
```

rmatsum836: The way it would actually be done is:
```
@pytest.mark.parametrize('top', ['typed_ar_system',
typed_water_system'])
def test_pmd_loop(self, top, request):
    write_top(request.getfixturevalue(top), 'system.top')
    pmd.load_file('system.top')
```

I don't think this is actually cleaner, but it ensures that the water system and ar system get tested separately.

mattwthompson: Thank you for figuring out the syntax!

I disagree that it's not cleaner; I expect that this structure makes it easier to add other fixtures to this loop since we don't need to re-write the internals of the function

PR #338, opened 2020-03-06
Add improper and impropertype

rsdefever:
Add improper and impropertype. I used the dihedral and dihedraltype to template these files.

As discussed at (I think) all hands, I advocate to distinguish between dihedrals and impropers since the physical connectivity is different. A proper dihedral being `atom1-atom2-atom3-atom4`, whereas an improper is `atom1-atom2, atom1-atom3, atom1-atom4`.

Tests were again templated from `test_dihedral.py`

Let me know if I'm missing anything else in the code base.


gmso/core/improper.py
```diff
@@ -0,0 +1,55 @@
...
+                connection_type=connection_type, name=name)
+
+
+def _validate_four_partners(connection_members):
+    """Ensure 4 partners are involved in Improper"""
+    assert connection_members is not None, "connection_members is not given"
+    if len(connection_members) != 4:
+        raise GMSOError("Trying to create an Improper "
+                "with {} connection members". format(len(connection_members)))
+    return connection_members
```
uppittu11: Is there a way to validate that the 4 connection members form an improper instead of a dihedral? This would be done easier with a bond graph, but maybe we can stick a hacky improper validation + test for now

rsdefever: Yeah I was wondering that as well. Unless I'm mistaken, we don't have any analogous checks for angles or dihedrals at the moment. It would however seem sensible to enforce the sorts of connectivity requirements that you describe. I didn't add anything because I'm a bit disconnected from the core development effort here and wasn't sure what the thinking was on that front.

rsdefever:
> Looks good so far

Thanks! Am I missing else that needs to be changed in terms of classes/files?

gmso/core/improper.py
```diff
@@ -0,0 +1,55 @@
...
+from gmso.core.connection import Connection
+from gmso.core.improper_type import ImproperType
+from gmso.exceptions import GMSOError
+
+
+class Improper(Connection):
+    """A 4-partner connection between sites.
+
+    This is a subclass of the gmso.Connection superclass.
+    This class has strictly 3 members in its connection_members.
```
uppittu11:
```suggestion
    This class has strictly 4 members in its connection_members.
```

gmso/core/improper.py
```diff
@@ -0,0 +1,55 @@
...
+    connection_type : gmso.ImproperType, optional, default=None
+        ImproperType of this improper.
+    name : str, optional, default=Improper
+        Name of the improper.
+
+    Notes
+    -----
+    Inherits some methods from Connection:
+        __eq__, __repr__, _validate methods
+    Addiitonal _validate methods are presented
```
uppittu11:
```suggestion
    Additional _validate methods are presented
```

gmso/core/improper.py
uppittu11: This validation could be part of another PR. I reccomend raising an issue with a running list of things left to do for full integration of impropers (add them for the ff and xml, write impropers to the various file formats, validate improper members, etc.)

gmso/core/improper.py
```diff
@@ -0,0 +1,55 @@
...
+from gmso.core.improper_type import ImproperType
+from gmso.exceptions import GMSOError
+
+
+class Improper(Connection):
+    """A 4-partner connection between sites.
+
+    This is a subclass of the gmso.Connection superclass.
+    This class has strictly 4 members in its connection_members.
+    The connection_type in this class corresponds to gmso.ImproperType
```
mattwthompson: Might be worth defining what make a dihedral improper, for those with less domain knowledge. Maybe some ASCII art?

gmso/tests/test_topology.py
```diff
@@ -352,7 +354,29 @@ def test_dihedral_dihedraltype_update(self):
...
+        top.add_site(site1)
+        top.add_site(site2)
+        top.add_site(site3)
+        top.add_site(site4)
+        top.add_connection(improper)
+
+        assert top.n_impropers == 1
+        assert len(top.improper_types) == 1
+        assert len(top.improper_type_expressions) == 1
+        assert len(top.atom_type_expressions) == 2
```
mattwthompson: What would this return? Should they be 0s or 1s?
```python3
        assert top.n_dihedrals == 0
        assert len(top.dihedral_types) == 0
        assert len(top.dihedral_type_expressions) ==0
```

General question - should query-like properties lump impropers in with dihedrals? And does this mean we should have `.n_dihedral`, `.n_proper`, and `.n_improper`?

rsdefever: I would argue for a complete separation of proper and improper dihedrals. I am open to other ideas, but I lean towards calling proper dihedrals 'dihedrals' and improper dihedrals 'impropers'. This would be consistent with lammps, which has `dihedral_style` and `improper style`. It would be consistent with Cassandra, which has `dihedrals` and `impropers` as different sections of the MCF file. It would be inconsistent with GROMACS, which has `proper dihedrals` and `improper dihedrals`.

mattwthompson: This fluid representation is the concern I have. I agree that propers and impropers should be separated, I'm just not sure how to do the bookkeepping. Is "dihedral" an umbrella term for everything (propers and impropers) or does it imply propers only? It seems like GROMACS thinks about it as the latter, and LAMMPS & Cassandra the former. My current preference is for the GROMACS way of doing things, although I can be convinced otherwise (and internal consistency matters more than perfect coherence to other libraries)

rsdefever: Completely agree that consistency and clarity is most important. 

I can be talked into `.n_dihedral = .n_proper + .n_improper`. Curious what others think?

justinGilmer: > Completely agree that consistency and clarity is most important.
> 
> I can be talked into `.n_dihedral = .n_proper + .n_improper`. Curious what others think?

That seems like a decent way to handle it. Keep the proper and impropers separate like we expect, then can write them out however the writer needs it.

At least it fits my mental model of proper and improper dihedrals.

uppittu11: I also agree with that, but I think we should rename `dihedral` in another PR. This one is already pretty big

gmso/core/improper.py
ahy3nz: Not sure the best place to put it, but we worked on [detecting bond patterns in an earlier PR](https://github.com/mosdef-hub/gmso/pull/140/files)

PR #331, opened 2020-03-03
Write charges to LAMMPS

rmatsum836:
Addressing one of the many improvements from #320, actually writing out `site.charge` to LAMMPS data files.  I want to do some cleanups and wait for the Unyt unit system to get merged before this gets reviewed.

PR #339, opened 2020-03-06
Write out bonds, angles, and dihedrals to LAMMPS data file

rmatsum836:
Additions to LAMMPS writer so that bonds and angles can be written out.  In the process, I added and rearranged some variables to make this a little cleaner.  

There are several modifications I want to make to the LAMMPS writer, but will do so in separate subsequent PRs

### DIFF
--- a/gmso/formats/lammpsdata.py
+++ b/gmso/formats/lammpsdata.py
@@ -40,12 +40,6 @@ def write_lammpsdata(topology, filename, atom_style='full'):
 
     # TODO: Support various unit styles
 
-    types = list()
-    idx_dict = dict()
-    for idx, site in enumerate(topology.sites):
-        types.append(site.atom_type.name)
-        idx_dict[site] = idx
-
     forcefield = True
     # Not sure if this is the best way to check if a FF exists
     if topology.sites[0].atom_type.name in ['', None]:
@@ -53,6 +47,7 @@ def write_lammpsdata(topology, filename, atom_style='full'):
 
     box = topology.box
 
+    types = [site.atom_type.name for site in topology.sites]
     unique_types = list(set(types))
     unique_types.sort(key=natural_sort)
 
@@ -65,7 +60,7 @@ def write_lammpsdata(topology, filename, atom_style='full'):
         data.write('{:d} atoms\n'.format(len(topology.sites)))
         if atom_style in ['full', 'molecular']:
             if len(topology.bonds) != 0:
-                data.write('{:} bonds\n'.format(len(topology.bonds)))
+                data.write('{:d} bonds\n'.format(len(topology.bonds)))
             else:
                 data.write('0 bonds\n')
             if len(topology.angles) != 0:
@@ -208,7 +203,8 @@ def write_lammpsdata(topology, filename, atom_style='full'):
 
         for i, site in enumerate(topology.sites):
             data.write(atom_line.format(
-                index=idx_dict[site]+1,type_index=unique_types.index(types[i])+1,
+                index=topology.sites.index(site),
+                type_index=unique_types.index(types[i])+1,
                 zero=0,charge=0, # TODO: handle charges from atomtype and/or site
                 x=site.position[0].in_units(u.angstrom).value,
                 y=site.position[1].in_units(u.angstrom).value,
@@ -220,8 +216,8 @@ def write_lammpsdata(topology, filename, atom_style='full'):
                 data.write('{:d}\t{:d}\t{:d}\t{:d}\n'.format(
                 i+1,
                 bond_dict[bond.connection_type],
-                idx_dict[bond.connection_members[0]],
-                idx_dict[bond.connection_members[1]]
+                topology.sites.index(bond.connection_members[0]),
+                topology.sites.index(bond.connection_members[1])
                 ))
 
         if topology.angles:
@@ -230,9 +226,9 @@ def write_lammpsdata(topology, filename, atom_style='full'):
                 data.write('{:d}\t{:d}\t{:d}\t{:d}\t{:d}\n'.format(
                 i+1,
                 angle_dict[angle.connection_type],
-                idx_dict[angle.connection_members[0]],
-                idx_dict[angle.connection_members[1]],
-                idx_dict[angle.connection_members[2]]
+                topology.sites.index(angle.connection_members[0]),
+                topology.sites.index(angle.connection_members[1]),
+                topology.sites.index(angle.connection_members[2])
                 ))
 
         if topology.dihedrals:
@@ -241,8 +237,8 @@ def write_lammpsdata(topology, filename, atom_style='full'):
                 data.write('{:d}\t{:d}\t{:d}\t{:d}\t{:d}\t{:d}\n'.format(
                 i+1,
                 dihedral_dict[dihedral.connection_type],
-                idx_dict[dihedral.connection_members[0]],
-                idx_dict[dihedral.connection_members[1]],
-                idx_dict[dihedral.connection_members[2]],
-                idx_dict[dihedral.connection_members[3]]
+                topology.sites.index(dihedral.connection_members[0]),
+                topology.sites.index(dihedral.connection_members[1]),
+                topology.sites.index(dihedral.connection_members[2]),
+                topology.sites.index(dihedral.connection_members[3])
                 ))
--- a/gmso/formats/lammpsdata.py
+++ b/gmso/formats/lammpsdata.py
@@ -62,7 +62,10 @@ def write_lammpsdata(topology, filename, atom_style='full'):
             else:
                 data.write('0 dihedrals\n')
 
-        data.write('{:d} atom types\n'.format(len(topology.atom_types)))
+        data.write('\n{:d} atom types\n'.format(len(topology.atom_types)))
+        data.write('{:d} bond types\n'.format(len(topology.bond_types)))
+        data.write('{:d} angle types\n'.format(len(topology.angle_types)))
+        data.write('{:d} dihedral types\n'.format(len(topology.dihedral_types)))
 
         data.write('\n')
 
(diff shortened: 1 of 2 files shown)


rmatsum836:
I think this is ready to be looked at.

gmso/formats/lammpsdata.py
```diff
@@ -71,12 +68,12 @@ def write_lammpsdata(topology, filename, atom_style='full'):
             str(datetime.datetime.now())))
         data.write('{:d} atoms\n'.format(len(topology.sites)))
         if atom_style in ['full', 'molecular']:
-            if bonds != 0:
-                data.write('{:} bonds\n'.format(len(bonds)))
+            if len(topology.bonds) != 0:
+                data.write('{:} bonds\n'.format(len(topology.bonds)))
```
mattwthompson: I'm just curious what purpose the colon serves here? I am unfamiliar with this syntax

rmatsum836: I think that's a typo.

rmatsum836:
I think this is ready for review.  I now see the beauty of having our parameters in indexed sets, as it takes care of all the indexing for us.  As a result, I removed all of the lists and dictionaries that was attempting to do so.

gmso/formats/lammpsdata.py
```diff
@@ -38,55 +37,37 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
     with open(filename, 'w') as data:
         data.write('{} written by topology at {}\n\n'.format(
             topology.name if topology.name is not None else '',
             str(datetime.datetime.now())))
         data.write('{:d} atoms\n'.format(len(topology.sites)))
         if atom_style in ['full', 'molecular']:
-            if bonds != 0:
-                data.write('{:} bonds\n'.format(len(bonds)))
+            if len(topology.bonds) != 0:
+                data.write('{:d} bonds\n'.format(len(topology.bonds)))
```
justinGilmer: So ive not used/encountered  [the format string syntax much.](https://docs.python.org/3.6/library/string.html#format-string-syntax)

But for those who were curious
The colon is used when replacing something in a format string, the colon precedes the format specification for that replacement. See here: [https://docs.python.org/3.6/library/string.html#format-specification-mini-language](https://docs.python.org/3.6/library/string.html#format-specification-mini-language) for more info on the possible specifications that can be used. In the case above ``d`` represents
>

 \'d\' | Decimal Integer. Outputs the number in base 10.
-- | --

Just put here since I went on a bit of a journey learning about this.

gmso/tests/test_lammps.py
```diff
@@ -12,3 +14,33 @@ def test_write_lammps_triclinic(self, topology_site):
         top = topology_site()
         top.box = Box(lengths=[1,1,1], angles=[60,90,120])
         write_lammpsdata(top, filename='data.triclinic')
+
+    def test_water_lammps(self, water_system):
```
justinGilmer: I wish there was a nice way to compare these files when we write out, but with anything that deals with floating point math, its going to be a bit difficult to validate. 

I think we should consider/discuss other ways we can further validate our writers. But for now, I am happy with this.

gmso/formats/lammpsdata.py
```diff
@@ -38,55 +37,37 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
-    xyz = list()
-    types = list()
-    for site in topology.sites:
-        xyz.append([site.position[0],site.position[1],site.position[2]])
-        types.append(site.atom_type.name)
+    # TODO: Support various unit styles
 
     forcefield = True
+    # Not sure if this is the best way to check if a FF exists
     if topology.sites[0].atom_type.name in ['', None]:
```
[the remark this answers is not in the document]
mattwthompson: I though we already implemented (or at least intended to) a `top.is_parametrized` sort of check?

gmso/formats/lammpsdata.py
mattwthompson: I was mostly confused about it when it was just `{:}`, since I don't think that does anything

```python3
>>> "x {} z".format('y')
'x y z'
>>> "x {:} z".format('y')
'x y z'
>>> "x {:s} z".format('y')
'x y z'

File: gmso/tests/test_lammps.py (hunk 3 above)
Comment (rmatsum836) [reply gmso/tests/test_lammps.py]: Yeah I agree.  Once the reader gets merged, I plan on using it to validate the actual contents of these files.

File: gmso/formats/lammpsdata.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (rmatsum836) [reply gmso/formats/lammpsdata.py]: You're right. I updated the writer to check this through `topology.is_typed`.

File: gmso/formats/lammpsdata.py [hunk 5]
```diff
@@ -38,55 +37,32 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
         if atom_style in ['full', 'molecular']:
-            if bonds != 0:
-                data.write('{:} bonds\n'.format(len(bonds)))
+            if len(topology.bonds) != 0:
+                data.write('{:d} bonds\n'.format(len(topology.bonds)))
             else:
                 data.write('0 bonds\n')
-            if angles != 0:
-                data.write('{:} angles\n'.format(len(angles)))
+            if len(topology.angles) != 0:
```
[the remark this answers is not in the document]
rmatsum836: I prefer using `topology.n_angles` so I'll make these changes.

gmso/formats/lammpsdata.py
```diff
@@ -180,22 +177,44 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+                z=site.position[2].in_units(u.angstrom).value))
+
+        if topology.bonds:
+            data.write('\nBonds\n\n')
+            for i, bond in enumerate(topology.bonds):
+                data.write('{:d}\t{:d}\t{:d}\t{:d}\n'.format(
+                i+1,
+                topology.bond_types.index(bond.connection_type)+1,
+                topology.sites.index(bond.connection_members[0]),
+                topology.sites.index(bond.connection_members[1])
```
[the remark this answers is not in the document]
rmatsum836: Good catch

gmso/formats/lammpsdata.py
```diff
@@ -140,34 +116,55 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
                 xy.value, xz.value, yz.value))
 
-        # Mass data
-        masses = [site.atom_type.mass for site in topology.sites]
-        mass_dict = dict([(unique_types.index(atom_type)+1,mass) for atom_type,mass in zip(types,masses)])
-
+        # Write out mass data
         data.write('\nMasses\n\n')
-        for atom_type,mass in mass_dict.items():
+        for atom_type in topology.atom_types:
```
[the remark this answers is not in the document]
rmatsum836: You're right, I will moved this block of code so it's only run if the system is typed.

rmatsum836:
Just need to fix the dihedral issue (dividing by 2) and then all comments should be addressed.

rmatsum836:
Fixed the dihedrals.  We will eventually need to support more dihedral types but I think that can be addressed in a separate PR.

justinGilmer:
Works for me, once the tests pass, i think its ready to merge.

rmatsum836:
I have one more fix before getting merged.

rmatsum836:
Tested this with the argon example and it works.  Initially, the water system wasn't working because I forgot to add the number of bonds, angles, and dihedrals information.  After fixing this, the water system is correctly loaded into lammps.

gmso/tests/test_lammps.py
mattwthompson: Could this test be re-written to use the `typed_water_system` fixture?

rmatsum836:
Updated the water tests, tests are passing, and a water system is running in lammps for me.

mattwthompson:
Could you share the LAMMPS script you used @rmatsum836?

rmatsum836:
> Could you share the LAMMPS script you used @rmatsum836?

Not necessarily accurate (due to lack of charges which is addressed in #331) but it runs: https://gist.github.com/rmatsum836/77351efb21661c525ee07034940b722e

mattwthompson:
image

Without charges, this simulation produces these suspiciously-CO2-looking water molecules. I'm not expecting a perfect simulation at this stage, but I checked just the ~force constant for the~ angle and it doesn't seem to match up:

in the XML:
```
  <AngleTypes expression="0.5 * k * (theta - theta_eq)**2">
    <ParametersUnitDef parameter="k" unit="kJ/mol/degree**2"/>
    <ParametersUnitDef parameter="theta_eq" unit="degree"/>
    <AngleType name="AngleType-Harmonic-1" type1="opls_116" type2="opls_117" type3="opls_117">
      <Parameters>
        <Parameter name="k" value="383.0"/>
        <Parameter name="theta_eq" value="1.91061193"/>
      </Parameters>
    </AngleType>
  </AngleTypes>
```

and in my data file:
```
Bond Coeffs

1	412.28489	1.00000

Angle Coeffs

1	45.76960	1.91061
```

My back-of-the-envelope calculation says the force constant is correct. The issue seems to be that LAMMPS wants energy/rad^2 for the force constant but degrees for the angle, so a reference of 1.9 degrees is going to prefer these linear H-O-H waters. 

https://lammps.sandia.gov/doc/angle_harmonic.html

However, after manually changing this value in the data file, the simulation spams NaNs. I've fiddled with it some, and I'm less worried about that than writing out the equilibrium angle with the right units.

rmatsum836:
>The issue seems to be that LAMMPS wants energy/rad^2 for the force constant but degrees for the angle, so a reference of 1.9 degrees is going to prefer these linear H-O-H waters

So the force constant is correct but the angle needs to be converted from radians to degrees?

mattwthompson:
Yes, and I can offer nothing but "its what LAMMPS does"

> The following coefficients must be defined for each angle type via the angle_coeff command as in the example above, or in the data file or restart files read by the read_data or read_restart commands:
K (energy/radian^2)
Theta (degrees)
Theta is specified in degrees, but LAMMPS converts it to radians internally; hence the units of K are in energy/radian^2.

PR #340, opened 2020-03-06
MCF writer

rsdefever:
MCF writer for GMSO. Currently supports LJ and Mie potentials for nonbonded functional forms. The section writing impropers is commented out until those Classes are merged (#338). Right now tests only cover nonbonded section. 

Need to add tests for bonded sections. Opening PR for preliminary feedback.

### DIFF
--- a/gmso/formats/mcf.py
+++ b/gmso/formats/mcf.py
@@ -1,0 +1,601 @@
+from __future__ import division
+
+import warnings
+import numpy as np
+import datetime
+import networkx as nx
+
+import unyt as u
+
+from topology.core.topology import Topology
+from topology.lib.potential_templates import LennardJonesPotential
+from topology.lib.potential_templates import MiePotential
+from topology.lib.potential_templates import HarmonicAnglePotential
+from topology.lib.potential_templates import HarmonicTorsionPotential
+from topology.lib.potential_templates import PeriodicTorsionPotential
+from topology.lib.potential_templates import RyckaertBellemansTorsionPotential
+from topology.lib.potential_templates import HarmonicImproperPotential
+from topology.utils.compatibility import check_compatability
+from topology.exceptions import TopologyError
+
+from math import sqrt
+
+__all__ = ['write_mcf']
+
+def write_mcf(top, filename):
+    """Generate a Cassandra MCF from a topology.core.Topology object.
+
+    Parameters
+    ----------
+    top : topology.core.Topology
+        Topology object
+    filename : str
+        Path of the output file
+
+    Notes
+    -----
+    See https://cassandra.nd.edu/index.php/documentation for
+    a complete description of the MCF format.
+
+    """
+
+    _validate_compatibility(top)
+
+    # Identify atoms in rings and Cassandra 'fragments'
+    in_ring,frag_list,frag_conn = _id_rings_fragments(top)
+
+    # TODO: Figure out how to infer/get 14 scaling
+    # For now setting LJ/Coul 14 to 0.0
+    lj14 = 0.0
+    coul14 = 0.0
+
+    # TODO: What oh what to do about subtops?
+    # Refuse topologies with subtops as MCF writer is for single molecules?
+    # Or write single MCF for each unique molecule?
+
+    # Now we write the MCF file
+    with open(filename, 'w') as mcf:
+
+        header = ( '!***************************************'
+                   '****************************************\n'
+                   '!Molecular connectivity file\n'
+                   '!***************************************'
+                   '****************************************\n'
+                   '!File {} written by topology at {}\n\n'.format(
+                   top.name if top.name is not None else '',
+                   str(datetime.datetime.now()))
+                 )
+
+        mcf.write(header)
+        _write_atom_information(mcf, top, in_ring)
+        _write_bond_information(mcf, top)
+        _write_angle_information(mcf, top)
+        _write_dihedral_information(mcf, top)
+        _write_improper_information(mcf, top)
+        _write_fragment_information(mcf, top, frag_list, frag_conn)
+        _write_intrascaling_information(mcf, lj14, coul14)
+
+        # That's all, folks!
+        mcf.write('\n\nEND\n')
+
+
+def _id_rings_fragments(top):
+    """Identifies the rings and fragments of the molecule
+
+    Parameters
+    ----------
+    top : topology.core.Topology
+        Topology object
+
+    Returns
+    ---------
+    in_ring : list
+        True for each atom in a ring
+    frag_list : list
+        Atom ids belonging to each fragment
+    frag_conn : list
+        Fragment ids of connected fragments
+
+    """
+
+    # Identify atoms in rings
+    bond_graph = nx.Graph()
+    bond_graph.add_edges_from([ [bond.atom1.idx, bond.atom2.idx]
+                                 for bond in top.bonds ])
+    if len(top.bonds) == 0:
+        warnings.warn("No bonds found. Cassandra will interpet "
+                    "this as a rigid species")
+        in_ring = [ False ] * len(top.sites)
+        frag_list = []
+        frag_conn = []
+        return in_ring, frag_list, frag_conn
+
+    # Check if entire molecule is connected. Warn if not.
+    if nx.is_connected(bond_graph) == False:
+        raise ValueError("Not all components of the molecule are connected. "
+                         "MCF files are for a single molecule and thus "
+                         "everything should be connected through bonds.")
+
+    all_rings = nx.cycle_basis(bond_graph)
+    in_ring = [False]*bond_graph.number_of_nodes()
+    adj_to_ring = [False]*bond_graph.number_of_nodes()
+    for ring in all_rings:
+        for idx in ring:
+            in_ring[idx] = True
+
+    # Identify fragments
+    # See Shah and Maginn, JCP, 135, 134121, 2011, doi:10.1063/1.3644939
+    frag_list = []
+    frag_conn = []
+
+    # First create a neighbor list for each atom
+    neigh_dict = {i:list(bond_graph.neighbors(i))
+                  for i in range(bond_graph.number_of_nodes())}
+    # First ID fused rings
+    fused_rings = []
+    rings_to_remove = []
+    for i in range(len(all_rings)):
+        ring1 = all_rings[i]
+        for j in range(i+1, len(all_rings)):
+            ring2 = all_rings[j]
+            shared_atoms = list(set(ring1) & set(ring2))
+            if len(shared_atoms) == 2:
+                fused_rings.append(list(set(ring1+ring2)))
+                rings_to_remove.append(ring1)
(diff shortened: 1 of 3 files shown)


gmso/tests/test_mcf.py
```diff
@@ -0,0 +1,157 @@
...
+from gmso.formats.mcf import write_mcf
+from gmso.tests.base_test import BaseTest
+from gmso.tests.utils import get_path
+from gmso.utils.io import get_fn
+from gmso.exceptions import EngineIncompatibilityError
+from gmso.external.convert_mbuild import from_mbuild
+
+
+class TestMCF(BaseTest):
+    def test_write_lj_mcf(self):
```
rmatsum836: Instead of creating a argon `Topology`, you can add the `typed_ar_system` fixture as an argument to this test to save yourself some lines of code.  Such as:

```
def test_write_lj_mcf(self, typed_ar_system):
    top = typed_ar_system
    write_mcf(top, 'ar.mcf')
```

gmso/tests/test_mcf.py
```diff
@@ -0,0 +1,157 @@
...
+            .in_units(u.K)
+            .value,
+        )
+        assert np.isclose(
+            float(mcf_data[atom_section_start + 2][7]),
+            ff.atom_types["Ar"].parameters["sigma"].in_units(u.Angstrom).value,
+        )
+
+    def test_write_mie_mcf(self):
+        top = from_mbuild(mbuild.Compound(name="Xe"))
```
rmatsum836: Similar to my comment above but you can create a pytest fixture in `base_test.py` to test the mid potentials.

gmso/tests/test_mcf.py
```diff
@@ -0,0 +1,157 @@
...
+        top.update_topology()
+
+        write_mcf(top, "ar.mcf")
+
+        mcf_data = []
+        with open("ar.mcf") as f:
+            for line in f:
+                mcf_data.append(line.strip().split())
+
+        for idx, line in enumerate(mcf_data):
```
rmatsum836: I think it would also be good to add two simpler tests that only check that mie and LJ systems are written out correctly (So the test would end with `write_mcf()`.  

Are there plans to add a MCF reader as well?  I'm not super familiar with this file format so I don't know if there is value in that.  But I figure the tests that check that the information was written out correctly will be refactored if a reader is added as well.

rsdefever: Thanks for the comments -- I don't personally see a major need for an MCF reader. I'm open to adding it if that is a desired feature. In that case, yes, I would refactor the tests.

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+            if len(shared_atoms) == 2:
+                fused_rings.append(list(set(ring1 + ring2)))
+                rings_to_remove.append(ring1)
+                rings_to_remove.append(ring2)
+    for ring in rings_to_remove:
+        all_rings.remove(ring)
+    all_rings = all_rings + fused_rings
+    # ID fragments which contain a ring
+    for ring in all_rings:
+        adjacentatoms = []
```
justinGilmer:
```suggestion
        adjacent_atoms = []
```

Makes it a tad easier to read.

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+
+
+def _check_compatibility(top):
+    """Check Topology object for compatibility with Cassandra MCF format"""
+
+    if not isinstance(top, Topology):
+        raise GMSOError("MCF writer requires a Topology object.")
+    if not all([site.atom_type.name for site in top.sites]):
+        raise GMSOError(
+            "MCF writing not supported without " "parameterized forcefield."
```
justinGilmer: also weird formatting, unless I am missing something

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
+from __future__ import division
+
```
mattwthompson: Unless you have a good reason to include these 2/3 compatibilities, my preference is not include them at all

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+from gmso.core.topology import Topology
+from gmso.lib.potential_templates import LennardJonesPotential
+from gmso.lib.potential_templates import MiePotential
+from gmso.lib.potential_templates import HarmonicAnglePotential
+from gmso.lib.potential_templates import HarmonicTorsionPotential
+from gmso.lib.potential_templates import PeriodicTorsionPotential
+from gmso.lib.potential_templates import OPLSTorsionPotential
+from gmso.lib.potential_templates import RyckaertBellemansTorsionPotential
+from gmso.lib.potential_templates import HarmonicImproperPotential
```
mattwthompson: This criticism is not at all directed at you, I just want to note it somewhere - long lists of imports is a hint that we should think about doing this in a different way

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+
+    # Identify atoms in rings and Cassandra 'fragments'
+    in_ring, frag_list, frag_conn = _id_rings_fragments(top)
+
+    # TODO: Figure out how to infer/get 14 scaling
+    # For now setting LJ/Coul 14 to 0.0
+    lj14 = 0.0
+    coul14 = 0.0
+
+    # TODO: What oh what to do about subtops?
```
mattwthompson: 😅

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+    names = [site.name for site in top.sites]
+    types = [site.atom_type.name for site in top.sites]
+
+    # Check constraints on atom type length and element name length
+    # TODO: Update these following Cassandra release
+    # to be more reasonable values
+    n_unique_names = len(set(names))
+    for name in names:
+        if len(name) > 2:
+            warnings.warn(
```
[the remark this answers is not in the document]
rsdefever: Either is fine. I'm not going to enforce anything.

gmso/formats/mcf.py
```diff
@@ -0,0 +1,700 @@
...
+            mcf.write("1\n")
+            mcf.write("1 2 1 2\n")
+        else:
+            warnings.warn(
+                "More than two atoms present but " "no fragments identified."
+            )
+            mcf.write("0\n")
+    else:
+        mcf.write("{:d}\n".format(len(frag_list)))
+        for i, frag in enumerate(frag_list):
```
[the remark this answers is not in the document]
rsdefever: There I actually think the i, j=i+1 loop is a bit more clear (though yes, less pythonic).

gmso/formats/mcf.py
rsdefever: Yes, seems to have come from `black`.

rsdefever:
As of now is there a way to know what the LJ 1-4 scaling should be from the `Topology` object?

mattwthompson:
It's loosely stored as metadata in the force field class at the moment

In the class:
https://github.com/mosdef-hub/gmso/blob/515cbcf302d4d1c0d1d6c7d183dfc0cebbeeb025/gmso/core/forcefield.py#L170

In the schema:
https://github.com/mosdef-hub/gmso/blob/0515c665478150bb9f58a0e83f81af28868bd0c0/gmso/utils/schema/ff-gmso.xsd#L141-L151

And a unit test that should be instructive:
https://github.com/mosdef-hub/gmso/blob/0515c665478150bb9f58a0e83f81af28868bd0c0/gmso/tests/test_forcefield_xml.py#L24-L26

rsdefever:
> It's loosely stored as metadata in the force field class at the moment

Unless I'm missing something that doesn't let you get to it from the `Topology` class though, right? Like in my MCF writer where we are just provided a `gmso.Topology` object.

mattwthompson:
You're right. At different points in time we've passed around the idea of having a `top.force_field` attribute that just points to a `ForceField` object. It doesn't seem like this is implemented, and much of the important data is stored by parsing through the type objects directly, not by passing through a force field class. The two ideas I can come up with are

1. Add a  `top.force_field` attribute that behaves as a pointer, so we could do things like `top.force_field.scaling_factors['whatever']`. This assumes, for now, that there is only one force field per topology, and also that the data contained in each should match.

2. Add the scaling factors to the `Topology` class directly, so that they can be accessed through `top.scaling_factors['whatever']`. This might be tricky to bookkeep without clear relationships between the classes, but can be done if we are careful

justinGilmer:
>     1. Add a  `top.force_field` attribute that behaves as a pointer, so we could do things like `top.force_field.scaling_factors['whatever']`. This assumes, for now, that there is only one force field per topology, and also that the data contained in each should match.
> 
>     2. Add the scaling factors to the `Topology` class directly, so that they can be accessed through `top.scaling_factors['whatever']`. This might be tricky to bookkeep without clear relationships between the classes, but can be done if we are careful

Hmm that is a good point. Ive been thinking this over and I think I am a bit more on board with number 1. However, it is sort of the opposite of how we do things now. The sites and such have the atom types, once it is a typed topology. I feel like either the atomtypes should carry this, which is a bit redundant or it is carried at the top of the topology's hierarchy.

Although I am not married to either idea yet. But this will be an issue we need to solve soon, since this information we will need when writing out typed systems to most engines.

Maybe we do just carry a pointer to the forcefield, and if the system is typed, we should be able to use the scaling factor from that forcefield.

Although this leads to another issue if you wanted to write out a topology object to a serializable format. Since the forcefield will not be following along (unless we decide to do that either by writing out the xml again, or including all the xml data in the json as well, or when loading in from a JSON file, we could specify if we want to bring the FF information contained in the json, or load a new ff with the structure.), if we are relying on the FF object for that information, we lose that.

gmso/formats/mcf.py
```diff
@@ -0,0 +1,705 @@
...
+        mcf.write(
+            "{:<4d}  {:<4d}  {:<4d}  {:<4d}  {:<4d}"
+            "  {:s}  {:8.3f}  {:8.3f}\n".format(
+                i + 1,
+                improper.atom1.idx + 1,
+                improper.atom2.idx + 1,
+                improper.atom3.idx + 1,
+                improper.atom4.idx + 1,
+                improper_type,
+                improper.type.psi_k * KCAL_TO_KJ,
```
justinGilmer: I feel like i must be overlooking it, but where is this conversion factor stored, and could we use unyt here?

gmso/formats/mcf.py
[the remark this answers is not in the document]
justinGilmer: Makes sense to me!

gmso/formats/mcf.py
rsdefever: Nope, you're not overlooking anything. I have the improper section commented out for now, so that function isn't used. I'll update it when I come back and add 1-4 scaling and add the tests for bonded systems. I'd vote to get this merged for now and then as we continue to expand the tests for GMSO I'll come back and finish off these few details. You'll notice there are some "TODO:"s scattered throughout the code.

PR #343, opened 2020-03-09
Get site indices by indexed set

rmatsum836:
This PR addresses the way that site indices are set and written out in `top.py`.  Originally, `idx` was added as a site attribute.  Now, the `indexed_set` of `top.sites` is used to grab the indices.

I also removed `bond_idx`, `angle_idx`, and `dihedral_idx` since those don't seem to get used.

### DIFF
--- a/gmso/formats/top.py
+++ b/gmso/formats/top.py
@@ -13,7 +13,6 @@ def write_top(top, filename):
 
     _validate_compatibility(top)
     top_vars = _get_top_vars()
-    _assign_indices(top)
 
     with open(filename, 'w') as out_file:
         out_file.write(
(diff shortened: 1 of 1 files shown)


gmso/formats/top.py
```diff
@@ -186,10 +185,10 @@ def write_top(top, filename):
                 '\t{8}'
                 '\t{9}'
                 '\t{10}\n'.format(
-                    dihedral.connection_members[0].idx,
-                    dihderal.connection_members[1].idx,
```
mattwthompson: oops!

rmatsum836:
> LGTM, as long as the performance doesn't take a hit (it shouldn't, but you never know)
> 
> Is this built off of the most recent master branch? a.k.a. does it have #330 built in? That round-trips with ParmEd would give me confidence that it writes out the right files

#330 is built in but let me check that the top files look correct.

rmatsum836:
I inspected the top file after the changes and everything looks good.  I did a simple benchmark writing out a system of 100 waters, and it takes the same amount of time to write out a top file before and after the changes.

Codecov is failing due to a drop in coverage though and I'm not sure why...

justinGilmer:
Yea that is odd. It it most likely just a tolerance issue. Merging.

rsdefever:
I'm late to the party here -- but quick question to help me understand the choice. 

What is the goal of this change? Readibility? Removing an attribute? 

Should I be accessing site indices this way in the MCF writer as well?

rmatsum836:
The main motivation for this change is that we were adding an attribute to `Site` outside of the `Site` module.  This was initially confusing for me.

After some discussion, it turns out this attribute wasn't needed as `topology.sites` get stored as an IndexedSet.  The IndexedSet assigns an index for each site which we can use for our writers.

And yes, you should be able to utilize this in your MCF writer as well.  Instead of keeping track of a site index with `enumerate`, you can just call `top.sites.index(site)` to grab the index for that site.  Let me know if you have any further questions.

rsdefever:
> The main motivation for this change is that we were adding an attribute to Site outside of the Site module. This was initially confusing for me.

This is what I was missing. Thanks!

PR #351, opened 2020-03-12
Add a LAMMPS data file reader V2

rmatsum836:
Based on #347, using some parts of the old code and some new.  Here, I'm trying to make the code more modular and general for all of the types of systems LAMMPS supports.

A couple features in particular:
- `get_units`: A function that contains a dictionary of dictionaries for various LAMMPS unit styles.  That way various unit styles can be supported
- `get_connection`: A function intended to grab all bonds, angles, and dihedrals in the LAMMPS data file.
- Added a `potential` argument to (eventually) read in multiple potential types.  Default is set to `LJ`.


gmso/formats/lammpsdata.py
```diff
@@ -199,3 +209,228 @@ def write_lammpsdata(topology, filename, atom_style='full'):
         # TODO: Write out bonds
         # TODO: Write out angles
         # TODO: Write out dihedrals
+
+
+def read_lammpsdata(filename, atom_style='full', unit_style='real', potential='lj'):
+    """
```
justinGilmer: For numpydoc style, the single line should start at the end of the triple quotes. 
```suggestion
    """Read in a lammps data file as a GMSO topology
```

gmso/formats/lammpsdata.py
```diff
@@ -199,3 +209,228 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+    Notes
+    -----
+    See http://lammps.sandia.gov/doc/2001/data_format.html for a full description of the LAMMPS data format.  
+    Currently only supporting LJ potential parameters.
+    """
+    # TODO: Add argument to ask if user wants to infer bond type
+    top = Topology()
+
+    # Validate 'atom_style'
+    if atom_style not in ['full']:
```
justinGilmer: this works now, but something that might be a bit easier to handle later is a list of acceptable `atom_styles` we check against, similar to the way we handle the potential templates for various writers.

gmso/formats/lammpsdata.py
```diff
@@ -199,3 +209,228 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+                connection_members=site_list,
+                connection_type=connection_type_list[int(line.split()[1])-1],
+                    )
+        topology.add_connection(connection)
+
+    return topology
+
+def _get_atoms(filename, topology, unit_style, type_list):
+    """
+    Function to parse the atom information in the LAMMPS data file
```
justinGilmer: same docstring comment as above

gmso/formats/lammpsdata.py
```diff
@@ -199,3 +209,228 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+        x_line = line.split()
+        y_line = lammps_file.readline().split()
+        z_line = lammps_file.readline().split()
+
+        x =  float(x_line[1])-float(x_line[0])
+        y =  float(y_line[1])-float(y_line[0])
+        z =  float(z_line[1])-float(z_line[0])
+
+        # Box Information
+        lengths = u.unyt_array([x,y,z], get_units(unit_style)['distance'])
```
justinGilmer: How do we want to handle the [tilt factors?](https://lammps.sandia.gov/doc/read_data.html)


>These are the recognized header keywords. Header lines can come in any order. The value(s) are read from the beginning of the line. Thus the keyword atoms should be in a line like “1000 atoms”; the keyword ylo yhi should be in a line like “-10.0 10.0 ylo yhi”; the keyword xy xz yz should be in a line like “0.0 5.0 6.0 xy xz yz”. All these settings have a default value of 0, except the lo/hi box size defaults are -0.5 and 0.5. A line need only appear if the value is different than the default.

Currently this will make orthogonal boxes.

gmso/tests/base_test.py
```diff
@@ -7,7 +7,7 @@
 from gmso.core.topology import Topology
 from gmso.core.element import Hydrogen, Oxygen
 from gmso.core.site import Site
-from gmso.core.angle import Angle
+from gmso.core.angle import Angle
```
justinGilmer: extraneous space
```suggestion
from gmso.core.angle import Angle
```

gmso/tests/base_test.py
```diff
@@ -50,13 +50,13 @@ def _topology(sites=1):
             top = Topology()
             top.box = Box(lengths=[1, 1, 1])
             H = Hydrogen
-            site1 = Site(name='site1',
-                         element=H,
-                         atom_type=AtomType(name="at1",
-                                            mass=H.mass),
-                         )
             for i in range(sites):
```
justinGilmer: if you set `site=3`, would this work?

or is it not expected?

gmso/formats/lammpsdata.py
```diff
@@ -199,3 +209,228 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+                return topology
+            if connection_type.capitalize() in line.split():
+                break
+    connection_type_lines = open(filename, 'r').readlines()[i+2:i+n_connection_types+2]
+    connection_type_list = list()
+    for line in connection_type_lines:
+        if connection_type == 'bond':
+            c_type = BondType(name=line.split()[0]
+                    )
+            c_type.parameters['k']=float(line.split()[1])*u.Unit(
```
justinGilmer: maybe a small comment that you are converting k/2 to k?

was not immediately obvious when i looked at it

gmso/tests/base_test.py
rmatsum836: There are better pytest fixtures that exist now, so I'm now using the `typed_ar_system` for these tests.  And then I plan to remove the `topology_site` fixture in a separate PR.

gmso/formats/lammpsdata.py
rmatsum836: Good point, I need to look into this further.

rmatsum836: Tried to add coverage for triclinic boxes but it definitely needs to be validated.

rmatsum836:
Need to verify that the units are correct based on discussions from #339

rmatsum836:
Reading triclinic boxes is working now.

gmso/formats/lammpsdata.py
```diff
@@ -225,3 +235,265 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+def _get_ff_information(filename, unit_style, topology):
+    """Function to parse atom-type information
+    """
+    with open(filename, 'r') as lammps_file:
+        types = False
+        for i, line in enumerate(lammps_file):
+            if 'atom' in line:
+                n_atomtypes = int(line.split()[0])
+                types = True
+            elif 'Masses' in line:
```
mattwthompson: Kinda worried about capitalization in cases like these; I don't know how LAMMPS treats them (`1 atom types` vs `1 Atom types`). The LAMMPS documentation is a beast to navigate but it seems like it's lowercase in the header and capitalized (`Atoms`) later on when actually enumerating the data. So this should work

https://lammps.sandia.gov/doc/read_data.html

gmso/formats/lammpsdata.py
```diff
@@ -225,3 +235,265 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+    potential: str, optional, default='lj'
+        Potential type defined in data file
+
+    Returns
+    -------
+    top : GMSO Topology
+        A GMSO Topology object
+
+    Notes
+    -----
```
mattwthompson: This needs some beefing up to clarify what is and is not supported right now. I think, for example, dihedrals aren't read in properly now

gmso/formats/lammpsdata.py
```diff
@@ -225,3 +235,265 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+    # Parse atom type information
+    top, type_list = _get_ff_information(filename, unit_style, top)
+    # Parse atom information
+    _get_atoms(filename, top, unit_style, type_list)
+    # Parse connection (bonds, angles, dihedrals) information
+    # TODO: Add more atom styles
+    if atom_style in ['full']:
+        _get_connection(filename, top, unit_style, connection_type='bond')
+        _get_connection(filename, top, unit_style, connection_type='angle')
+
```
mattwthompson: I don't know if this is needed, but it would make me more comfortable
```suggestion
    top.update_top()
```

rmatsum836:
Thanks for the comments @mattwthompson.  Will address in the next couple of days.

gmso/formats/lammpsdata.py
```diff
@@ -225,3 +235,265 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+def get_units(unit_style):
+    """Get units for specific LAMMPS unit style
+    """
+    # Need separate angle units for harmonic force constant and angle
+    unit_style_dict = {
+            'real': {'mass': u.g,
+                      'distance': u.angstrom,
+                      'energy': u.kcal/u.mol,
+                      'angle_k': u.radian,
+                      'angle': u.degree,
```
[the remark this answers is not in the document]
rmatsum836: Yeah I wasn't sure about this either.  My thoughts converting the exceptions is that this isn't as explicit (as you mentioned).  This might get overhauled anyways once we get Unyt unit objects set up.

gmso/tests/test_lammps.py
```diff
@@ -17,3 +16,66 @@ def test_write_lammps_triclinic(self, topology_site):
...
+                filename='data.water')
+        water = read_lammpsdata('data.water')
+
+        # Add once charges are added
+        #assert u.array.allclose_units(water.sites[0].charge,
+        #        u.unyt_array(-0.834, u.elementary_charge))
+        assert water.n_sites == 6
+        assert water.n_connections == 6
+
+    def test_read_lammps_triclinic(self, topology_site):
```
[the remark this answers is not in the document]
rmatsum836: I agree, I replaced it with `typed_ar_system`.

gmso/formats/lammpsdata.py
rmatsum836: I agree this is something we should keep an eye out for in the future.

gmso/formats/lammpsdata.py
rmatsum836: Add some information as you suggested.  But I think there's room for improvement in future PRs.

rmatsum836:
I think I've addressed most of @mattwthompson 's comments.  There are still lingering issues to be mindful of, but good to merge?

mattwthompson:
I just looked through this - comments in one tab, diff in another, - and everything I brought up has either been addressed or agreed to as something to deal with later. Again, nice work :+1:

PR #356, opened 2020-03-16
NetworkX BondGraph Connectivity (#140 Duplicate)

uppittu11:
Same as #140. Most of @ahy3nz code still works. Might need a few updates.


umesh-timalsina:
I have refactored `connectivity.py` to consolidate some common code. Also, I think `connectivity.py` should be moved to `utils` module.

ahy3nz:
I'm somewhat confident in Networkx's ability to identify these bonding patterns, but functions people probably need to look at: trimming any "redundant" bonded tuples and formatting the atom order of the bonded tuples (for example the improper is formatted with central atom first)

rsdefever:
Is there a reason the ethane test doesn't check the number of dihedrals? Even the methane test could confirm that the number of dihedrals is 0.

uppittu11:
Ok one issue I found, which is related to @ahy3nz's comment, is that if `top.enumerate_connectivity()` is run twice, duplicate connections are added. One fix would be to loop through the connections to make sure that the `connection_members` list don't match the `connection_members` list for any of the connections identified by `enumerate_connectivity()`. However, this would be super slow for a large system. 

Does anyone know a better way to prevent duplicate connections from being added?

uppittu11:
I'm an idiot and I messed up rebasing this PR like 4 times 😬. But at least now we have a clean commit log in the "Commits" tab, and there are no more additions/deletions to `topology/core...` instead of `gmso/core...`

umesh-timalsina (approved):
Merged with master and tested locally, everything's working.

PR #360, opened 2020-03-16
XML  Writer for a GMSO ForceField (Closes #250)

### LINKED ISSUE #250: XML Writer for Forcefield
Create a writer for forcefield class.

umesh-timalsina:


gmso/core/forcefield.py
```diff
@@ -77,6 +82,69 @@ def atom_class_groups(self):
...
+        angle_type_groups = self._group_by_expression(_type='AngleType')
+        dihedral_type_groups = self._group_by_expression(_type='DihedralType')
+        return {
+            'atom_types': atom_type_groups,
+            'bond_types': bond_type_groups,
+            'angle_types': angle_type_groups,
+            'dihedral_types': dihedral_type_groups
+        }
+
+    def _group_by_expression(self, _type):
```
bdice: Adding developer docs to all methods (including internal methods) would be helpful. I don't know what this function is meant to do. I would also suggest `type_` instead of `_type`, as [described by PEP-8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles).
> `_single_leading_underscore`: weak "internal use" indicator. E.g. `from M import *` does not import objects whose names start with an underscore.
> `single_trailing_underscore_`: used by convention to avoid conflicts with Python keyword, e.g.
> `Tkinter.Toplevel(master, class_='ClassName')`

gmso/core/forcefield.py
```diff
@@ -77,6 +82,69 @@ def atom_class_groups(self):
...
+        grouped_coll = {}
+        this_type = type_groups.get(_type, [])
+        for type_name in this_type:
+            this_expression = str(this_type[type_name].expression)
+            grouped_coll[this_expression] = grouped_coll.get(this_expression, [])
+            grouped_coll[this_expression].append(this_type[type_name])
+        return grouped_coll
+
+    def to_xml(self, filename=None, overwrite=True):
+        """Save this forcefield as a topology XML file
```
bdice: Should this docstring replace "topology" with "gmso"? Unclear to me which references are supposed to be the (former) package name vs. actually meaning "topology."
```suggestion
        """Save this forcefield as a topology XML file
```

gmso/tests/test_forcefield_xml.py
```diff
@@ -136,6 +138,12 @@ def test_elementary_charge_to_coulomb(self, ff):
         elementary_charge = ff.atom_types['Li'].charge.to(u.elementary_charge)
         assert elementary_charge.units == u.Unit(u.elementary_charge)
 
+    def test_write_xml(self, ff):
+        dir_name = mkdtemp()
```
bdice: https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp
> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

I strongly recommend removing `mkdtemp` and using `tempfile`'s context managers for tests that need to write files.
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory
e.g. `with TemporaryDirectory() as tempdir: ...`

gmso/utils/topology_writer.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _assign_parameters(params_unit_dict, parent_tag):
+    for param, unit in params_unit_dict.items():
+        param_unit_el = etree.SubElement(parent_tag, 'ParametersUnitDef')
+        param_unit_el.attrib['parameter'] = param
+        param_unit_el.attrib['unit'] = unit
+
+
+def _create_element_tree(types_dict, parent_el, types):
+    _type = types[0:-1]
```
bdice: `type_` as mentioned above.

gmso/utils/topology_writer.py
```diff
@@ -0,0 +1,134 @@
...
+        _assign_parameters(params_unit_dict, types_grp)
+        for core_type in types_dict[expr]:
+            type_el = etree.SubElement(types_grp, _type)
+            _populate_attribs(type_el, core_type, attributes[_type])
+
+
+class TopologyXMLWriter:
+    @staticmethod
+    def write(**kwargs):
+        """Write a topology xml based on the key-word arguments"""
```
bdice:
```suggestion
        """Write a topology xml based on the keyword arguments."""
```

gmso/utils/topology_writer.py
```diff
@@ -0,0 +1,134 @@
...
+        # Create the root element
+        ff_el = etree.Element('ForceField')
+        ff_el.attrib['name'] = kwargs.get('name', 'ForceField')
+        ff_el.attrib['version'] = kwargs.get('version', '1.0.0')
```
bdice: Should force field versions default to 1.0.0 or something like "None" / "No version specified"?

gmso/utils/topology_writer.py
```diff
@@ -0,0 +1,134 @@
...
+        params_unit_dict = _parse_parameters_units(types_dict)
+        _assign_parameters(params_unit_dict, types_grp)
+        for core_type in types_dict[expr]:
+            type_el = etree.SubElement(types_grp, _type)
+            _populate_attribs(type_el, core_type, attributes[_type])
+
+
+class TopologyXMLWriter:
+    @staticmethod
+    def write(**kwargs):
```
bdice: I generally caution against functions that take only `**kwargs` unless directly forwarding those to another function. It's essentially impossible to document a method's behavior when it accepts only `**kwargs`. There are several fields like "name" and "version" that take on default values. Maybe consider making those explicit keyword arguments instead of `**kwargs`.

Another option would be to accept an instance of some data class that supports specific keys (e.g. "name" and "version" but not "my_random_key") if you have a codified schema.

Also, if a user provides an unsupported keyword argument, I didn't see any kind of validation or warning/error for that.

umesh-timalsina:
This needs some restructuring, a merge against master as well some of the suggestions pointed out by @bdice. Closing this and will open another PR for this.

PR #362, opened 2020-03-19
Add a class method in potential to create Potential Object. Closes #358.

### LINKED ISSUE #358: Add a class method to create a Potential out of PotentialTemplate

umesh-timalsina:
Closes #358.

### DIFF
--- a/gmso/core/potential.py
+++ b/gmso/core/potential.py
@@ -224,19 +224,20 @@ class Potential(object):
 
         Returns
         -------
-        gmso.Potential,
+        gmso.Potential
             The potential object created
 
         Raises
         ------
-        GMSOError :
+        GMSOError
             If potential_template is not of instance PotentialTemplate
         """
         from gmso.lib.potential_templates import PotentialTemplate
         if not isinstance(potential_template, PotentialTemplate):
             raise GMSOError(f'Object {type(potential_template)} is not an instance of PotentialTemplate.')
 
-        return cls(expression=potential_template.expression,
+        return cls(name=potential_template.name,
+                   expression=potential_template.expression,
                    independent_variables=potential_template.independent_variables,
                    parameters=parameters,
                    topology=topology,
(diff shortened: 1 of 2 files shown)


gmso/core/potential.py
```diff
@@ -209,6 +209,39 @@ def __repr__(self):
...
+        potential_template : gmso.lib.potential_templates.PotentialTemplate,
+                            The potential template object
+        parameters : dict,
+                    The parameters of the potential object to create
+        topology : gmso.Topology, default=None
+                   The topology to which the created potential object belongs to
+
+        Returns
+        -------
+        gmso.Potential,
```
justinGilmer: Should it be this?
```suggestion
        gmso.Potential :
```

gmso/tests/test_potential.py
```diff
@@ -222,3 +224,13 @@ def test_set_expression_and_params_mismatch(self):
                     'u': 1.0*u.g,
                     'v': 1.0*u.m},
             )
+
```
justinGilmer: Would it make sense to have a test that instantiates a type of potential (harmonic bond or anything is fine) the traditional way, and then also make one with the template fashion and check if all of their attributes are the same?

gmso/core/potential.py
umesh-timalsina: According to numpy docstring guide, there should neither be ',' nor ':'  in the docstring for Returns in this case:
```


Explanation of the returned values and their types. Similar to the Parameters section, except the name of each return value is optional. The type of each return value is always required:

Returns
-------
int
    Description of anonymous integer return value.

If both the name and type are specified, the Returns section takes the same form as the Parameters section:

Returns
-------
err_code : int
    Non-zero value indicates error code, or zero on success.
err_msg : str or None
    Human readable error message, or None on success.


```

gmso/tests/test_potential.py
umesh-timalsina: Ok, I will add the test.

Issue #364, opened 2020-03-20
Profiling LAMMPS functions


rmatsum836:
I'm starting to look at the performance of the LAMMPS writer and reader.  I looked at a tip3p water system with `n` molecules, where `n = [100, 500, 1000, 2000]`.  Comparing `write_lammpsdata` in `mBuild` and `GMSO`, here is what I got:

[times.pdf](https://github.com/mosdef-hub/gmso/files/4361546/times.pdf)

Obviously this performance isn't acceptable, but I'm not quite sure why the performance is so bad in comparison to mBuild.  I'm going to spend some time digging deeper into this.

rmatsum836:
https://gist.github.com/rmatsum836/45f1ac390cd7908a27ea45cb30a59e2a

rmatsum836:
One of the culprits was calling `top.is_typed`.  Adding the argument `updated=True` makes this function significantly faster though.  Will submit a PR

[times.pdf](https://github.com/mosdef-hub/gmso/files/4361968/times.pdf)

uppittu11:
I'll try to summarize what we talked about over Slack so we have it on record here. 

Notebook I used for testing: https://gist.github.com/uppittu11/6789bd71ce77e6d31141626e7c9f6a42

Another major bottleneck in the lammpswriter was calling `top.sites.index(someSite)`. This is because `top.sites` is a tuple, which has slow O(N) lookup.
https://github.com/mosdef-hub/gmso/blob/5781bf8a4917e83cac541a0c5d9343a4d2cbc1f5/gmso/core/topology.py#L227-L229
![image](https://user-images.githubusercontent.com/33552857/77213751-f0e1d180-6ad9-11ea-96a5-fbd7c7304bda.png)


Another option was making top.sites return an indexed set: 
```python3
 @property 
 def sites(self): 
     return IndexedSet(self._sites) 
```

This made lookup twice as fast, but still O(N) since every time type.sites is called, the IndexedSet needs to be rebuilt. 
![image](https://user-images.githubusercontent.com/33552857/77213795-225a9d00-6ada-11ea-8153-db9832b24e4c.png)

Currently we are looking at options for how to improve the output of `top.sites`. 

1. @umesh-timalsina  suggested that we could have a `top.get_site_index(site)` function that would get the index of a site using the internal `top._sites` `IndexedSet` in O(1) time. This can be also be extended to bonds, angles, dihedrals, and impropers. 
[...]

PR #366, opened 2020-03-20
Update `is_typed` call in LAMMPS writer

rmatsum836:
Discussed in #364, the `is_typed` call was one of the culprits causing `write_lammpsdata` to be slow.  By adding the `updated=True` argument to the function, the writer is now much faster.


rmatsum836:
You're right.  It might be best to close this and focus on improving `is_typed` then.

PR #367, opened 2020-03-22
Add get_index method in topology (Closes #365)

### LINKED ISSUE #365: Performance
GMSO's topology object's methods are not performing well when compared with mbuild as pointed by @rmatsum836 in #364. It might be worthwhile to investigate this and make changes to the functions/properties.

umesh-timalsina:
Spawned off from #364, Closes #365

1. Add `get_index` method in the Topology class, which makes indexing of members O(1) vs O(N) currently pointed in #364.


rmatsum836:
This is looking good so far.  I think we'll also want `get_index` to cover the `atom_types`, `bond_types`, etc. iterators as well since this is necessary information for the writers.  But I think those would get handled the same way sites, bond, etc. get handled.

umesh-timalsina:
Do we need `Potential` objects in the topology to be indexable as well? That would be a little bit tricky and would be handled differently.

rmatsum836:
> Do we need `Potential` objects in the topology to be indexable as well? That would be a little bit tricky and would be handled differently.

I don't think so.  Off the top of my head, I can't think of a situation where potentials need to be  indexed.  But may need some other opinions as well.

umesh-timalsina:
As discussed, we should also be able to index `Potential` objects in the topology.

gmso/core/topology.py
```diff
@@ -610,6 +611,34 @@ def update_topology(self):
         self.update_connection_types()
         self.is_typed(updated=True)
 
+    def get_index(self, member):
+        """Get index of a member in the topology
+
+        Parameters
+        ----------
+        member : one of {gmso.Site, gmso.Bond, gmso.Angle, gmso.Dihedral, gmso.Improper, gmso.Connection}
```
rmatsum836: I think remove `gmso.Connection` since we agreed today that initializing a generic `Connection` would not be valid.

rmatsum836: Could we also change this to:
```
member : A GMSO core object of `Site`, `Bond`, `Angle`, `Dihedral`, or `Improper`    
```

Also could we add a `Notes` section? Something like:
```
Notes
------
`get_index` method currently not supported for Potential objects.
```

gmso/tests/test_topology.py
```diff
@@ -418,4 +418,45 @@ def test_topology_atom_type_changes(self):
...
+        a_dihedral = Dihedral(connection_members=[conn_members[0], conn_members[1], conn_members[2], conn_members[3]])
+        an_improper = Improper(connection_members=[conn_members[0], conn_members[1], conn_members[2], conn_members[3]])
+
+        top.add_site(a_site)
+        top.add_connection(a_bond)
+        top.add_connection(an_angle)
+        top.add_connection(a_dihedral)
+        top.add_connection(an_improper)
+
+        assert top.get_index(a_site) == 9
```
justinGilmer: Just to make sure i am properly understanding. We add those 5 sites in the `for` loop, and when we create a connection member, if those sites are not in the topology to begin with, those are added too?

So in this case, we add 5 sites in the for loop `0`, and then `1,2,3,4` are from the connection members, then the additional 4 sites from the rest of the for loop: `5,6,7,8`

and then `a_site` at `9`?

I can understand that, but I am confused about the connection members below.

Does this:
```python3
            top.add_connection(Angle(
                connection_members=[conn_members[0], conn_members[1], conn_members[2]]))
```
Not just keep adding the same connection over and over again, but since it is the same, its just added once?
I would assume that the index for each connection type after the for loop would only have `0`, just the single one.

Then a second one after the for loop?

umesh-timalsina: https://github.com/mosdef-hub/gmso/blob/master/gmso/core/topology.py#L373-L375
Since the internal representation is an IndexedSet. Only one instance of a `Site` can exist in the topology.

Issue #354, opened 2020-03-12
Remove and Replace `topology_site` fixture


rmatsum836:
Since we've recently added pytest fixtures for untyped and typed Ar and water systems, I think we can remove the `topology_site` fixture and replace it.  

I believe this fixture is currently only being used in the lammps and openmm tests.

mattwthompson:
+1, the Argon fixture could probably be dropped down from 100 atoms, or maybe modified to take in N in order to accomadate systems of different sizes, if we need just a single-site system

daico007:
I agree, modifying the Ar to take in N as an argument seems like the better option

rmatsum836:
I agree, this would be useful for the unit tests @rsdefever wrote for his MCF writer #340.

PR #368, opened 2020-03-23
Remove topology_site fixture from base_test

rmatsum836:
Resolving #354 by removing the `topology_site` fixture and replacing with `typed_ar_system` in the LAMMPS and OpenMM tests.  Going to work on adding a `n_sites` argument as well.

### DIFF
--- a/gmso/tests/base_test.py
+++ b/gmso/tests/base_test.py
@@ -49,15 +49,7 @@ class BaseTest:
 
     @pytest.fixture
     def ar_system(self, n_ar_system):
-        #ar = mb.Compound(name='Ar')
-
-        #packed_system = mb.fill_box(
-        #    compound=ar,
-        #    n_compounds=10,
-        #    box=mb.Box([3, 3, 3]),
-        #)
-
-        return from_mbuild(n_ar_system(n_sites=100))
+        return from_mbuild(n_ar_system())
 
     @pytest.fixture
     def n_ar_system(self):
--- a/gmso/tests/base_test.py
+++ b/gmso/tests/base_test.py
@@ -48,7 +48,19 @@ class BaseTest:
 
 
     @pytest.fixture
-    def ar_system(self):
+    def ar_system(self, n_ar_system):
+        #ar = mb.Compound(name='Ar')
+
+        #packed_system = mb.fill_box(
+        #    compound=ar,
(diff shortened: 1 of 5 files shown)


rsdefever:
Can we get that `n_sites` argument added before merging? We can then update the relevant tests and have it done at once.

rmatsum836:
> Can we get that `n_sites` argument added before merging? We can then update the relevant tests and have it done at once.

I looked into this a little bit last week.  I was following this blog post: https://alysivji.github.io/pytest-fixures-with-function-arguments.html.  I was running into an issue where the pytest fixture ended up being a function, rather than a GMSO topology.  I could have been doing something wrong though and will look through this again today.

rsdefever:
@rmatsum836 if this is too much trouble or there is urgency to getting this PR merged don't let it hold you up. I just thought it was a neat idea and fit nicely with this PR.

rmatsum836:
So tests are going to fail, but my last commit adds an `n_sites` to `ar_system`.  This seems clunky to me since `ar_system` ends up being passed as a function instead of a topology.  The tests in `test_top.py` have been updated to reflect these changes.  Could y'all take a look and see what you think?

umesh-timalsina:
> This seems clunky to me since ar_system ends up being passed as a function instead of a topology. The tests in test_top.py have been updated to reflect these changes.

You could add two fixtures. `typed_ar_system` would still be a topology, that uses `n_typed_ar_system`  fixture which would would be a function. If needed, could be used by other test cases.

```py
@pytest.fixture
def n_typed_ar_system(self):
    def _topology(n_sites=10):
           ..........
             return typed_ar
    return _topology

@pytest.fixture()
def typed_ar_system(self, n_typed_ar_system):
      return n_typed_ar_system(default_value)

```

rmatsum836:
Thanks @umesh-timalsina for that suggestion.  I added two pytest fixtures: `n_ar_system` and `n_typed_ar_system`.  These two fixtures return a function and both contain the argument `n_sites` so that a user can specify how many sites to include in the topology.  

The current tests `ar_system` and `typed_ar_system` have been refactored so that `n_ar_system` and `n_typed_ar_system` respectively are called.

@rsdefever I also removed your `single` pytest fixtures and replaced them with `n_typed_`. 

Please let me know your comments and suggestions.

PR #370, opened 2020-03-24
Add delete site method in topology (Closes #161)

### LINKED ISSUE #161: remove_site functionality in Topology
Do we have a `remove_site()`/`delete_site()` function in `Topology`? If we don't, do we want to add one?

umesh-timalsina:
Closes #161. Adds `delete_site` method and a simple test.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -642,17 +642,19 @@ class Topology(object):
         site : gmso.Site
             A Site object that exists on the topology.
 
-        Returns
+        Raises
         -------
-        gmso.Site
-            The site object removed from the topology.
+        GMSOError
+            If `site` doesn't exist in the topology's sites.
 
         Notes
         -----
         Removing a site from the topology removes any bond,
         angles, dihedrals and impropers associated with the site.
         """
-        if site in self._sites:
+        if site not in self._sites:
+            raise GMSOError(f"Site {site} does not exist in the topology")
+        else:
             self._sites.discard(site)
             self._dead_sites.add(site)
             for connection in site.connections:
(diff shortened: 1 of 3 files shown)


umesh-timalsina:
This PR now depends on (#378): as deleting a `connection_type` or `atom_type` based on a deleted site would mean that `connection_type`/`atom_type` should not be associated with any other `connection`/`site` in the topology.

gmso/core/topology.py
```diff
@@ -610,6 +631,59 @@ def update_topology(self):
...
+        Removing a site from the topology removes any bond,
+        angles, dihedrals and impropers associated with the site.
+        """
+        if site not in self._sites:
+            raise GMSOError(f"Site {site} does not exist in the topology")
+        else:
+            self._sites.discard(site)
+            self._dead_sites.add(site)
+            for connection in site.connections:
+                if connection in self._connections:
```
mattwthompson, line 661: This, and making a half-dozen new internal variables, seems like a lot of overhead for this behavior. But I believe it carefully deals with our mess of pointers that we've gotten ourselves in and the only alternative I can propose is moving this behavior into the `Connection` class itself and letting it do the big if block there. Actually, I kinda like that idea, in case there is another need to delete sites from elsewhere.

gmso/core/topology.py
```diff
@@ -610,6 +631,59 @@ def update_topology(self):
...
+                        if partner is not site:
+                            partner.connections.discard(connection)
+        site.connections.clear()
+
+    def _compact_set(self, member, member_type='sites'):
+        current_dead, current_set = self._del_ref_sets[member_type]
+        if member in current_dead:
+            current_set._compact()
+            current_dead.discard(member)
+
```
mattwthompson, line 686: I feel like a `top.update_top()` is needed somewhere in here to clean up `ConnectionType` objects

umesh-timalsina: There's an open PR to get Associations for a `Potential` or an `AtomType` (#378). That PR when merged would deal with removing them more efficiently IMO.

daico007:
Replaced by https://github.com/mosdef-hub/gmso/pull/761

Issue #350, opened 2020-03-12
OpenMM conversion


daico007:
Now that the Parmed conversion is almost in place #334 (please have a look at it if you guys have a chance), I think we should also add conversion function (to and from) for OpenMM topology structure.

PR #371, opened 2020-03-26
OpenMM conversion

daico007:
This PR will add OpenMM conversion function to the GMSO package. This include the addition of from conversion functions and beefing up the to openmm conversion functions. Relate to #350.


daico007:
Question: since OpenMM topology' atoms doesn't seem to carry position information, do you know where can we extract that information (if not, how should we handle this miss of information)?

rmatsum836:
> Question: since OpenMM topology' atoms doesn't seem to carry position information, do you know where can we extract that information (if not, how should we handle this miss of information)?

I don't think the OpenMM Topology carries position information for atoms.  I would focus on the `modeler` class for now as it carries positions.  For a `topology` we could consider having the user pass in a coordinate file (like `gro` or `mol2`), but I'm not sure how useful this might be.

mattwthompson:
I see the `Topology` class much more often at interoperability points than the `Modeller` class so my inclination would be to go for that.

I don't know if this is bulletproof but you can build a `Modeller` from a `Topology`, having thrown other stuff in, and can also go backwards with [`Modeller.getTopology`](http://docs.openmm.org/development/api-python/generated/simtk.openmm.app.modeller.Modeller.html#simtk.openmm.app.modeller.Modeller.getTopology). Consider also Christoph's [original implementation](https://github.com/mosdef-hub/foyer/blob/1870d8eaeb4d39ba36c6eefbcb417e239429a0f7/foyer/forcefield.py#L899-L909) of carrying both `topology` and `positions` objects along for the ride.

ahy3nz:
It's definitely going to be a little messy because of the distinction in openmm objects: System for force field information, topology for atoms/bonds, and something separate for positions (the openmm system context stores positions but might not be set if there's no context).

>Consider also Christoph's original implementation of carrying both topology and positions objects along for the ride

> For a topology we could consider having the user pass in a coordinate file (like gro or mol2), but I'm not sure how useful this might be

If you can be flexible and modular, it would be nice to populate a gmsoTopology based on the information it has available to it (openmm system, openmm topology, openmm context, or some other input that a gmsoTopology could use)

daico007:
Hey @justinGilmer, so far I have only added a convert from box vectors method in the `box.py` (my reference is parmed `geometry.py` files). If you have written up a more general methods for gmso box, I can put this off and wait for your PR (I only need this for the box conversion of the openmm conversion, so I can work on other stuff first)

daico007:
Closed for later

PR #378, opened 2020-03-27
Add get_associations method to topology class (Closes #375)

### LINKED ISSUE #375: Add number of connections assocaited with a connection_types in a topology
Add extra bookkeeping record for sites_addition, deletion  etc... among others.

umesh-timalsina:
adds `get_association` method to return a set of sites associated with an `atomtype` or
a set of bonds associated with a bond_type and so on.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -137,11 +137,16 @@ class Topology(object):
         self._impropers = IndexedSet()
         self._subtops = IndexedSet()
         self._atom_types = {}
+        self._atom_types_associations = {}
         self._connection_types = {}
         self._bond_types = {}
+        self._bond_types_associations = {}
         self._angle_types = {}
+        self._angle_type_associations = {}
         self._dihedral_types = {}
+        self._dihedral_types_associations = {}
         self._improper_types = {}
+        self._improper_types_associations = {}
         self._combining_rule = 'lorentz'
         self._set_refs = {
             ATOM_TYPE_DICT: self._atom_types,
@@ -149,6 +154,13 @@ class Topology(object):
             ANGLE_TYPE_DICT: self._angle_types,
             DIHEDRAL_TYPE_DICT: self._dihedral_types,
             IMPROPER_TYPE_DICT: self._improper_types,
+        }
+        self._association_refs = {
+            ATOM_TYPE_DICT: self._atom_types_associations,
+            BOND_TYPE_DICT: self._bond_types_associations,
+            ANGLE_TYPE_DICT: self._angle_type_associations,
+            DIHEDRAL_TYPE_DICT: self._dihedral_types_associations,
+            IMPROPER_TYPE_DICT: self._improper_types_associations
         }
 
     @property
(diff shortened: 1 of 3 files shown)


gmso/tests/test_topology.py
```diff
@@ -418,4 +418,23 @@ def test_topology_atom_type_changes(self):
...
+        atom_type = AtomType(name='test_atomtype')
+        for i in range(10):
+            top.add_site(Site(name=f'site{i+1}', atom_type=atom_type), update_types=True)
+        assert len(top.get_associations(atom_type)) == 10
+
+    def test_get_association_atom_types_after_changes(self, typed_ar_system):
+        typed_ar_system.atom_types[0].name = 'Typed Ar System'
+        assert typed_ar_system.get_associations(typed_ar_system.atom_types[0]) == set(typed_ar_system.sites)
+
+    def test_get_association_bond_types(self, typed_water_system):
```
justinGilmer, line 432: not sure if this fits here, but what if we had a system with bonds (water would be an example: H-O-H)

In this case, bondtype should have a set that contains all 3 sites. If we then removed a bond  between one of the H and O (H     O-H)

We expect that the `len(get_associations(broken_water)` is now 2.

Would be a good test to have. and this same example should work for angletype too.

gmso/core/topology.py
```diff
@@ -543,7 +568,7 @@ def add_subtopology(self, subtop):
         subtop.parent = self
         self._sites.union(subtop.sites)
 
-    def is_typed(self, updated=False):
+    def is_typed(self, updated=False, ):
```
rmatsum836, line 571: Is there a reason for this comma?

gmso/core/topology.py
```diff
@@ -137,11 +137,16 @@ def __init__(self, name="Topology", box=None):
         self._impropers = IndexedSet()
         self._subtops = IndexedSet()
         self._atom_types = {}
+        self._atom_types_associations = {}
         self._connection_types = {}
         self._bond_types = {}
+        self._bond_types_associations = {}
```
rmatsum836, line 143: Really small nit but types is plural in `_atom_types_associations` and `_bond_types_associations` but is singular in `_angle_type_associations`.  I think I would actually prefer these all be singular type (`_atom_type_associations`, `_bond_type_associations`, etc.) because it reads a little bit better to me, but it's worth getting opinions from the rest of @mosdef-hub/mosdef-contributors.

Either way the naming should be consistent.

umesh-timalsina: The reason for naming it with `_****_types_associations` was because it is a dictionary of all the connection types in the topology, which will contain a set. I have no preference one way or other.

gmso/core/topology.py
umesh-timalsina: Not really. No Idea why it was added.

gmso/core/topology.py
rmatsum836: Got it, I'm fine with `_****_types_associations then`.

umesh-timalsina:
Probably better to close it and open it again.

PR #380, opened 2020-03-27
Enforce Connection Rules (WIP)

### LINKED ISSUE #377: How do we want to create layered potentials?
@rsdefever That's a good point. maybe instead of merging equivalent connections, we could make sure that the connection members are sorted when writing to file to make it easier to debug. 

Maybe we discussed it elsewhere, but to layer potentials, do we want to create new `Connections` or do we want to be able to add multiple `ConnectionTypes` to a single `Connection`?

_Originally posted by @uppittu11 in https://github.com/mosdef-hub/gmso/issues/361#issuecomment-600638642_

umesh-timalsina:
This is a work in progress to enforce connection rules between sites. The related issue is #377.

To close #377, We need the following:

- [x] Enforce connection rules between the sites and define equivalencies of a connection (3b35008)[WIP]
- [ ] Allow a connection to have multiple connection types

### DIFF
--- a/gmso/core/angle.py
+++ b/gmso/core/angle.py
@@ -35,6 +35,10 @@ class Angle(Connection):
         super(Angle, self).__init__(connection_members=connection_members,
                 connection_type=connection_type, name=name)
 
+    def get_equivalent_partners(self):
+        # ToDO: Add proper docstring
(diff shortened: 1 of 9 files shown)


umesh-timalsina:
Closing this in favor of #384.

Issue #313, opened 2020-03-02
Support pair potential matrix parameters


joaander:
For example, a typical pair potential for bead spring polymers is:
```
A-A: epsilon=1.0, sigma=1.0, r_cut=2.5
A-B: epsilon=1.0, sigma=1.0, r_cut=2**(1/6)
B-B: epsilon=1.0, sigma=1.0, r_cut=2.5
```

I realize that this example only varies r_cut along the matrix which is possibly out of scope for GMSO. There are other cases where the potential parameters themselves will vary.

uppittu11:
Should we add a `PairType` object, with is similar to the `BondType`, but applies to non-bonded pair potentials? That way you can define custom unique pair potentials (like my tabulated CG non-bonded for example).

mattwthompson:
Parashara and I talked about this at length and we eventually agreed there should be a new class that defines cross-interactions. The current `AtomType` class is implicitly self-self, and it would be useful to have another class, that also subclasses out of `Potential` that more clearly represents a cross-interaction. There could be some utility functions that take in objects `atom_type_i` and `atom_type_j` and produces `cross_interaction_ij`, but this would not necessarily be the intended constructor.

@zijiewu3 needs this so he will be working on it. @shivrajkotkar and @jingxiangguo will also find this useful so it would be worth working on it together

I think it is also useful to consider a way to either have this behavior as a thin wrapper around existing objects or finagle the existing `AtomType` class to point to different objects. But there are downsides to those that prevent me from advocating for them - mostly that pointers/wrappers don't as explicitly store this data in memory.

b-butler:
I think it makes sense for a `PairType`, and in this case move all the potential parameters to it and don't store potential parameters in `AtomType`. The two could interface using the `AtomType` to inform choice of potential parameters. It seems to be a bifurcation of the same concept into two classes to have self interactions in `AtomType` and mixture interactions in another class.

daico007:
Basic support via https://github.com/mosdef-hub/gmso/pull/484

PR #382, opened 2020-04-02
Add cross pairpotentials to topology

zijiewu3:
As a first attempt, here is how I enabled the topology to store information about nonbonded pair potentials that do not follow combination rules, as discussed in #313 .
-A pair potential is implemented as `PairPotentialType` class that inherits from `Potential` class. I used `member_types` parameter to specify the pair of `AtomTypes` this pair potential refers to.
-in `gmso.core.Topology`, I added an `add_pairpotential` function that allows adding `PairpotentialType` object to a `Topology` instance. I tried to make it disconnected from the `sites` and `connections` and remain as standalone as possible.
https://github.com/zijiewu3/gmso/blob/9b5af6028dafa2fe9cb26e334e4d0644f961f9c1/gmso/core/topology.py#L501-L513
-I also added a test case in `test_topology.py`.
https://github.com/zijiewu3/gmso/blob/9b5af6028dafa2fe9cb26e334e4d0644f961f9c1/gmso/tests/test_topology.py#L359-L372
I guess the biggest question I have is how this interacts with the combining rule that is already there. My naive thought will be when the backend applies nonbonded interactions between pairs, it first checks if a `pairpotential_type` is specified for the given pair of `AtomTypes`. If so, the interaction is defined by `pairpotential_type`. If not, the combining rule is used.


gmso/core/topology.py
```diff
@@ -546,6 +578,22 @@ def update_atom_types(self):
...
+        See Also:
+        ---------
+        gmso.PairPotentialType: A description of nonbonded interaction between 2        atomtypes
+        """
+        if not isinstance(pair_potential_type, PairPotentialType):
+            raise GMSOError(
+                'Non pairpotential type {}'.format(pair_potential_type))
+
+        elif pair_potential_type not in self._pairpotential_types:
+            pair_potential_type.topology = self
```
umesh-timalsina, line 593: Since it looks like a `PairPotentialType` is a descripton of the nonbondedinteraction between 2 atomtypes, it might help to check if the `AtomTypes` exist in the topology before adding a `PairPotentialType` between the.

gmso/core/pairpotential_type.py
```diff
@@ -0,0 +1,117 @@
...
+class PairPotentialType(Potential):
+    """A descripton of the nonbondedinteraction between 2 atomtypes.
+
+    This is a subclass of the gmso.core.Potential superclass.
+
+    PairPotentialType represents a pair potential type and includes the functional form describing
+    its nonbonded interactions that doesn't follow usual cross-average combining rules.
+    The functional form of the potential is stored as a
+    `sympy` expression and the parameters, with units, are stored explicitly.
+    The AtomTypes that are used to define the pair potential type are stored as
```
rmatsum836, line 20: Nit but the names of the atom types in the form of strings are being stored in `member_types`.

rmatsum836:
Thanks for the contribution, this is a great start!  Just a few minor comments.

In addition to the tests that Umesh mentioned, I think it would also be good to add a test that raises an Error when a `PairPotentialType` is added with `member_types` that do not exist in Topology.

> You should be able to get a PairPotentialType interactions between two arbitary AtomTypes if they interact. Although, @mosdef-hub/mosdef-maintainers could chime in more about this

@umesh-timalsina Do you mind clarifying this sentence?  In this case, `PairPotentialTypes` can be thought of as exceptions to mixing rules between two atom types, so we should only generate them explicitly.

> I guess the biggest question I have is how this interacts with the combining rule that is already there. My naive thought will be when the backend applies nonbonded interactions between pairs, it first checks if a pairpotential_type is specified for the given pair of AtomTypes. If so, the interaction is defined by pairpotential_type. If not, the combining rule is used.

As far as I know we're not currently keeping track of mixing rules in GMSO.  In `foyer` the mixing rule can be specified, and I think the plan is to encode the mixing rules within the FF XMLs.  I think checking `pairpotential_types` vs implied mixing rules will be specific to the engine we're writing out to.  In GROMACS, I believe the potential interactions are implied by the specified mixing rule unless otherwise defined.  In this case, we would just write out the potential parameters for each `AtomType` and `PairPotentialType`.  In LAMMPS, I could be wrong but I think either atom_i potentials can be written out (mixing rules are inferred), or atom_ij potentials can be written out (i_j interactions explicitly specified).  In the case of `PairPotentialType`, we would need to ensure that the `PairPotentialType` gets written out instead of the mixing rule.  I think this can be done by building up a list of pair combinations of `AtomType`, and compare to the member types in `PairPotentialType`.  I'm not sure about other engines though so @mosdef-hub/mosdef-contributors may want to chime in.

zijiewu3:
Thanks for your review and feedback! I made all the changes Ray requested.
In terms of the two points Umesh suggested:

> Since it looks like a `PairPotentialType` is a descripton of the nonbondedinteraction between 2 atomtypes, it might help to check if the `AtomTypes` exist in the topology before adding a `PairPotentialType` between them.
Since `PairpotentialType` is detached from other parts of a topology, I wonder if this is necessary. If one of the `Atomtypes` are not in the topology, that shouldn't break anything. There are also some merits of allowing this behavior. For example, the user can specify the cross interaction potentials first before they add atoms, rather than having to do it the other way around. There are also cases in my workflow where I have an array of similar simulations, some of which have a certain type of atoms and the others don't. Allowing this behavior might also make it convenient to batch-process these systems. Definitely welcome to other thoughts on this issue.

>You need to add a test that makes sure that the `PairPotentialType` exists in the topology even after some changes are made to its properties.
Would you mind clarifying what exactly you would like over here?

rmatsum836:
We've had some recent discussion regarding the API and how to handle these N-body potentials in GMSO.  Currently a `Connection` can represent a pair potential.  The lingering question is whether or not we should have an abstract class higher than `Connection` that can correctly encapsulate both the physical connections (bonds, angles, dihedrals) and the non-physical associations (N-body potentials).

Issue #361, opened 2020-03-17
Standardize indexing in `connection_members`


rmatsum836:
When writing out angles and dihedrals, it will be important to follow the syntax for specific engines.  For example, LAMMPS requires angles to be written out as `angle-1 angle-2 angle-3` where angle-2 is the central atom (https://lammps.sandia.gov/doc/2001/data_format.html).  

In `connection_members` we should standardize the indexing so that it's consistent and easier to abide by the syntax set by the various engines.

uppittu11:
Maybe we could also have a function that would merge duplicate connections into one. For example, an angle with members 1, 2, 3 and another angle with members 3, 2, 1.

rsdefever:
> we should standardize the indexing so that it's consistent and easier to abide by the syntax set by the various engines.

Agree. I also like the `i-j-k` order where `j` is the central atom.  

> Maybe we could also have a function that would merge duplicate connections into one. For example, an angle with members 1, 2, 3 and another angle with members 3, 2, 1.

We need to be careful here. IIRC, we want to support multiple connections being defined with the same members, i.e., layering potentials.

uppittu11:
@rsdefever That's a good point. maybe instead of merging equivalent connections, we could make sure that the connection members are sorted when writing to file to make it easier to debug. 

Maybe we discussed it elsewhere, but to layer potentials, do we want to create new `Connections` or do we want to be able to add multiple `ConnectionTypes` to a single `Connection`?

rsdefever:
> Maybe we discussed it elsewhere, but to layer potentials, do we want to create new `Connections` or do we want to be able to add multiple `ConnectionTypes` to a single `Connection`?

Very curious others opinions here. I can think of it either way: 

* Multiple `ConnectionTypes` to a single `Connection`: There is only one physically meaningful connection. We are applying some model potential to that physical entity to make it behave the way we want. Sometimes we need multiple 'layered' potentials to get the physics right. This is ultimately a modelling issue and does not change the fact that there is one connection that we are trying to model appropriately. Therefore, there is only one `connection` even if it has multiple layered `ConnectionTypes`. 
* Multiple `Connections`: the `Connection` is the way we indicate that the application of a potential to a some set of sites. Therefore, it makes sense to keep a 1-to-1 mapping between a `Connection` and its associated `ConnectionType`.

[...]

PR #384, opened 2020-04-02
Ensure that all connections are unique

uppittu11:
Related to #380 and #377. 

This PR makes sure that equivalent connections are not added based on the connection rules defined in https://github.com/mosdef-hub/gmso/issues/361#issuecomment-601473060. 

So far, merging this PR will...
* add get_equivalent_members function to get a unique data structure to represent the connection members.
* modify topology.add_connection to prevent adding an existing connection or equivalent connection to the topology.
* add tests to verify that equivalent connection aren't added to the topology.

### DIFF
--- a/gmso/core/angle.py
+++ b/gmso/core/angle.py
@@ -35,13 +35,13 @@ class Angle(Connection):
         super(Angle, self).__init__(connection_members=connection_members,
                 connection_type=connection_type, name=name)
 
-    def get_equivalent_members(self):
-        """Get a unique dataset representing the connection
+    def equivalent_members(self):
+        """Get a set of the equivalent connection member tuples
 
         Returns
         _______
-        tuple
-            A unique tuple to represent the connection members
+        frozenset
+            A unique set of tuples of equivalent connection members
 
         Notes
         _____
@@ -49,12 +49,32 @@ class Angle(Connection):
             i, j, k == k, j, i
         where i, j, and k are the connection members.
         """
+        return frozenset([
+                tuple(self.connection_members),
+                tuple(reversed(self.connection_members))
+                ])
 
-        return tuple([
+    def _equivalent_member_hash(self):
+        """Get a hash representing the connection
+
+        Returns
+        _______
+        int (hash)
+            A unique hash to represent the connection members
+
+        Notes
+        _____
+        For an angle:
+            i, j, k == k, j, i
+        where i, j, and k are the connection members.
+        Here, j is fixed and i and k are replaceable.
+        """
+
+        return hash(tuple([
             self.connection_members[1],
             frozenset([self.connection_members[0],
                        self.connection_members[2]])
-            ])
+            ]))
 
 def _validate_three_partners(connection_members):
     """Ensure 3 partners are involved in Angle"""
--- a/gmso/core/angle.py
+++ b/gmso/core/angle.py
@@ -19,7 +19,7 @@ class Angle(Connection):
     connection_type : gmso.AngleType, optional, default=None
         AngleType of this angle.
     name : str, optional, default="Angle"
-        Name of the angle. 
+        Name of the angle.
 
     Notes
     -----
@@ -35,6 +35,26 @@ class Angle(Connection):
         super(Angle, self).__init__(connection_members=connection_members,
                 connection_type=connection_type, name=name)
 
+    def get_equivalent_members(self):
+        """Get a unique dataset representing the connection
+
+        Returns
+        _______
+        tuple
+            A unique tuple to represent the connection members
+
+        Notes
+        _____
+        For an angle:
+            i, j, k == k, j, i
+        where i, j, and k are the connection members.
+        """
+
+        return tuple([
+            self.connection_members[1],
+            frozenset([self.connection_members[0],
+                       self.connection_members[2]])
+            ])
 
 def _validate_three_partners(connection_members):
     """Ensure 3 partners are involved in Angle"""
--- a/gmso/core/dihedral.py
+++ b/gmso/core/dihedral.py
@@ -40,6 +40,29 @@ class Dihedral(Connection):
         super(Dihedral, self).__init__(connection_members=connection_members,
                 connection_type=connection_type, name=name)
 
+    def get_equivalent_members(self):
+        """Get a unique dataset representing the connection
+
+        Returns
+        _______
+        tuple
+            A unique tuple to represent the connection members
+
+        Notes
+        _____
+        For an dihedral:
+            i, j, k, l == l, k, j, i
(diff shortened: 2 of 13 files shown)


uppittu11:
One change I made was to make it so that `topology.add_connection()` returns the connection object that is added (or already in) the topology. This way if some one tries to add an equivalent connection and wants to modify the connection that is in the topology, they can just catch the output of the function and modify that. 

One thing that needs to be added (in this PR or probably another one) is to define what behavior we want when we try to add an equivalent connection.
* Do nothing
* Replace the current metadata with the new connection's metadata (like connection_types and name)
* Append the connections metadata

I'm open to suggestions.

daico007:
May be we can have an `append` flag that is set to `False` by default (or 'True', idk). So if we try to add an equivalent connection when `append` is `False`, the code just gonna throw an error (or warning).

gmso/core/topology.py
```diff
@@ -369,11 +370,25 @@ def add_connection(self, connection, update_types=True):
         update_types : bool, default=True
             If True also add any Potential object associated with connection to the
             topology.
+
+        Returns
+        _______
+        Connection
```
justinGilmer, line 376: can you update the returns section a bit more?
```suggestion
        Connection : gmso.Connection
```

And for the other methods, specifying the `gmso.Bond` etc. would probably be nice.

gmso/core/angle.py
```diff
@@ -35,6 +35,26 @@ def __init__(self, connection_members=[], connection_type=None, name="Angle"):
...
+        For an angle:
+            i, j, k == k, j, i
+        where i, j, and k are the connection members.
+        """
+
+        return tuple([
+            self.connection_members[1],
+            frozenset([self.connection_members[0],
+                       self.connection_members[2]])
+            ])
```
umesh-timalsina, line 57: This is really good. Would it help to make this structure more accessible?  There could be other use cases for the equivalent connection members of a connection. This is tailored for use inside the `Topology` class dictionary key. However, if there are other utilities for this method, then it should probably return a tuple of all possible combinations of equivalencies, but should have an internal property that could then be levaraged by `Topology` class.

gmso/core/topology.py
```diff
@@ -369,11 +370,25 @@ def add_connection(self, connection, update_types=True):
...
+            The Connection object or equivalent Connection object that
+            is in the topology
         """
+        # Check if an equivalent connection is in the topology
+        equivalent_members = connection.get_equivalent_members()
+        if equivalent_members in self._unique_connections:
+            warnings.warn('An equivalent connection already exists.')
```
justinGilmer, line 383: I think it would make sense to give one more sentence about the process happening under the hood.

```python3
warnings.warn('An equivalent connection already exists. Providing the equivalent Connection.')
```
Or something to that effect.

uppittu11:
@umesh-timalsina's comment (https://github.com/mosdef-hub/gmso/pull/384#pullrequestreview-386805944) 
Having bonds that are not accounted for in Site.n_connections is a problem. A hot fix would be to make adding a duplicate connection for a site an error instead of a warning. That way `bond2` would never be created. However, I don't really see the point of tracking connections in each Site object outside of a Topology. It seems like there would need to be a lot of complex bookkeeping to always ensure that these are correct. In my mind there is a hierarchy of objects: Sites are the lowest level objects and shouldn't "see" the Connections they are involved in; Connections are made up of sites, but shouldn't "see" the Topology they are contained in; Topology is the highest level object and contains all Sites and Connections. If we want to be able to get all of the connections that a site is involved with, this should be the responsibility of the Topology.

gmso/core/angle.py
uppittu11: So do you think `self.get_equivalent_members()` should return something like this:
```python3
((self.connection_members[0], self.connection_members[1], self.connection_members[2]),
 (self.connection_members[2], self.connection_members[1], self.connection_members[0]))
```
and then the current `get_equivalent_members()` should be renamed and become an internal method?

gmso/core/dihedral.py
```diff
@@ -40,6 +40,29 @@ def __init__(self, connection_members=[], connection_type=None, name="Dihedral")
...
+        """
+
+        return tuple([frozenset([
+            frozenset([self.connection_members[0],
+                       self.connection_members[1]]),
+            frozenset([self.connection_members[1],
+                       self.connection_members[2]]),
+            frozenset([self.connection_members[2],
+                       self.connection_members[3]])
+            ])])
```
rsdefever, line 65: Can you explain the logic here in words? I follow the bond and angle logic. It seems to work based on the tests but I'm not quite following.

gmso/core/improper.py
```diff
@@ -45,6 +45,27 @@ def __init__(self, connection_members=[], connection_type=None, name="Improper")
...
+            i, j, k, l == i, k, j, l
+        where i, j, k, and l are the connection members.
+        """
+
+        return tuple([
+            self.connection_members[0],
+            self.connection_members[3],
+            frozenset([self.connection_members[1],
+                       self.connection_members[2]])
+            ])
```
rsdefever, line 68: Can you explain the logic here in words as well?

gmso/core/site.py
```diff
@@ -184,13 +190,20 @@ def _validate_atom_type(val):
...
     if site not in connection.connection_members:
         raise GMSOError("Error: Site not in connection members. Cannot add the connection.")
-    if connection in site.connections:
-        return None
+
+    # Check if an equivalent connection is in the topology
+    equivalent_members = connection.get_equivalent_members()
+    if equivalent_members in site._unique_connections:
+        warnings.warn('An equivalent connection already exists.')
+        return site._unique_connections[equivalent_members]
```
rsdefever, line 207: Nice!

gmso/core/site.py
```diff
@@ -68,12 +68,18 @@ def __init__(self,
         self._charge = _validate_charge(charge)
         self._mass = _validate_mass(mass)
         self._connections = IndexedSet()
+        self._unique_connections = {}
 
     def add_connection(self, connection):
         connection = _validate_connection(self, connection)
+
         if connection:
```
rsdefever, line 76: Why is this if statement here? Won't that always evaluate to `True`?

umesh-timalsina:
Depends on #390.

uppittu11:
Why do we have both of these showing up in the status checks?
https://dev.azure.com/mosdef/mosdef/_build?definitionId=9&_a=summary
https://dev.azure.com/mattthompson0739/gmso/_build?definitionId=2&_a=summary

daico007:
Hey @uppittu11, can you update this PR and resolve the conflicts that are popping up?

uppittu11:
Closed in favor of #460

Issue #387, opened 2020-04-08
Hierarchy of class structures


vyasr:
This issue is an offshoot of #385 which I'm using to document a number of questions about the data structures in GMSO. I took a brief look at GMSO's core this morning, and here are my somewhat organized thoughts. I'm sure a lot of these are dumb questions that have already been considered and addressed in during the development process, but hopefully the viewpoint of someone on the outside looking in will help clarify some of the concepts. I can update this if I think of more things in the future.

### Topologies

1. What exactly is the purpose of a subtopology? Is the goal to represent something like a chain in a protein, since for instance a PDB file can contain multiple chains? I'm trying to understand why it makes sense to have a distinct subtopology object, rather than simply allowing topologies to have a parent (that could be `None`) and reusing that class. Is there really a meaningful difference between what could constitute a topology and what could constitute a subtopology, and are there limits to the nesting?
2. The properties all have to convert from an IndexedSet to a list every time they're called; is that really necessary? I don't know exactly what the use cases you foresee for these are, but would it be so bad to force the user to do the conversion since it could be kind of expensive if they put it in a loop or something?

### Sites
1. The Site class claims to "as general as possible" to support all kinds of simulations. This is handled by allowing all of the properties of a `site` to be `None` if desired. What is the most general kind of site that could be permitted? A site with name "Site" and position `(0, 0, 0)`, with all other properties as `None`? If so, I'm wondering if it makes more sense to subclass site for specific types (e.g. `Atom`) and put the properties there.
2. Should sites really know their positions, or are they more about the "nature" of the site (i.e. a certain kind of atom or coarse-grained bead). This point and (1) are pretty closely related to point (3) in #385 I believe.
[...]

Issue #385, opened 2020-04-06
Site Properties


umesh-timalsina:
1. Currently, a site in GMSO can access its connection, should we remove it?
2. `n_connections` property in a site. Is it worth it?
3. Should an atom_type be associated with a site? Or, should that information be stored in a Topology only?

uppittu11:
Here is my opinion on this:
1. The hierarchy of physical GMSO objects is 
`Topology` (highest) -> `Subtopology` -> `Connection` -> `Site` (lowest). 
And each object shouldn't be able to access the objects above it in the hierarchy. So, a `Site` shouldn't be able to access `Connection`s.  
2. `n_connections` isn't necessary to store in site. The only time we might need it is if the atom type assigned to a site depends on its hybridization state (sp2, sp3, etc.). And even so, this would only matter if the site is in a topology. So, it might be worth having a dictionary mapping `{site: connections}` within topology, but the number of connections doesn't need to be a site property. 
3. This one is a little more controversial, but I think that a sites should not be able to access their `atom_type`. The assignment of atom types to a site is only relevant when the sites and atom_types are inside of a topology, so the `{site: atom_type}` mapping should be done only in topology.

rmatsum836:
1.  I agree with @uppitu11.  This would remove a lot of the bookkeeping, and would allow us to get rid of the update functions which have been a big bottleneck.
2. I'm a little bit more on the fence here.  I think I would be okay with this if we did have a dictionary that mapped sites to connections within topology, that way that information is still easy for a user to access if wanted.  A little more off-topic, but I'm not sure `n_connections` being a sum of all bonds, angles, and dihedrals is super useful.
3. I think I'm actually okay with this, as long as we have the `{site: atom_type}` dictionary mapping within topology.  But I think I would at least want to have an attribute `is_typed` or something like that to distinguish between sites that have types and sites that don't.  I'm also unsure about the scenario where a user may want to take a subset of atoms that have `atom_type I` and tweak a parameter.  In this case, should we direct them to initialize a new `atom_type J` that would be point to those specified atoms?

daico007:
1. I am agree with both @uppittu11 and @rmatsum836 on this matter.
2. Having a dictionary mapping sounds good to me.
[...]

PR #390, opened 2020-04-13
Remove connection reference from Site (closes #389)

### LINKED ISSUE #389: Class Hierrarchy Enforcement
1. The hierarchy of physical GMSO objects is 
`Topology` (highest) -> `Subtopology` -> `Connection` -> `Site` (lowest). 
And each object shouldn't be able to access the objects above it in the hierarchy. So, a `Site` shouldn't be able to access `Connection`s.  
2. `n_connections` isn't necessary to store in site. The only time we might need it is if the atom type assigned to a site depends on its hybridization state (sp2, sp3, etc.). And even so, this would only matter if the site is in a topology. So, it might be worth having a dictionary mapping `{site: connections}` within topology, but the number of connections doesn't need to be a site property. 
3. This one is a little more controversial, but I think that a sites should not be able to access their `atom_type`. The assignment of atom types to a site is only relevant when the sites and atom_types are inside of a topology, so the `{site: atom_type}` mapping should be done only in topology.

_Originally posted by @uppittu11 in https://github.com/mosdef-hub/gmso/issues/385#issuecomment-610509597_

umesh-timalsina:
* As discussed over #385, #387, #389, refactor sites such that they are not able to reference their connections
* Refactor tests such that they are passing

### DIFF
--- a/gmso/core/site.py
+++ b/gmso/core/site.py
@@ -35,11 +35,6 @@ class Site(object):
        The atom type of the site containing functional forms, interaction parameters, and other properties such as mass and charge.  
        See `atom_type.py` for more information.
 
-    Attributes
-    ----------
-    n_connections : int
-       Number of connections for the site
-
     """
     def __init__(self,
                  name='Site',
--- a/gmso/core/site.py
+++ b/gmso/core/site.py
@@ -1,13 +1,9 @@
 import warnings
-from functools import reduce
-
 
 import numpy as np
 import unyt as u
-from boltons.setutils import IndexedSet
 
 from gmso.core.atom_type import AtomType
-from gmso.exceptions import GMSOError
 
 
 class Site(object):
@@ -41,8 +37,6 @@ class Site(object):
 
     Attributes
     ----------
-    connections : IndexedSet
-       Set that contains connection information for the site
     n_connections : int
        Number of connections for the site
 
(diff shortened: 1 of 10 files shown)


gmso/core/site.py
```diff
@@ -41,8 +37,6 @@ class Site(object):
 
     Attributes
     ----------
-    connections : IndexedSet
-       Set that contains connection information for the site
     n_connections : int
```
rmatsum836, line 40: We're removing `n_connections` as well.

gmso/core/topology.py
```diff
@@ -403,82 +403,6 @@ def add_connection(self, connection, update_types=True):
...
-
-        This method is an alias for gmso.Topology.update_connections.
-
-        See Also
-        --------
-        gmso.Topology.update_connections : Update all the Bonds, Angles, Dihedrals, and Impropers in the topology.
-        """
-        self.update_connections(update_types)
-
     def update_connection_types(self):
```
rmatsum836, line 406: Can we get rid of `update_connection_types` as well?

gmso/tests/test_improper.py
```diff
@@ -15,17 +15,8 @@ def test_improper_nonparametrized(self):
         site3 = Site(name='site3')
         site4 = Site(name='site4')
 
-        assert site1.n_connections == 0
```
rmatsum836, line 18: This goes for a lot of these tests, but instead of testing a connection of a site, should we instead check that the `connection` object we initialize contains the sites we expect them to have?

rmatsum836:
> Just a couple of minor changes. There are still probably a handful of `update_` functions lingering in our converters and writers, so let's ensure these are all removed before merging.

Quickly glanced through and I believe all of our converters, writers, and examples use `update_topology` which has been properly updated.  Tests are passing locally for me, and the LAMMPS data file that was written out from this PR is correct.

gmso/core/topology.py
umesh-timalsina: I don't think so, because its iterating over topology.connections.

umesh-timalsina:
#389 will be closed by #431.

Issue #225, opened 2020-02-10
Handling `charge` in both Site and Atomtype


rmatsum836:
Had an in-depth discussion with @uppittu11 right now about handling charges in both `site` and `atom_type`.  He ran into issues creating a box of water where `site.charge` is the value from whatever water model he's using, but `site.atom_type.charge` is 0.  The gromacs top writer ends up grabbing the charges from `site.atom_type.charge` so the charges all end up being 0.

One solution we discussed was having a private function that validates that `site.charge` and `site.atom_type.charge` are consistent with each other.  However, I think we should sit down and map out a clear plan for how we will handle storing charges in both data structures, and how we will write out these charges to files.

PR #393, opened 2020-04-14
Allow topology to update single atomtype for a site (Closes #392)

### LINKED ISSUE #392: Allow sites to operate on a single atomtype in topology
When inside a topology, a change in site's  atom-type properties will be propagated through the entire topology. Users should have options to control this propagation.

umesh-timalsina:
As discussed in #392, this PR adds a  method called `change_atom_type_properties` in topology with the following signature:
```py
change_atom_type_properties(site, should_propoagate=False, **kwargs)
```
This method takes in an individual site and changes the atomtype properties for it (provided via) kwargs. Flag `should_propagate` can be used to control whether the property changes are for this site only (default behaviour) or the entire topology.
If this method passes the scrutiny and becomes usable, this will close #392 and similar extensions can be provided to hold `BondType/AngleType/DihedralType/ImproperType` changes from the topology. 
 One example snippet to test this branch is given below:
```py
    from gmso import Topology, Site, AtomType
    site = Site()
    site2 = Site()
    atom_type = AtomType()
    site.atom_type = atom_type
    site2.atom_type = atom_type
    top = Topology()
    top.add_site(site)
    top.add_site(site2)
    top.change_atom_type_properties(site2, should_propagate=False,
                                    name='AtomType2',
                                    charge=2.0)
    print(site2.atom_type.charge, site.atom_type.charge)
```

### DIFF
--- a/gmso/core/atom_type.py
+++ b/gmso/core/atom_type.py
@@ -6,6 +6,7 @@ from gmso.core.potential import Potential
 from gmso.utils.misc import unyt_to_hashable
 from gmso.utils.decorators import confirm_dict_existence
 from gmso.utils._constants import ATOM_TYPE_DICT
+from gmso.exceptions import GMSOError
 
 
 class AtomType(Potential):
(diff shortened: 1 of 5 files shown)


rmatsum836:
Thanks @umesh-timalsina, this looks good so far.  I think this brings up several fundamental questions that we've yet to fully address.  One is related to charge, discussed in #225.  With `change_atom_type_properties` users will have the ability to change `atom_type.charge` so the question is: if the user wants to modify the charge, should they do so through `site.charge` or `site.atom_type.charge`?  I think it would be confusing to allow the user to have the option to modify both.  Further, we would have to have to have a two-way check where `site.charge` gets updated if `site.atom_type.charge` is updated and vice-versa.  I think the same applies for a site's mass.

daico007:
I think this method works. But should we have a way to prevent the user from doing something like this:
`site.atom_type.name = "Test"` 
At this moment (as in in this PR), the above behavior is still allow and it actually change the properties of the AtomType itself (so equivalent to always have `should_propagate=True`)

umesh-timalsina:
Closing this until structures and interactions are decided.

PR #395, opened 2020-04-15
Fix indexing of connection types in `update_connection_types`

rmatsum836:
Originally added in #367, dictionaries of connection types were created to keep track of indexing.  The `_connection_types_idx` for angles and dihedrals are being set by `len(self._bond_types)` which seems to just be a typo.  This PR fixes the indexing for these connection types.


umesh-timalsina:
This was my mistake. Thanks @rmatsum836 for finding this out. The tests fail because of `assert 0` in `test_topology_get_index_angle_type` in `gmso/test/test_topolpgy.py`. If you change the first line of the test case to they will pass. 
```py
assert typed_chloroethanol.get_index(typed_chloroethanol.angles[0].connection_type) == 0
```
Having said that, it might not hurt to add few more test cases to make positive on the desired indexing behavior.

PR #398, opened 2020-04-17
Update testing and codecov

### LINKED ISSUE #397: Why are we running pytest twice?
https://github.com/mosdef-hub/gmso/blob/ec541a47ce87138b67770e521c654088ed35a9fb/azure-pipelines.yml#L70-L71

#316 is the culprit? We could probably delete the first pytest. 

On a related note, AZP has better integration with codecov now, so we can use the bash uploader instead: https://docs.codecov.io/docs/about-the-codecov-bash-uploader. This way the token won't be accessible to the public (although I guess it won't be too hard to look through the git history).

uppittu11:
Closes #397

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,7 +55,7 @@ jobs:
           conda config --add channels janschulz
           conda config --add channels conda-forge
           conda config --add channels mosdef
-        displayName: Add relavent channels
+        displayName: Add relevant channels
 
       - bash: |
           conda create -n test-environment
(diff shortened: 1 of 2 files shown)


uppittu11:
We also get this "Tests" tab now, which is pretty neat
image

azure-pipelines.yml
```diff
@@ -55,34 +55,32 @@ jobs:
...
+        displayName: Install requirements and testing branch
 
       - bash: |
           source activate test-environment
-          python -m pytest -v gmso/tests/ 
-          python -m pytest -v --cov=gmso --cov-report=html --pyargs gmso
-        displayName: Run Tests
+          pip install pytest-azurepipelines
+          python -m pytest -v --cov=gmso --cov-report=html --pyargs gmso --no-coverage-upload
+        displayName: Run tests
```
umesh-timalsina, line 72: What is the relation of `pytest-azurepipelines` with the task `PublishCodeCoverageResults@1`? And Is the `--no-coverage-upload` flag mandatory?

uppittu11: I think `pytest-azurepipelines` pipes the warnings and errors to azure's logger, so it gets published to that "Tests" page. The warnings also pop up in the build dashboard so you can search and filter those. I think it's a pretty cool feature. 

It can also upload an html coverage report to the "Code Coverage" tab in the azure build if you exclude the `--no-coverage-upload` flag. But the issue is that (1) we don't want to upload a coverage report for each build, only the linux+python3.7 build and (2) the report it publishes is kinda lame and has boring formatting. The` PublishCodeCoverageResults@1` publishes the html report using the Cobertura parser which applies some formatting and puts those line coverage bar graph that makes the "Code Coverage" tab look a little cleaner.

The reason I didn't add `pytest-azurepipelines` to the `requirements-test.txt` is because I don't want users to install that stuff or else it messes up your pytest. I'm not opposed to adding it in there if we also make a `requirements-dev.txt` for devs to use without all of the azure packages.

umesh-timalsina: I think this is a good approach to follow. Since, this is a package required for `azp` and `azp` only, It might not be necessary to add it to requirements-test.txt. However, in the future if there are more specific installation customization necessary for `azp`, we could write a script to append the files before running tests. Something like `.azp/extra-packages.sh`.

```bash
echo pytest-azurepipelines >> requirements-test.txt
```
and run this command if needed. However, this is fine for now.

PR #402, opened 2020-05-15
Add XML Converters

rmatsum836:
The goal of this PR is to add XML converters for foyer -> gmso and gmso -> foyer.  The function `from_foyer` was taken from code written by @umesh-timalsina.

Checklist:

- [ ] Handle Improper Dihedrals
- [ ] Handle conversion between `type` and `class`
- [ ] Properly convert units specified in GMSO XML to foyer units
- [ ] Unit Tests

### DIFF
--- a/gmso/external/convert_foyer.py
+++ b/gmso/external/convert_foyer.py
@@ -31,6 +31,7 @@ def from_foyer(foyer_xml, gmso_xml):
             'harmonic_bond_types': [],
             'harmonic_angle_types': [],
             'urey_bradley_angle_types': [],
+            'rb_torsion_dihedral_types': [],
             'periodic_torsion_dihedral_types': []
             }
 
@@ -69,6 +70,11 @@ def from_foyer(foyer_xml, gmso_xml):
     for ptf in periodic_torsion_force_el:
         for dihedral_type in ptf.getiterator('Proper'):
             f_kwargs['periodic_torsion_dihedral_types'].append(dihedral_type)
+
+    rb_torsion_force_el = foyer_xml_tree.findall('RBTorsionForce')
+    for rbf in rb_torsion_force_el:
+        for dihedral_type in rbf.getiterator('Proper'):
+            f_kwargs['rb_torsion_dihedral_types'].append(dihedral_type)
 
     _write_gmso_xml(gmso_xml, **f_kwargs)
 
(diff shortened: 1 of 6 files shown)


rmatsum836:
Updated checklist:

- [ ] Make sure GMSO xml has correct units
- [ ] Can GMSO xml have `class` and `type` like foyer xml? If not how to deal with this conversion?
- [ ] Handle impropers
- [ ] Periodic dihedrals to foyer
- [ ] Add unit conversions for urey Bradley angles, impropers, and periodic dihedrals

gmso/external/convert_foyer.py
```diff
@@ -0,0 +1,582 @@
...
+            'urey_bradley_angle_types': [],
+            'rb_torsion_dihedral_types': [],
+            'periodic_torsion_dihedral_types': []
+            }
+
+    # Try to load in AtomType section
+    # Load in AtomTypes section otherwise
+    atom_types_el = foyer_xml_tree.findall('AtomType')
+    if len(atom_types_el) == 0:
+        atom_types_el = foyer_xml_tree.findall('AtomTypes')
```
justinGilmer, line 42: What happens if this is empty too? Or if it is a properly formatted foyer xml, but is empty?

gmso/external/convert_foyer.py
```diff
@@ -0,0 +1,582 @@
...
+    f_kwargs['angle_type_units'] = param_dict
+
+    urey_bradley_angle_el = gmso_xml_tree.findall('AmoebaUreyBradleyForce')
+    for ubf in urey_bradley_angle_el:
+        for angle_type in ubf.getiterator('UreyBradley'):
+            f_kwargs['urey_bradley_angle_types'].append(angle_type)
+
+    dihedral_torsion_force_el = gmso_xml_tree.findall('DihedralTypes')
+    for tf in dihedral_torsion_force_el:
+        if 'k * (1 + cos(n * phi - delta))' in tf.get('expression'):
```
justinGilmer, line 163: I dont know if this would work if the same expression was there, but shuffled around a bit. We might need to use sympify here?

gmso/external/convert_foyer.py
```diff
@@ -0,0 +1,582 @@
...
+            for dihedral_type in tf.getiterator('DihedralType'):
+                f_kwargs['periodic_torsion_dihedral_types'].append(dihedral_type)
+            # TODO: Add support from Impropers
+        elif 'c0 * cos(phi)**0' in tf.get('expression'):
```
justinGilmer, line 167: same as the comment above about sympify

rmatsum836:
I've neglected this working on other projects and I know it still needs a fair amount of work before merging.  @justinGilmer Thanks for your comments, I'll make sure to address them when I revisit this.

rmatsum836:
Added some simple tests as a check to ensure everything is working properly.  Some will fail for now.

PR #405, opened 2020-05-26
Validate site positions and change default to None (Closes #404)

### LINKED ISSUE #404: Site position needs to be validated upon assignment
```py
>>> from gmso import Site
>>> site = Site()
>>> site.position = 'a'
>>> site.position
'a'
```

umesh-timalsina:
1. Site position is [`np.nan`, `np.nan`, `np.nan`] (default)
2. position now accessible by property


gmso/core/site.py
```diff
@@ -125,14 +124,29 @@ def atom_type(self, val):
...
     def __repr__(self):
         return "<Site {}, id {}>".format(self.name, id(self))
 
 
 def _validate_position(position):
+    if position is None:
+        return None
+
     if not isinstance(position, u.unyt_array):
+        try:
```
daico007, line 144: Since we are doing the checking, should we also check/enforce that the given positions have length of 3?

justinGilmer: I agree with @daico007 , the `position.shape` should be of size `(3,)`.

gmso/tests/test_site.py
```diff
@@ -93,3 +95,16 @@ def test_add_connection_non_member(self):
             site3.add_connection(bond1)
             site2.add_connection(bond3)
             site1.add_connection(bond2)
+
```
justinGilmer, line 98: You might already have this, but it would be nice to test for an invalid `position`.
```suggestion
def test_invalid_position_size(self):
    with pytest.raises(GMSOError) as e:
        site1 = Site(name=site, position=[0.0,0.0]
```

uppittu11:
I must've missed this convo. What was the reasoning behind changing the default position to None?

gmso/core/site.py
```diff
@@ -125,14 +124,29 @@ def atom_type(self, val):
...
+    def position(self, position):
+        self._position = _validate_position(position)
+
     def __repr__(self):
         return "<Site {}, id {}>".format(self.name, id(self))
 
 
 def _validate_position(position):
+    if position is None:
+        return None
```
uppittu11, line 141: As per the slack discussion we had yesterday, I think the default position should be np.array([np.nan]*3).

umesh-timalsina: The site positions are now `u.unyt_array(np.array([np.nan]*3), u.nm)`.

PR #407, opened 2020-05-27
Integrate pydantic for an abstract base class for Site and Atom (Closes #406)

### LINKED ISSUE #406: Integrate Pydantic for Abstract Base Classes and Validation
As discussed before for #403, We should start by integrating an abstract base class for `Site` as per our discussions in #385, #387, #389 and many others briefly summarized in https://hackmd.io/7fSq5fzpRii0h0jnrIOrqA. The current `Site` class is an `Atom`, which needs to be refactored into two different classes (an abstract class for `Site` as well as an `Atom` ). While starting we should also integrate #390 as well.

umesh-timalsina:
This initial commit has all the tests passing. This commit introduces
module gmso.abc with an Abstract base class for Site that has generic
site properties. The class `gmso.core.Site` has been refactored into
Atom, which inherits for `gmso.abc.Site`. Typing and Validation is
done using pydantic.


gmso/core/topology.py
```diff
@@ -504,13 +426,13 @@ def update_connection_types(self):
                     self._bond_types_idx[c.connection_type] = len(self._bond_types) - 1
                 if isinstance(c.connection_type, AngleType):
                     self._angle_types[c.connection_type] = c.connection_type
-                    self._angle_types_idx[c.connection_type] = len(self._angle_types) - 1
+                    self._angle_types_idx[c.connection_type] = len(self._bond_types) - 1
```
daico007, line 429: I think you mean `_angle_type` here? (and also those below)

umesh-timalsina: My bad. This should be `_bond_types` and everything below should be updated accordingly.

umesh-timalsina:
See https://github.com/samuelcolvin/pydantic/issues/1577 to understand extended `Config` for the baseclass.

daico007:
So I try to build the docs locally, I can access the index pages and contributing, etc., but I could not click on any of those under Core Classes, Potential Classes. I just run `make html` on the docs folder (after installed `sphinx` and `sphinx_rtd_theme`), so did I miss anything?

Issue #416, opened 2020-06-19
Potential Expression and Parameters need to be managed together?


umesh-timalsina:
Currently, we have expression and parameters as properties in a lot of our Potential Classes. Inorder to manage them what we are essentially doing is a lot of book keeping to check if parameters are set independent of expressions and vice-versa. Since, these two entities are closely related to each other i.e. a parameter `dict` only makes sense if there's an expression associated with a `Potential`. And for `PotentialTemplate`, there would be no parameters. Inorder to deal with this issue, I propose the following change: 

- Potential Objects should carry around a single `Expression` instance (hashable) which carries the information regarding both Expression and the Parameters. This would essentially delegate the bookkeeping responsibility to the new Expression class we create and hopefully clean a lot of bookeeping in Potential class.

umesh-timalsina:
Closed by #416.

PR #419, opened 2020-06-22
Expression Class (Implements #416)

umesh-timalsina:
As discussed in #416, decoupling expression and parameters handling logic from Potential Class.


gmso/tests/test_expression.py
```diff
@@ -0,0 +1,243 @@
...
+                   'evaluate potential expression. Missing symbols: {b, a}' in e
+
+    def test_invalid_indep_vars(self):
+        with pytest.raises(ValueError) as e:
+            expression = _PotentialExpression(
+                expression='a*x+b',
+                independent_variables='j',
+                parameters=None
+            )
+            assert "symbol j is not in expression's free symbols Cannot " \
```
justinGilmer, line 64: Just a comment that I really like this check of the type of exception raised. Will be a way for me to update some of my own tests.

gmso/tests/test_expression.py
```diff
@@ -0,0 +1,243 @@
...
+    def test_non_parametric_expression(self):
+        expression = _PotentialExpression(
+            expression='a^2+2*a*b+b^2',
+            independent_variables='a',
+            parameters=None
+        )
+        assert expression.is_parametric is False
+        with pytest.raises(AttributeError) as e:
+            assert expression.parameters
+            assert 'Object of type _PotentialExpression has no attribute parameters'
```
justinGilmer, line 77: Not sure if you intended this assertion to be like the str comparisons above, but that is missing.

gmso/tests/test_expression.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def test_set_indep_variables(self):
+        expression = _PotentialExpression(
+            expression='a^2+2*a*b+b^2',
+            independent_variables='a',
+            parameters=None
+        )
+        expression.independent_variables = {'b'}
+        assert sympy.Symbol('b') in expression.independent_variables
```
justinGilmer, line 86: Just want to be sure of the expected behavior, when we change the `expression.independent_variables` to `b`,  we assert below that `a` is no longer in `expression.independent_variables`, what is `a` now at this point of the code? I would expect it to be a `parameter`, but it would not have a value, nor was it specified.

gmso/tests/test_expression.py
```diff
@@ -0,0 +1,243 @@
...
+            expression.parameters = {
+                    'l': 2.7 * u.nm,
+                    'm': 2.8 * u.nm,
+                    'n': 220.0 * u.hertz
+                }
+
+        assert expression.parameters['a'] == u.unyt_quantity(2.6, units='nm')
+        assert expression.parameters['b'] == u.unyt_quantity(2.7, units='nm')
+        assert expression.parameters['c'] == u.unyt_quantity(22.8, units='hertz')
+        assert 'w' not in expression.parameters
```
justinGilmer, line 191: Could we instead use one of the values that raised the value error above. i dont think those are in scope, but would be a small change.
```suggestion
        assert 'm' not in expression.parameters
```

gmso/tests/test_expression.py
```diff
@@ -0,0 +1,243 @@
...
+            independent_variables={'phi'}
+        )
+
+        expression_2 = _PotentialExpression(
+            expression='e^4 + e^2 + phi*2',
+            independent_variables={'phi'}
+        )
+
+        expression_3 = _PotentialExpression(
+            expression='e^4 + e^2 + phi * 8',
```
justinGilmer, line 205: I doubt sympy makes assumptions like this, but when I saw `e` above, I immediately thought of the transcendental number [`e`](https://en.wikipedia.org/wiki/E_(mathematical_constant))

However, after a bit of research, using that in sympy is defined either as [`exp(1)` or `E`](https://docs.sympy.org/latest/modules/core.html#exp1)

gmso/utils/expression.py
```diff
@@ -0,0 +1,293 @@
...
+import sympy
+
+import unyt as u
+
+from gmso.utils.misc import unyt_to_hashable
+
+class _PotentialExpression:
+    """A general Expression class with parameters
+
+    Parameters
```
justinGilmer, line 12: Maybe add a note here about the underscore class and its intended usage. Will be helpful for developers.

gmso/utils/expression.py
```diff
@@ -0,0 +1,293 @@
...
+                'Please enter a string, sympy expression or None for expression'
+            )
+
+        return expression
+
+    @staticmethod
+    def _validate_parameters(parameters):
+        """Check to see that parameters is a valid dictionary with units"""
+        if not isinstance(parameters, dict):
+            raise ValueError("Please enter dictionary for parameters")
```
justinGilmer, line 214:
```suggestion
            raise ValueError("Please enter a dictionary for parameters")
```

gmso/utils/expression.py
```diff
@@ -0,0 +1,293 @@
...
+                raise ValueError(
+                    '`independent_variables` argument was a list '
+                    'or set of mixed variables. Please enter a '
+                    'list or set of either only strings or only '
+                    'sympy symbols'
+                )
+        else:
+            raise ValueError(
+                'Please enter a string, sympy expression, '
+                'or list or set thereof for independent_variables'
```
justinGilmer, line 250:
```suggestion
                'list or set thereof for independent_variables'
```

justinGilmer:
Some of the "changes" are actually just comments, feel free to resolve them as they are updated and I will do the same.

gmso/tests/test_expression.py
umesh-timalsina: Missed it.

gmso/tests/test_expression.py
umesh-timalsina: This is intended behaviour. The `_PotentialExpression` instance in this case in `non-parametric`, which means it will have no parameters. So, `a` will be just a free variable. So, let's look at this example

```py
expression = _PotentialExpression(
            'a^x + b^y + c^z',
            independent_variables={'x', 'y', 'z'},
            parameters={
                'a': 2.6 * u.nm,
                'b': 2.7 * u.nm,
                'c': 22.8 * u.hertz
            }
        )
```
Now if you are to change the `independent_variables` for the expression above, following error is raised:

```py
expression.independent_variables = {'a', 'b', 'c'}
ValueError: Missing necessary dependencies to evaluate potential expression. Missing symbols: {y, x, z}
``` 
However, if you were to do the same for non-parametric parameters=None, it doesn't complain, because there are no parameters to infer.

May be a classmethod should be added here as well, to convert non-parametric `PotentialExpression` to a parametric `PotentialExpression`.

gmso/core/potential.py
```diff
@@ -47,28 +44,35 @@ def __init__(self,
                  template=False,
                  topology=None
                  ):
+        self._name = name
+
+        if expression is None:
```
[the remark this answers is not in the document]
umesh-timalsina: Yeah. This might have to do with a lot of our core classes constructor arguments. While it make life easier to toy around and write tests to just instantiate a class, some keyword arguments could/should be changed to positional. However, I would argue that this is not in the scope of this PR.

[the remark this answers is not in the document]
daico007: We definitely need to rethink this default expression for the potential, but yeah, I agree it's kinda belong to its own PR

requirements-test.txt
```diff
@@ -6,7 +6,7 @@ lxml
 pytest
 mbuild >= 0.10.6
 openbabel
-foyer
+foyer <= 0.7.5
```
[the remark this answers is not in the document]
umesh-timalsina: This file should not be commited. My bad.

gmso/tests/test_expression.py
daico007: I was initially confused by this too (about the test and the behavior of `independent_variables` setter). Idk if you had the same confusion as I did @justinGilmer, but I initially thought that `expression.independent_variables = {'b'}` just "kicks" `a` out of the `independent_variables` list, but it turn out the setter just overwrites whatever `independent_variables` set earlier, and just check to make sure the new outcome is consistent if the parameters have already been specified.

justinGilmer: Ok this makes a lot more sense now! Thank you both! That behavior makes complete sense. We should probably make this an example in the docs somewhere.

justinGilmer:
Do we want this merged against master, or the develop branch?

umesh-timalsina:
This should be merged to master as this PR was opened for master.

justinGilmer:
Ok perfect, just making sure!

PR #420, opened 2020-06-22
Fix element parsing by iter_subtrees_topdown 

umesh-timalsina:
In recent updates to foyer because of lark parser (see https://github.com/mosdef-hub/foyer/pull/341), `elements_by_smarts_string` is failing in gmso. An attempt to fix it is made here.


umesh-timalsina:
I have changed the docstring for `element_by_smarts_string`. Although, this module might need a refactor which warrants its own PR.

PR #421, opened 2020-06-22
Fix Channel Priority in azure pipelines and add .coveragerc (Fix #418)

### LINKED ISSUE #418: Fix channel priority in azure pipelines
In AZP, the channel priority is set in such a way that `mbuild` is always grabbed from `mosdef` channel. Since, the recent releases of `mbuild` have been in `conda-forge` channel, we find a way to fix channel priorities in `azure-pipelines`.

umesh-timalsina:
Draft PR. Depends on #420.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,8 +47,8 @@ jobs:
           conda config --set always_yes yes --set changeps1 no
           conda config --add channels omnia
           conda config --add channels janschulz
-          conda config --add channels conda-forge
           conda config --add channels mosdef
+          conda config --add channels conda-forge
           conda update conda -yq
         displayName: Add relevant channels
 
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -47,8 +47,8 @@ jobs:
           conda config --set always_yes yes --set changeps1 no
           conda config --add channels omnia
           conda config --add channels janschulz
```
justinGilmer, line 49: Since the [janschulz channel no longer is maintained](https://anaconda.org/janschulz/repo) and we do not use these packages anymore, I think we can drop this channel completely.

umesh-timalsina: Removed janschulz

umesh-timalsina:
As discussed, I have pinned mbuild version (`>=10.6`). Removed `janschulz` from the channels list. This IMO is good to go.

PR #423, opened 2020-06-25
Add support for periodic potentials in top writer

uppittu11:
This was inspired by our workflow for SciPy 2020. Just wanted to extend the support of the top writer to periodic potentials and also clean up some of the functions to make them more modular. 

* Add PeriodicTorsionPotential to accepted_potentials
* Make accepted potentials accessible to all functions
* Add a connection writer that determines the potential form and chooses the correct gmx potential type
* Ensure that units are converted correctly for dihedrals

### DIFF
--- a/gmso/formats/top.py
+++ b/gmso/formats/top.py
@@ -188,10 +188,13 @@ def _accepted_potentials():
     harmonic_bond_potential = templates['HarmonicBondPotential']
     harmonic_angle_potential = templates['HarmonicAnglePotential']
     periodic_torsion_potential = templates['PeriodicTorsionPotential']
+    rb_torsion_potential = templates['RyckaertBellemansTorsionPotential']
     accepted_potentials = [lennard_jones_potential,
                            harmonic_bond_potential,
                            harmonic_angle_potential,
-                           periodic_torsion_potential]
+                           periodic_torsion_potential,
+                           rb_torsion_potential,
+                           ]
     return accepted_potentials
 
 def _validate_compatibility(top):
@@ -229,7 +232,7 @@ def _write_connection(top, connection):
     """
 
     accepted_potentials = _accepted_potentials()
-    for ref in _accepted_potentials():
+    for ref in accepted_potentials:
         if sympy.simplify(ref.expression - connection.connection_type.expression) == 0:
             potential_name = ref.name
             break
--- a/gmso/formats/top.py
+++ b/gmso/formats/top.py
@@ -1,12 +1,11 @@
 import datetime
 
 import unyt as u
-import sympy
 
 from gmso.core.element import element_by_atom_type
 from gmso.lib.potential_templates import PotentialTemplateLibrary
 from gmso.utils.compatibility import check_compatibility
-from gmso.exceptions import GMSOError, EngineIncompatibilityError
+from gmso.exceptions import GMSOError
 
 
 def write_top(top, filename, top_vars=None):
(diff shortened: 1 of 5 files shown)


uppittu11:
An additional unit test is coming soon.

docs/design_principles.rst
```diff
@@ -23,26 +23,26 @@ of elements with each particle).
 ``GMSO`` enables the following features:
 
 * Supporting a variety of models in the molecular simulation/computational
-  chemistry community: No assumptions are made about an interaction site
-  representing an atom or bead, instead supported atomistic,
+  chemistry community: No assumptions are made about an interaciton site
+  represetning an atom or bead, instead supported atomistic,
```
rmatsum836, line 27: *representing

docs/design_principles.rst
```diff
@@ -23,26 +23,26 @@ of elements with each particle).
 ``GMSO`` enables the following features:
 
 * Supporting a variety of models in the molecular simulation/computational
-  chemistry community: No assumptions are made about an interaction site
-  representing an atom or bead, instead supported atomistic,
+  chemistry community: No assumptions are made about an interaciton site
```
rmatsum836, line 26: interaction

docs/design_principles.rst
```diff
@@ -23,26 +23,26 @@ of elements with each particle).
...
+  represetning an atom or bead, instead supported atomistic,
   united-atom/coarse-grained, polarizable, and other models!
 
-* Greater flexibility for exotic potentials: The ``AtomType`` (and analogue
-  classes for intramolecular interactions) uses ``sympy`` to store any
+* Greater flexibility for exotic potentials: The ``AtomType`` (and analoug
```
rmatsum836, line 30: *analogue

gmso/formats/top.py
```diff
@@ -174,29 +151,7 @@ def write_top(top, filename):
         )
```
rmatsum836, line 151: Small change but the dihedral comment on line 150 should match the dihedral potential form.

uppittu11: Good point. How should we handle this in the case that we have multiple dihedral types?

rmatsum836: Do you mean in the case where both harmonic and RB dihedrals would be present in the system?

uppittu11: Yeah. This would be a lot easier if there was an easier way to group dihedrals by type. Another improvement would be to have a function that returns the potential type name if you pass it a potential (kind of like what `_write_connection` does).

uppittu11:
No clue what happened with that commit message. this is what it looks like on my computer
```
$ git commit -m "Add support for custom `defaults`
> * Address the doc fixes proposed by @rmatsum836
> * Print out element to the `atoms` directive
> * Add basic support for defining combining rules
> * Add support for setting defaults (1-4 scaling and `gen-pairs`)"
```

daico007:
May need to update local branch before merging

daico007:
I am not sure what happened, but this PR still seems off from master. I saw you merged master earlier, but from the diff files, it still seems like this PR is removing all the new lines in README and undo some typo fix in `docs/design_principles.rst`.

uppittu11:
Ok somehow I screwed up my branch, but I just rebased and it seems to be fixed. You will probably need to refresh the diff and delete/re-fetch the branch locally to test it out again

gmso/formats/top.py
```diff
@@ -1,22 +1,23 @@
 import datetime
 
 import unyt as u
+import sympy
```
justinGilmer, line 4: Maybe i have missed it, but do we use sympy here?

uppittu11: Removed that and the EngineIncompatibilityError import, which are both unused

PR #428, opened 2020-07-14
Make Element class hashable

daico007:
I notice that our Element class is not hashable and comparable so I just added the __eq__ and __hash__ method for our Element class.

### DIFF
--- a/gmso/core/element.py
+++ b/gmso/core/element.py
@@ -6,6 +6,7 @@ from collections import namedtuple
 import unyt as u
 
 from gmso.exceptions import GMSOError
+from gmso.utils.misc import unyt_to_hashable
 
 class Element(namedtuple('Element', 'atomic_number, name, symbol, mass')):
     """Chemical element object
@@ -41,7 +42,7 @@ class Element(namedtuple('Element', 'atomic_number, name, symbol, mass')):
     def __hash__(self):
         return hash((self.name, self.symbol,
                      self.atomic_number,
-                float(self.mass.to('amu').value)))
+                unyt_to_hashable(self.mass)))
 
 
 def element_by_symbol(symbol):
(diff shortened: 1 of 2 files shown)


gmso/core/element.py
```diff
@@ -35,6 +35,17 @@ def __repr__(self):
                                                                       self.name, self.symbol,
                                                                       self.atomic_number,
                                                                       self.mass)
+    def __eq__(self, other):
+        return (self.__class__ == other.__class__ and
+        self.name == other.name and
+        self.symbol == other.symbol and
+        self.atomic_number == other.atomic_number and
+        self.mass == self.mass)
```
umesh-timalsina, line 43: Can't this equality be changed to hash(self) == hash(other)

gmso/core/element.py
```diff
@@ -35,6 +35,13 @@ def __repr__(self):
                                                                       self.name, self.symbol,
                                                                       self.atomic_number,
                                                                       self.mass)
+    def __eq__(self, other):
+        return hash(self) == hash(other)
+
+    def __hash__(self):
+        return hash((self.name, self.symbol,
+                     self.atomic_number,
+                float(self.mass.to('amu').value)))
```
uppittu11, line 44: Maybe we should use the `unyt_to_hashable` function for this just to be consistent. 

https://github.com/mosdef-hub/gmso/blob/master/gmso/utils/misc.py

daico007:
> Only had one small change.
> 
> Just curious, what was the motivation for this PR? Comparison of elements was working even before this PR:
> 
> ```python
> import gmso
> import unyt as u
> 
> h1 = gmso.core.element.Element(atomic_number=1, name='hydrogen', symbol='H', mass=1.0079 * u.amu)
> h2 = gmso.core.element.Element(atomic_number=1, name='hydrogen', symbol='H', mass=1.0079 * u.amu)
> print(h == h2)
> ```
> 
> The output of this is
> 
> ```
> True
> ```

This arise when I try and fail to add a gmso element to a set because our element is hashable. So this PR is more for the hashable part and not the comparison part.

PR #442, opened 2020-09-10
Extend XML schema to create ConnectionType based on AtomType atomclass (Closes #441)

### LINKED ISSUE #441: Allow connection types to be defined by `name` or `atom class`
In the current foyer xml implementation, connections can be defined either by `name` or `class` (example: https://github.com/mosdef-hub/foyer/blob/master/foyer/forcefields/xml/oplsaa.xml).  This is very useful because OPLS-AA, for example, defines bond types between two specific classes.  Rather than defining redundant bond types between different atom types, a single bond type between two classes can be defined.

I believe we could support something similar in GMSO by changing how the XML is parsed in `ff_utils.py`.  Rather than just checking the types in the connection types to the `atom_type` names, we could check for both valid `name` and `atomclass`.

umesh-timalsina:
Extend ff schema, extend _parse_atom_types in ff_utils
- Firstly, extend the schema to include class information for all the connection types in GMSO
- Second, use a chain map to create a view of both atom_type mapping based on names as well as atomclasses
- Lastly, at its current form, we need some extra logic to be added to _parse_ff_connection_types to make it work seamlessly.

Checklist:
- [x] Extend ForceField XML schema 
- [x] Parse `atomclass` information when parsing atomtypes.
- [x] Choose correct atomtype if multiple atomtypes of same class is found found (Warrants some discussion, hence this draft PR).
- [x] Tests


gmso/tests/files/ff-example0.xml
```diff
@@ -6,14 +6,14 @@
...
-        <AtomType name="Ar" element="Ar" charge="0.0" mass="39.948" definition="Ar" description="Argon atom">
+        <AtomType name="Ar" element="Ar" charge="0.0" atomclass="Ar" mass="39.948" definition="Ar" description="Argon atom">
             <Parameters>
                 <Parameter name="A" value="0.1"/>
                 <Parameter name="B" value="4.0"/>
                 <Parameter name="C" value="0.5"/>
             </Parameters>
         </AtomType>
-        <AtomType name="Xe" element="Xe" charge="0.0" mass="131.293" definition="Xe" description="Xenon atom">
+        <AtomType name="Xe" element="Xe" atomclass="Xe" charge="0.0" mass="131.293" definition="Xe" description="Xenon atom">
```
justinGilmer, line 16: per discussion in call, we should change atomclass to class, as well as the kwarg in `AtomType` too

daico007 (approved):

justinGilmer (approved):

gmso/tests/files/ff-example0.xml
umesh-timalsina: See #445

Issue #452, opened 2020-09-29
Lammpsdata writer bug 


daico007:
- The lammpsdata writer is missing an import statement to do the dihedral conversion.
- Also, need to change the trigger at line 174 from comparing name to comparing expression.

PR #453, opened 2020-09-30
Fix bug found in lammps writer

daico007:
Relate to this issue #452. Added missing import statement. Changed from comparing dihedral_type.name to comparing dihedral_type.expression for more general case. Add extra test for lammps writer (typed ethane system)

### DIFF
--- a/gmso/formats/lammpsdata.py
+++ b/gmso/formats/lammpsdata.py
@@ -17,7 +17,7 @@ from gmso.core.topology import Topology
 from gmso.core.box import Box
 from gmso.core.element import element_by_mass
 from gmso.utils.conversions import convert_opls_to_ryckaert, convert_ryckaert_to_opls
-
+from gmso.lib.potential_templates import PotentialTemplateLibrary
 
 def write_lammpsdata(topology, filename, atom_style='full'):
     """Output a LAMMPS data file.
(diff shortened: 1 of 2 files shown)


rmatsum836 (approved):
LGTM, Thanks for fixing this!

gmso/formats/lammpsdata.py
```diff
@@ -168,16 +170,18 @@ def write_lammpsdata(topology, filename, atom_style='full'):
...
+                    if (dihedral_type.expression == sympify(rbtorsion.expression) or
+                        dihedral_type.name == rbtorsion.name):
                         dihedral_type = convert_ryckaert_to_opls(dihedral_type)
                     data.write('{}\t{:.5f}\t{:5f}\t{:5f}\t{:.5f}\n'.format(
                         idx+1,
-                        dihedral_type.parameters['k1']/2,
-                        dihedral_type.parameters['k2']/2,
-                        dihedral_type.parameters['k3']/2,
-                        dihedral_type.parameters['k4']/2
+                        dihedral_type.parameters['k1'].in_units(u.Unit('kcal/mol')).value,
```
justinGilmer, line 181: The docstring says it supports different `atom_style`s other than `full`. Is that true in this case, since we are converting to `kcal/mol`?

justinGilmer (changes requested):
Just need to verify this one question of mine, then it is good to go

gmso/formats/lammpsdata.py
daico007: @rmatsum836

rmatsum836: Yes, the `atom_style` defines the information that gets written out to the data file: https://lammps.sandia.gov/doc/atom_style.html.  Currently we only support the `real` style of units in lammps which uses energy units of `kcal`.

justinGilmer (approved):
LGTM

PR #460, opened 2020-10-02
Ensure that duplicate connections are pruned

uppittu11:
Same as PR #384. 

Makes sure that there is only one connection between any equivalent sets of connection members (e.g. you can't have a bond between atoms 1 and 2 and another one between atoms 2 and 1 since they are the same). This will be useful for applying layered potentials in future PRs.

### DIFF
--- a/gmso/core/bond.py
+++ b/gmso/core/bond.py
@@ -52,7 +52,7 @@ class Bond(Connection):
         where i and j are the connection members.
         """
         return frozenset([
-                tuple(self.connection_members),
+                self.connection_members,
                 tuple(reversed(self.connection_members))
                 ])
 
@@ -70,10 +70,10 @@ class Bond(Connection):
         Here, i and j are interchangeable.
         """
 
-        return hash(tuple([
+        return hash(
             frozenset([self.connection_members[0],
                       self.connection_members[1]])
-            ]))
+            )
 
     def __setattr__(self, key, value):
         if key == 'connection_type':
--- a/gmso/core/bond.py
+++ b/gmso/core/bond.py
@@ -39,6 +39,42 @@ class Bond(Connection):
         # ToDo: Deprecate this?
         return self.__dict__.get('bond_type_')
 
+    def equivalent_members(self):
+        """Get a set of the equivalent connection member tuples
+        Returns
+        _______
+        frozenset
+            A unique set of tuples of equivalent connection members
+        Notes
+        _____
+        For a bond:
+            i, j == j, i
+        where i and j are the connection members.
+        """
+        return frozenset([
+                tuple(self.connection_members),
+                tuple(reversed(self.connection_members))
+                ])
+
+    def _equivalent_members_hash(self):
+        """Returns a unique hash representing the connection
+        Returns
+        _______
+        int
+            A unique hash to represent the connection members
+        Notes
+        _____
+        For a bond:
+            i, j == j, i
+        where i and j are the connection members.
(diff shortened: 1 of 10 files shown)


umesh-timalsina:
Can you check if the index in 0? This test fails sometimes when the index is 0.

uppittu11:
Nevermind. It looks like it passed in CI

justinGilmer (approved):
Changes LGTM! One question though, this seems like it will be a quiet thing that occurs, which is fine. But we might want to consider if this should raise a warning or not. If someone tried to add multiple bonds because they were trying to emulate double or triple bonds, it would appear to work, but only ever show one connection

rsdefever (approved):
Two questions: (1) Why do we `hash(tuple(frozenset(things)))` rather than `hash(frozenset(things))`? 

and (2) right now this evaluates to `False`. Is that the desired/correct behavior?

```python
import gmso
from gmso.core.atom import Atom
from gmso.core.bond import Bond
atom1 = Atom(name="atom1")
atom2 = Atom(name="atom2")

bond1 = Bond(connection_members=[atom1, atom2])
bond2 = Bond(connection_members=[atom2, atom1])
bond1==bond2
```

Other than those questions, the changes look good. Unit tests look thorough. Tests passing locally. LGTM!

gmso/core/bond.py
```diff
@@ -39,6 +39,42 @@ def connection_type(self):
...
+        frozenset
+            A unique set of tuples of equivalent connection members
+        Notes
+        _____
+        For a bond:
+            i, j == j, i
+        where i and j are the connection members.
+        """
+        return frozenset([
+                tuple(self.connection_members),
```
umesh-timalsina, line 55: Same comment as above.

umesh-timalsina (changes requested):
As pointed by @rsdefever. The above equality check can be can be achieved by overriding the `__eq__` dunder here. 

```py
class Bond(Connection):
     ....

    def __eq__(self, other):
        if self is other:
            return True
        if self.name == other.name and self._equivalent_member_hash() = other._equivalent_member_hash():
            return True
       else:
            return False
```

uppittu11:
@rsdefever and @umesh-timalsina I think we still want to do ID based hashing/equality for bonds. IIRC we decided that all "physical" objects such as atoms and connections should be hashed via ID. But only when adding to a Topology, where you have a collection of physical objects, you need to make sure you don't have any duplicates. But I'm open to changing this if it is a better option.

rsdefever:
>> IIRC we decided that all "physical" objects such as atoms and connections should be hashed via ID.

IIRC, you are correct. Sorry for trying to throw you off. I didn't realize that this is also `False`:

```python
import gmso
from gmso.core.atom import Atom
from gmso.core.bond import Bond
atom1 = Atom(name="atom1")
atom2 = Atom(name="atom2")

bond1 = Bond(connection_members=[atom1, atom2])
bond2 = Bond(connection_members=[atom1, atom2])
bond1==bond2
```

I still don't understand this:
>> (1) Why do we hash(tuple(frozenset(things))) rather than hash(frozenset(things))?

Apologies if I'm asking an ignorant question.

uppittu11:
@rsdefever Just fixed the `hash(tuple(frozenset(things)))` for the case of bonds and dihedrals. For angles and impropers there are atoms with fixed positions (i.e. not interchangeable with other atoms) so instead of being in a `frozenset` they are stored in a `tuple`, which is ordered, before being hashed.

PR #469, opened 2020-10-08
Cache unit conversions when hashing unyt quantities

uppittu11:
From previous profiling studies, we found that the `unyt_to_hashable` function is called many times and uses most of the time when `update_topology` is called. Here I made a dictionary cache to store conversion factors to speed up subsequent unit conversions after the first. Benchmarks for `tip3p` water shows that the `update_topology` command is ~10x faster with caching compared to the current master branch.

image


gmso/utils/misc.py
```diff
@@ -1,29 +1,31 @@
+from functools import lru_cache
 import unyt as u
 from unyt.exceptions import UnitConversionError
 
 
+#UNYT_CACHE = dict()
```
justinGilmer, line 6: can remove this

gmso/utils/misc.py
```diff
@@ -1,29 +1,31 @@
...
     """Convert a (list of) unyt array or quantity to a hashable tuple."""
     if unyt_or_unyt_iter is None:
         return unyt_or_unyt_iter
-    hash_coll = []
+
     if isinstance(unyt_or_unyt_iter, list):
-        for val in unyt_or_unyt_iter:
-            hash_coll.append(_unyt_to_hashable_single(val))
-        return tuple(hash_coll)
+        hash_coll = tuple(_unyt_to_hashable_single(val) for val in unyt_or_unyt_iter)
```
justinGilmer, line 14: For future developers sake, can we rename `hash_coll` to `hash_collision` or `collision`? If someone is less familiar with hashing and hashmaps, this might help.

justinGilmer (changes requested):
Two small changes/suggestions. Let me know what you think of the variable naming choice.

gmso/utils/misc.py
umesh-timalsina: Is this supposed to mean `collection` rather than `collision`?

justinGilmer: I think you are correct, @umesh-timalsina. In that case, `collection` definitely makes more sense to me.

umesh-timalsina (approved):
Looks Good. LGTM. Thanks @uppittu11.

PR #470, opened 2020-10-08
Improve performance of from_mbuild

uppittu11:
The `from_mbuild` conversion is very costly. As per previous discussions, we hypothesized that this was because we are repeatedly and unnecessarily running `update_types`/`update_connections`/etc. commands which loop through every atom (O(N^2) complexity). Here I'm trying to reduce the number of calls to cut down the time it takes to convert from mbuild by applying the `update_types=False` flag when adding sites and connections. As of this first commit, here are the performance gains for `from_mbuild` conversions for boxes of tip3p water.

image


gmso/core/subtopology.py
```diff
@@ -70,17 +70,23 @@ def parent(self, parent):
...
+        update_types parameter is set to true (default
+        behavior), this method will also check if there
+        is an gmso.AtomType associated with the site and
+        it to the sub-topology's AtomTypes collection.
 
         Parameters
         ----------
         site : gmso.Atom
             The site to be added to this sub-topology
+        update_types : (bool), default=True
```
daico007, line 88: This is just nitpicking, but can you remove the parentheses (so just `update_types : bool, default=True`) just so we are consistent with other method.

daico007 (approved):
LGTM! Only some nitpicking with the docstring.

justinGilmer (approved):

PR #471, opened 2020-10-08
Prevent a nondeterministic AssertionError in unit tests

uppittu11:
Sometimes the `test_topology_get_index_bond_type` fail randomly. This _might_ be due to dictionaries/sets not being ordered (IDK if this is there reason) causing the index of the last BondType to be 0, which causes an assertion error. As of the first commit to this PR, the tests would no longer fail but the nondeterministic nature of indexing Atom/ConnectionTypes is not addressed. If anyone can figure this out, please push to this PR (or start a new one).

### DIFF
--- a/gmso/tests/test_topology.py
+++ b/gmso/tests/test_topology.py
@@ -472,7 +472,9 @@ class TestTopology(BaseTest):
 
     def test_topology_get_index_bond_type(self, typed_methylnitroaniline):
         assert typed_methylnitroaniline.get_index(typed_methylnitroaniline.bonds[0].connection_type) == 0
-        assert typed_methylnitroaniline.get_index(typed_methylnitroaniline.bonds[-1].connection_type)
+        assert isinstance(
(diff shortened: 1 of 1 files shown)


justinGilmer:
I think it is due to the `types` objects being regular sets and not indexed sets: https://github.com/mosdef-hub/gmso/blob/c21e70b64d4f2d6349d7aae4bbd00d3f41934a6e/gmso/core/topology.py#L595

https://github.com/mosdef-hub/gmso/blob/c21e70b64d4f2d6349d7aae4bbd00d3f41934a6e/gmso/core/topology.py#L144

uppittu11:
The index is coming from a dictionary called `_bond_types_idx `, which stores each BondType mapped to its index (based on the order it was added). So I think the randomness arises prior to or during the conversion from parmed. 

https://github.com/mosdef-hub/gmso/blob/c21e70b64d4f2d6349d7aae4bbd00d3f41934a6e/gmso/core/topology.py#L602

daico007 (approved):

PR #472, opened 2020-10-09
Add conversions to/from networkx graphs

uppittu11:
In this PR I added a very simple conversion to and from networkx graphs. At the moment, only site and bond data is transferred when converting. Angles, dihedrals, impropers, subtops, metadata (mixing rules), etc. are ignored making it a lossy conversion to networkx. Let me know how this could be improved.

### DIFF
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -41,7 +41,7 @@ def from_networkx(graph):
     node_mapping = dict()
     for node in graph.nodes:
         if not isinstance(node, Site):
-            raise GMSOError("Nodes must be gmso.core.Site objects")
+            raise GMSOError("Nodes must be instances of gmso.abc.Site")
         else:
             top.add_site(node)
 
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -32,6 +32,10 @@ def from_networkx(graph):
     object it was created from
     """
 
+    if not isinstance(graph, nx.Graph):
+        raise TypeError("Type mismatch, graph object is expected to be "
+                        "an instance of networkx.Graph, was provided: {}"
+                        .format(type(graph)))
     top = Topology()
 
     node_mapping = dict()
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -41,7 +41,7 @@ def from_networkx(graph):
     node_mapping = dict()
     for node in graph.nodes:
         if not isinstance(node, Site):
-            raise GMSOError("Nodes must be instances of gmso.abc.Site")
+            raise TypeError("Nodes must be instances of gmso.abc.Site")
         else:
             top.add_site(node)
 
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -78,6 +78,9 @@ def to_networkx(top):
 
     graph = nx.Graph()
 
+    for n in top.sites:
+        graph.add_node(n)
+
     for b in top.bonds:
         graph.add_edge(b.connection_members[0], b.connection_members[1], connection=b)
 
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -1,0 +1,84 @@
+import networkx as nx
+
+from gmso.exceptions import GMSOError
+from gmso.abc.abstract_connection import Connection
+from gmso.core.bond import Bond
+from gmso.abc.abstract_site import Site
+from gmso.core.topology import Topology
+
+
+def from_networkx(graph):
+    """Convert a networkx.Graph to a gmso.Topology
+
+    Creates a topology from the graph where each node is a site and each
+    edge becomes a connection.
+
+    Parameters
+    ----------
+    graph : networkX.Graph
+        networkx.Graph instance that need to be converted
+
+    Returns
+    -------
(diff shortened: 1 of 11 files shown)


rsdefever:
What is the use case for `from_networkx`? I can see the value of  `to_networkx` graph from the topology (as a more abstract representation of the connectivity), but why support the inverse?

justinGilmer:
Personally, the `from_networkx` functionality provides another way to create topologies, from a much more abstract representation of a molecule. Especially if you wanted to iterate through various bonded variations it can sometimes be easier to use a graph-based algorithm.

Although similar things could be done with smiles strings, I want to note.

We can also export a lot of compounds and such as bond graphs, which are a bit more general and can be visualized using a plethora of graph visualization libraries like graphviz. Being able to store these as graph objects and then convert back to a topology can be beneficial.

rsdefever:
> Personally, the `from_networkx` functionality provides another way to create topologies, from a much more abstract representation of a molecule. Especially if you wanted to iterate through various bonded variations it can sometimes be easier to use a graph-based algorithm.

That is a great point, thanks for helping me understand.

gmso/external/convert_networkx.py
```diff
@@ -0,0 +1,84 @@
...
+    Notes
+    -----
+    - While a lot of information is lost from converting to a graph object
+    (e.g. metadata, mixing rules, etc.), the graph representation is a
+    useful way to manipulate and extract connectivity information from
+    Topology objects.
+    - The edge has a `connection` attribute, which stores the Bond
+    object it was created from
+    """
+
```
justinGilmer, line 34: I think we should have a validation check here to either ensure it is a `networkx` graph object, or try to coerce into a `networkx` graph

gmso/external/convert_networkx.py
```diff
@@ -0,0 +1,84 @@
...
+    -----
+    While a lot of information is lost from converting to a graph object
+    (e.g. metadata, mixing rules, etc.), the graph representation is a
+    useful way to manipulate and extract connectivity information from
+    Topology objects.
+    """
+
+    graph = nx.Graph()
+
+    for b in top.bonds:
```
justinGilmer, line 81: What about the case where a top has no bonds, they would just be edgeless nodes. Would that work?
I dont think a node would be created based on the logic.

Also, if there were two ethanes in the system, would the `Graph` contain two separate graphs within it, or two sets of connected nodes?

gmso/tests/test_convert_networkx.py
```diff
@@ -0,0 +1,29 @@
...
+from gmso.core.subtopology import SubTopology as SubTop
+from gmso.core.atom import Atom
+from gmso.external.convert_networkx import from_networkx, to_networkx
+from gmso.tests.base_test import BaseTest
+
+class TestConvertNetworkX(BaseTest):
+    def test_to_networkx_ethane(self, ethane):
+        ethane_to_nx = to_networkx(ethane)
+
+        assert ethane.n_sites == ethane_to_nx.number_of_nodes()
```
justinGilmer, line 15: these are great tests, if we can add in the possibilities above, this is good to go afterwards

justinGilmer (changes requested):
Some small suggestions and a few edge cases (pun _not_ intended)

RainierBarrett:
One slightly hacky way to preserve some more information than just the molecular structure would be to make an unconnected sub-graph with one node, and just make a rule for how that node's weight will hold the metadata etc. that is currently being lost... Any thoughts?

justinGilmer (changes requested):
Looks like the unit tests handle pretty much every case I am concerned with, my only change would be in the from_networkx method. I think we still need a check to ensure that the graph object is a nx graph. Although that could be abstracted out to a separate method like: _valdidate_graph. Although I am not sure that is necessary. Other than that, this is good to merge for me.

gmso/external/convert_networkx.py
```diff
@@ -0,0 +1,87 @@
...
+    - The edge has a `connection` attribute, which stores the Bond
+    object it was created from
+    """
+
+    top = Topology()
+
+    node_mapping = dict()
+    for node in graph.nodes:
+        if not isinstance(node, Site):
+            raise GMSOError("Nodes must be gmso.core.Site objects")
```
umesh-timalsina, line 40:
```suggestion
            raise GMSOError("Nodes must be gmso.core.Site objects")
            raise GMSOError("Nodes must be instances of gmso.abc.Site")
```

umesh-timalsina (approved):

justinGilmer (approved):

CalCraven:
If you import gmso.external, then you can only see openmm, mbuild, and parmed converters. You have to manually import gmso.external.convert_networkx to get this functionality. Probably worth it to make that converter automatically import when importing gmso.external. 

Also, but just a side note: you have to separately import gmso.external already. Not sure if that is intended or something we want to have available during the initial gmso import.

umesh-timalsina:
```python
(gmso-dev) umesh@isisdell:~$ python
Python 3.6.10 | packaged by conda-forge | (default, Apr 24 2020, 16:44:11) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from gmso.external import convert_networkx
>>> 

```
I am able to import it like the snippet above. @CalCraven, could you point a code snippet to reproduce the issue you mentioned.

Issue #464, opened 2020-10-02
Add an attribute called member_classes for bonded potentials


umesh-timalsina:
After #442, we can also create bonded potentials based on atom classes, So, it might be worth it to add `member_classes` to the bonded potentials. This will be required for serialization of a `ForceField` for example.

umesh-timalsina:
The tricky part might be when there's a ForceField that has a bond defined between an atomClass  and atomType

uppittu11:
How is this handled in foyer? Is there a `member_classes` attribute?

justinGilmer:
The conclusion currently is that all of these attributes like atomType vs. atomClass

Or bonds being between classes vs types, should all be moved to "class" attributes, if you want the specificity for a singular atomtype, make the atomclass of that type a more specific name

umesh-timalsina:
Closed by https://github.com/mosdef-hub/gmso/issues/464#issuecomment-705720449

PR #478, opened 2020-10-22
Donot fail in XML parsing if atomclass present

### LINKED ISSUE #477: _check_atomtype_existence should also check for atomclass when parsing a GMSO XML.
Currently, creation of `ForceField` from a GMSO XML raises an error if there's  Potential based on `atomclass`. This function should also check if there's an atomclass that corresponds to the connection potential atomclasses.
https://github.com/mosdef-hub/gmso/blob/595a6f17a163296d4ffed58bb403ff4413f967c2/gmso/utils/ff_utils.py#L88-L101

umesh-timalsina:
Closes #477.

### DIFF
--- a/gmso/utils/ff_utils.py
+++ b/gmso/utils/ff_utils.py
@@ -99,7 +99,7 @@ def _check_atomtype_existence(tag, ref_dict):
         if member == '*':
             continue
         if member not in ref_dict and member not in atom_classes_set:
-            raise ForceFieldParseError('AtomTypes {} not present in AtomTypes reference in the xml'.format(member))
+            raise ForceFieldParseError('AtomType/AtomClass {} not present in AtomTypes reference in the xml'.format(member))
     return member_types
 
 
--- a/gmso/utils/ff_utils.py
+++ b/gmso/utils/ff_utils.py
@@ -91,12 +91,14 @@ def _check_atomtype_existence(tag, ref_dict):
     at3 = tag.attrib.get('type3', tag.attrib.get('class3', None))
     at4 = tag.attrib.get('type4', tag.attrib.get('class4', None))
 
+    atom_classes_set = set(value.atomclass for value in ref_dict.values())
+
     member_types = list(filter(lambda x: x is not None, [at1, at2, at3, at4]))
     member_types = ['*' if mem_type == '' else mem_type for mem_type in member_types]
     for member in member_types:
         if member == '*':
             continue
-        if member not in ref_dict:
+        if member not in ref_dict and member not in atom_classes_set:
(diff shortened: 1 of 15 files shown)


rmatsum836:
Loading in the XML you provided (`PFE_gmso.xml`) worked for me using this PR and is showing the correct number of atom types and connection types.  Also one of our test xml files (`ff-example0.xml`) some of the bond types are defined between a mixture of `class` and `type`.  Do we have a preference for how users will define connection types (class1 vs type1)?

It may be a good idea to revisit our GMSO XML files that we use for testing and consolidate/clean them up.

umesh-timalsina:
We currently can have a mix of `atomclass` and `types` to define `ConnectionType` but we decided everything will be linked to an `atomclass`. I agree, it might be nice to port some standard foyer `forcefields` to `gmso` after your PR #438 is merged. See #464 specifically https://github.com/mosdef-hub/gmso/issues/464#issuecomment-705720449.

gmso/utils/ff_utils.py
```diff
@@ -91,13 +91,15 @@ def _check_atomtype_existence(tag, ref_dict):
     at3 = tag.attrib.get('type3', tag.attrib.get('class3', None))
     at4 = tag.attrib.get('type4', tag.attrib.get('class4', None))
 
+    atom_classes_set = set(value.atomclass for value in ref_dict.values() if value)
+
     member_types = list(filter(lambda x: x is not None, [at1, at2, at3, at4]))
```
justinGilmer, line 96: everything looks good to me, but i have one question about this line. I know its unrelated to the PR itself, but what exactly does this do?

Looks like the lambda function creates a boolean mask and checks if any of the `at1->at4` are `None`. Im not sure exactly what happens afterwards.

justinGilmer (approved):
LGTM, but i added one small question.

gmso/utils/ff_utils.py
umesh-timalsina: A `filter` function takes in a callable and an iterable, and returns a new iterable containing the members of the original iterable for which the callable function evaluates to truthy.

It this case:
1. the lambda function is the callable, [at1, at2, at3, at4] is the iterable. So, it will create a new iterable (casted as a list in the end). So, in this case you could endup with a list containing either 2, 3 or 4, elements, depending on which part of the XML file is being parsed (2 for BondType for example). Basically, the whole filter/list is a syntactic sugar for:
```py
member_types = []
for at in [at1, at2, at3, at4]:
    if at is not None:
        member_types.append(at)
```

justinGilmer: Oh neat! Makes sense! Thanks for the clarification : )

daico007 (approved):

PR #484, opened 2020-11-12
New pair potential implementation in GMSO backend

zijiewu3:
**PR Summary**  

This is the new implementation of `PairPotentialType` that allows storage of custom pair potentials between `AtomTypes` that do not follow combination rules in the backend. (Following closed #382)
Here is an [example](https://gist.github.com/zijiewu3/dfde5262e49db5ade09234b0611b7583) showcasing how this PR works.
1. `PairPotentialType` is a class inheriting from `ParametricPotential` class that describes customized potential form between 2 (for now, TODO: expand for multi-body interactions) `AtomType`s.
1. In a `Topology`, all `PairPotentialTypes` are stored in a dictionary `_pairpotential_type` that uses **frozensets of its member `AtomTypes`** as the keys.  The general idea is to keep the pair potentials standalone from the `connection`-`potential` duo for bonds, angles, dihedrals, etc, since pair potentials are w.r.t. `AtomType`s, while the bonded potentials are w.r.t. specific sites.
1. Included simple unit tests for adding and removing `PairPotentialTypes` to a `Topology`.


zijiewu3:
Added a gist to an example on Jupyter Notebook.

gmso/core/pairpotential_type.py
```diff
@@ -0,0 +1,71 @@
...
+import unyt as u
+
+from pydantic import Field
+
+from gmso.core.parametric_potential import ParametricPotential
+from gmso.utils._constants import PAIRPOTENTIAL_TYPE_DICT
+from gmso.core.atom_type import AtomType
+
+class PairPotentialType(ParametricPotential):
+    __base_doc__ = """A descripton of custom pairwise potential between 2 AtomTypes that does not follow combination rule.
```
umesh-timalsina, line 11:
```suggestion
    __base_doc__ = """A description of custom pairwise potential between 2 AtomTypes that does not follow combination rule.
```

gmso/core/topology.py
```diff
@@ -465,7 +482,47 @@ def update_connection_types(self):
                     c.connection_type = self._dihedral_types[c.connection_type]
                 if isinstance(c.connection_type, ImproperType):
                     c.connection_type = self._improper_types[c.connection_type]
+    def add_pairpotentialtype(self, pairpotentialtype):
```
umesh-timalsina, line 485: Just a comment, this method adds a `PairPotentialType` between two `AtomType`s only when they are in the topology. It might be nice to add a flag to `update_atom_types` just to provide and option to update the newly added Atom's  `atomtype` in the topology.

gmso/core/pairpotential_type.py
```diff
@@ -0,0 +1,71 @@
...
+
+    This is a subclass of the gmso.core.Potential superclass.
```
umesh-timalsina, line 13:
```suggestion
    This is a subclass of the gmso.core.ParametricPotential superclass.
```

umesh-timalsina (changes requested):
Looks good, couple of minor suggestions and few questions:

1. Why is there a file called `gmso/core/pairpotential.py`?
2. Should the added `PairPotentialType` be indexed? If so, the `get_index` method in `Topology` should be extended to incorporate `PairPotentialType` as well.

gmso/core/topology.py
zijiewu3: Addressed most of the feedback from Umesh.


> 2. Should the added `PairPotentialType` be indexed? If so, the `get_index` method in `Topology` should be extended to incorporate `PairPotentialType` as well.

Regarding whether `PairPotentialType` should be indexed, I wonder if it is needed? we can already directly access a `PairPotentialType` by pairs of member `AtomType`s. As I am not fully sure about when to or not to index, please feel free to let me know your thought about this.

zijiewu3:
Addressed most of the feedback from Umesh.


> 2. Should the added `PairPotentialType` be indexed? If so, the `get_index` method in `Topology` should be extended to incorporate `PairPotentialType` as well.

Regarding whether `PairPotentialType` should be indexed, I wonder if it is needed? we can already directly access a `PairPotentialType` by pairs of member `AtomType`s. As I am not fully sure about when to or not to index, please feel free to let me know your thought about this.

gmso/core/pairpotential_type.py
```diff
@@ -0,0 +1,71 @@
...
+    PairPotentialType represents a type of pairwise potential between two 
+    Atomtypes that does not follow combination rule, and includes the functional 
+    form describing its interactions. The functional form of the potential is 
+    stored as a `sympy` expression and the parameters, with units, are stored
+    explicitly.  The AtomTypes that are used to define the dihedral type are
+    stored as `member_types`.
+
+
+    Notes
+    ----
```
uppittu11, line 24:
```suggestion
    -----
```

uppittu11 (commented):
Looks good so far! Had a couple points to put out there that could either be features or bugs depending on what y'all want to use this function for. 
* So it seems that you can't add custom pair potentials between specific atoms so this won't fix the problem with the CHARMM force field, where specific 1-4 interactions have custom potentials. One solution is to create a `Pair` object that functions similar to a `Bond`, `Angle`, etc. to which we could assign a `PairPotentialType`. 
* At the moment you can't define pair potentials between atom classes (like we can for bonds, angles, etc.). If we want to allow that then you'd need a different way to index them in the dictionary instead of using `member_types`. 
* We can't assign multiple potentials to a single pair. IDK why you'd do that in practice, but it is something that (I think) is possible in most simulation engines.
* A note for the future: custom pair should override the regular mixing rules, which I'm guessing would be handled when writing out the file. We should probably keep that in mind when we start extending this functionality to the file writers.

uppittu11:
> Addressed most of the feedback from Umesh.
> 
> > 1. Should the added `PairPotentialType` be indexed? If so, the `get_index` method in `Topology` should be extended to incorporate `PairPotentialType` as well.
> 
> Regarding whether `PairPotentialType` should be indexed, I wonder if it is needed? we can already directly access a `PairPotentialType` by pairs of member `AtomType`s. As I am not fully sure about when to or not to index, please feel free to let me know your thought about this.

I think it should be indexed so the order that the potentials show up in the file will be the same every time you write it out. That would make for easier comparison. Either way, it doesn't hurt to index them since it wouldn't noticeably affect efficiency (correct me if I'm wrong on this).

zijiewu3:
first of a series of commits, this one addressing the using of `strs` as `member_types` for `PairPotentialTypes`

zijiewu3:
Finally! Modified the XML schema and parsing files so that PairPotentialType info can be included in the gmso-xml files.
Will include an example and test cases soon

gmso/core/pairpotential_type.py
```diff
@@ -0,0 +1,70 @@
...
+from gmso.core.parametric_potential import ParametricPotential
+from gmso.utils._constants import PAIRPOTENTIAL_TYPE_DICT
+
+class PairPotentialType(ParametricPotential):
+    __base_doc__ = """A description of custom pairwise potential between 2 AtomTypes that does not follow combination rule.
+
+    This is a subclass of the gmso.core.ParametricPotential superclass.
+
+    PairPotentialType represents a type of pairwise potential between two 
+    Atomtypes that does not follow combination rule, and includes the functional
```
justinGilmer, line 15:
```suggestion
    Atomtypes that does not follow a specific combination rule, and includes the functional 
```

gmso/core/topology.py
```diff
@@ -491,7 +510,57 @@ def update_connection_types(self):
...
+        if not isinstance(pairpotentialtype,PairPotentialType):
+            raise GMSOError('Non-PairPotentialType {} provided'.format(pairpotentialtype))
+        for atype in pairpotentialtype.member_types:
+            if atype not in [t.name for t in self.atom_types]:
+                if atype not in [t.atomclass for t in self.atom_types]:
+                    raise GMSOError('There is no name/atomclass of AtomType {} in current topology'.format(atype))
+        self._pairpotential_types[pairpotentialtype] = pairpotentialtype
+        self._pairpotential_types_idx[pairpotentialtype] = len(self._pairpotential_types) - 1
+
+    def remove_pairpotentialtype(self,pair_of_types):
```
justinGilmer, line 543:
```suggestion
    def remove_pairpotentialtype(self, pair_of_types):
```

gmso/utils/ff_utils.py
```diff
@@ -336,6 +340,42 @@ def parse_ff_connection_types(connectiontypes_el, child_tag='BondType'):
 
     return connectiontypes_dict
 
+def parse_ff_pairpotential_types(pairpotentialtypes_el):
+    """Given an XML etree Element rooted at PairPotentialTypes, parse the XML to create topology.core.PairPotentialTypes,"""
```
justinGilmer, line 344:
```suggestion
    """Given an XML etree Element rooted at PairPotentialTypes, parse the XML to create topology.core.PairPotentialTypes."""
```

justinGilmer (changes requested):
Aside from some small style nits, this looks good! Thanks @zijiewu3 !

gmso/core/topology.py
```diff
@@ -563,6 +585,66 @@ def update_connection_types(self):
...
+        self._pairpotential_types_idx[pairpotentialtype] = (
+            len(self._pairpotential_types) - 1
+        )
+
+    def remove_pairpotentialtype(self, pair_of_types):
+        """Remove the custom pairwise potential between two AtomTypes/Atomclasses
+
+        Parameters
+        ----------
+        pair_if_types: list-like of strs
```
umesh-timalsina, line 631:
```suggestion
        pair_of_types: list-like of strs
```

gmso/utils/ff_utils.py
```diff
@@ -398,6 +402,56 @@ def parse_ff_connection_types(connectiontypes_el, child_tag="BondType"):
     return connectiontypes_dict
 
 
+def parse_ff_pairpotential_types(pairpotentialtypes_el):
```
umesh-timalsina, line 405: Can this function be removed? And `parse_ff_connection_types` be used and it should be renamed to `parse_ff_potentials`.

umesh-timalsina (changes requested):
Looks Great. Some small suggestions. After that LGTM!!!

Co reviewers: @justinGilmer, @daico007

gmso/utils/ff_utils.py
zijiewu3: Sorry could you explain a bit more what you mean here?

umesh-timalsina: This is not absolutely necessary but the function `parse_ff_connection_types` already does what is acheived by this function. However, an argument could be made that since `PairPotential`s are different from `AngleType`, `BondType`  etc... the usage of this function is justified.

zijiewu3: I see. I personally would vote for keeping these two functions separate as `PairPotential` and `connection_types`, although similar in form, have a crucial difference of one toggled to atomtypes and the other toggled to connections. So I think it makes more sense to deal with them separately while possible. This might also be helpful if in the future we want to send the XML syntax of these two onto slightly branching directions. But on the other hand, you are right that the codes of these two functions are basically identical.

umesh-timalsina: Makes sense

umesh-timalsina (approved):

daico007 (approved):

PR #488, opened 2020-12-10
Import networkx conversion functions in gmso.external

umesh-timalsina:
Pointed out by @CalCraven, fixes networkx functions import issues and adheres to the import conventions of `gmso.external`.


daico007 (approved):
LGTM! Pretty sure test is failing due to other reasons

PR #491, opened 2021-01-09
Include angle and dihedral information on conversion to networkx

CalCraven:
Include during the to_networkx conversion angle and dihedral information. This is stored as `angles` and `dihedrals` Networkx graph attributes for each node. The information stored there is accessible in the same way as we access this information stored within the topology objects (as the Angle class from gmso/core/angle.py and the Dihedrals class from gmso/core/dihedral.py. Therefore, any single node can be selected, and the dihedrals and angles it is an element of can be iterated through and printed. 

Once a graph object is returned through the command:
`Graph = gmso/external/to_networkx(topology)`

The information can be accessed with code such as:
`for node in Graph.nodes:`
    `print(Graph.nodes[node]['angles'])`

And likewise for dihedrals by replacing 'angles' with 'dihedrals'.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -580,6 +580,14 @@ class Topology(object):
         self.update_connection_types()
         self.is_typed(updated=True)
 
+    def _get_bonds_for(self, site):
+        """return a list of the bonds that contain Site"""
+        bonds = []
+        for bond in self.bonds:
+            if site in bond.connection_members:
+                bonds.append(bond)
+        return bonds
+
     def _get_angles_for(self, site):
         """return a list of the angles that contain Site"""
         angles = []
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -580,6 +580,22 @@ class Topology(object):
         self.update_connection_types()
         self.is_typed(updated=True)
 
+    def _get_angles_for(self, site):
+        """return a list of the angles that contain Site"""
+        angles = []
+        for angle in self.angles:
+            if site in angle.connection_members:
+                angles.append(angle)
+        return angles
+
+    def _get_dihedrals_for(self, site):
+        """return a list of the dihedrals that contain Site"""
+        dihedrals = []
+        for dihedral in self.dihedrals:
+            if site in dihedral.connection_members:
+                dihedrals.append(dihedral)
+        return dihedrals
+
     def get_index(self, member):
         """Get index of a member in the topology
 
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -1,3 +1,4 @@
+import warnings
 import networkx as nx
 
 from gmso.exceptions import GMSOError
@@ -39,6 +40,7 @@ def from_networkx(graph):
     top = Topology()
 
     node_mapping = dict()
+
     for node in graph.nodes:
         if not isinstance(node, Site):
             raise TypeError("Nodes must be instances of gmso.abc.Site")
@@ -55,6 +57,12 @@ def from_networkx(graph):
             conn = Bond(connection_members=edge)
             top.add_connection(conn)
 
+    warnings.simplefilter('once', UserWarning) 
+
+    for node in graph.nodes:
+        if graph.nodes[node]['angles'] or graph.nodes[node]['dihedrals']:
+            warnings.warn("Angle and Dihedral information is not converted.") 
+    
     return top
 
 def to_networkx(top):
--- a/gmso/external/convert_networkx.py
+++ b/gmso/external/convert_networkx.py
@@ -60,9 +60,12 @@ def from_networkx(graph):
     warnings.simplefilter('once', UserWarning) 
 
     for node in graph.nodes:
-        if graph.nodes[node]['angles'] or graph.nodes[node]['dihedrals']:
+        try:
+            graph.nodes[node]['angles'] or graph.nodes[node]['dihedrals']:
             warnings.warn("Angle and Dihedral information is not converted.") 
-    
+        except KeyError:
+            pass
+
     return top
 
 def to_networkx(top):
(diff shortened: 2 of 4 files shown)


umesh-timalsina (commented):
Few general comments:
- Why is it the case we need these methods only for `angles` and `dihedrals`? What about `bond` and `angles`?
- Also, we could add some test cases to test these methods.

CalCraven:
> * Why is it the case we need these methods only for `angles` and `dihedrals`? What about `bond` and `angles`?
I added in a method for _get_bonds_from(self,site). I think that covers this question. 

> * Also, we could add some test cases to test these methods.
Added a few basic test cases to tests/test_topology.py and tests/test_convert_networkx.py. Could potentially use some feedback to add a better suite of tests for these functions.

gmso/core/topology.py
```diff
@@ -580,6 +580,30 @@ def update_topology(self):
         self.update_connection_types()
         self.is_typed(updated=True)
 
+    def _get_bonds_for(self, site):
+        """return a list of the bonds that contain Site"""
```
umesh-timalsina, line 584:
```suggestion
        """Return a list of bonds in this Topology that the site is a part of"""
```

gmso/core/topology.py
```diff
@@ -580,6 +580,30 @@ def update_topology(self):
...
+        bonds = []
+        for bond in self.bonds:
+            if site in bond.connection_members:
+                bonds.append(bond)
+        return bonds
+
+    def _get_angles_for(self, site):
+        """return a list of the angles that contain Site"""
```
umesh-timalsina, line 592:
```suggestion
        """Return a list of angles in this Topology that the site is a part of"""
     
```

gmso/core/topology.py
```diff
@@ -580,6 +580,30 @@ def update_topology(self):
...
+        angles = []
+        for angle in self.angles:
+            if site in angle.connection_members:
+                angles.append(angle)
+        return angles
+
+    def _get_dihedrals_for(self, site):
+        """return a list of the dihedrals that contain Site"""
```
umesh-timalsina, line 600:
```suggestion
        """Return a list of dihedrals in this Topology that the site is a part of"""
```

umesh-timalsina (changes requested):
I have suggested some minor changes to the docstrings of the added methods.

umesh-timalsina (approved):

umesh-timalsina:
LGTM!!

gmso/external/convert_networkx.py
```diff
@@ -88,4 +88,10 @@ def to_networkx(top):
     for b in top.bonds:
         graph.add_edge(b.connection_members[0], b.connection_members[1], connection=b)
 
+    for node in graph.nodes:
+        graph.nodes[node]['angles'] = top._get_angles_for(node)
+
+    for node in graph.nodes:
+        graph.nodes[node]['dihedrals'] = top._get_dihedrals_for(node)
+
```
uppittu11, line 96: So if you convert a topology that has bonds, angles, and dihedrals to networkX, you preserve the angle and dihedral info. But if you convert it back to a topology using `from_networkx` you lose that info. Is this the behavior we want or should there be something that parses through the angles and dihedrals in each node and adds them to the topology in `from_networkx`?

umesh-timalsina: Good Catch. One way to do this would be to add a flags something like `include_angles` and `include_dihedrals` in this method and add a warning when converting back to topology from nx.Graph.

CalCraven: Yeah that's something that was on my mind. It definitely seems like that would be possible. However, I think being able to convert to and from networkx to a topology just isn't a serious way to be playing around with these topology objects. I think this networkx method is a way to take a topology and get some information from it. But I think being able to build usable topologies from a networkx graph just isn't the way we want our users importing information into gmso. So it might just not be worth the time unless someone can suggest a use case that is important to them.

umesh-timalsina: Yeah. Not carry extra information that what's needed, I think we can have these node attributes stored conditionally. Basically, these are just references to the original objects inside the topology. So, it might be nice to include them when its absolutely necessary to avoid potential unintended behaviour.

CalCraven: However, if we want to include warnings in the `from_networkx` to make sure users know that information is being lost, that would be real quick and easy.

uppittu11: Warnings would be fine.

> But I think being able to build usable topologies from a networkx graph just isn't the way we want our users importing information into gmso.

If this is the case, we could just get rid of the `from_networkx` function. [Here is what Justin had to say regarding the use cases for `from_networkx`](https://github.com/mosdef-hub/gmso/pull/472#issuecomment-711308217). But it would be just as easy to force people to only use mbuild for building/modifying connectivity though.

justinGilmer: I think warnings are completely fine for this case. 

My one argument for keeping from_networkx approximately the same as it is now is that the extra information can be generated on the fly with `identify_connections`. I think for use cases like these, we don’t need to make a lossless conversion necessary. 

Absolutely open to other thoughts on this though.

daico007: Agree with Justin about keeping this method relatively the same for now. Adding a warning should be sufficient.

gmso/tests/test_convert_networkx.py
```diff
@@ -16,7 +16,12 @@ def test_to_networkx_ethane(self, ethane):
         assert ethane.n_sites == ethane_to_nx.number_of_nodes()
         assert ethane.n_bonds == ethane_to_nx.number_of_edges()
 
+
```
daico007, line 19: unrelated, but can we remove these white spaces?

daico007 (approved):
LGTM! should be merge-able after adding that warning for the `from_networkx` method

uppittu11 (approved):

PR #492, opened 2021-01-10
Functions to plot networkx graphs of topology objects

CalCraven:
This PR builds off of the convert_networkx PR. 

This uses the to_networkx(topology) to plot networkx graphs of the topology. It is broken down into the ability to plot by-
atomtype: `gmso.formats.interactive_networkx_atomtypes`
bonds: `gmso.formats.interactive_networkx_bonds`
angles: `gmso.formats.interactive_networkx_angles`
dihedrals: `gmso.formats.interactive_networkx_dihedrals`

See the attached help strings for how to change the inputs to see different aspects of a topology. 

I considered making these functions a method of the gmso.core.topology.Topology class. However, there are import issues. The gmso/external/convert_networkx.py file imports the topology class so it can create a topology from a networkx object. Since these functions make use of the to_networkx function in this file, either the to_networkx and from_networkx functions need to be separated, or we use this a separate file with these plot_networkx functions. I though it acceptable that they were part of the gmso/formats folder, since networkx is another potential way to output information about these structures.

I'm hoping to continue to add some interactive functionality for working with these networkx objects. This will use `ipywidget.interact`.


gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+
+    return (fig, ax)
+
+
+def interactive_networkx_atomtypes(topology, list_of_labels=None):
+    """Get an interactive networkx plot showing the atom types of a topology object.
+
+    Parameters
+    ----------
+    topology : A gmso.core.topology.Topology object
```
umesh-timalsina, line 46:
```suggestion
    topology : gmso.Topology
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+        This should be a gmso topology object that you want to visualize the atom types
+        that have been parameterized.
+    list_of_labels : Additional labels you would like to have on the plot.
```
umesh-timalsina, line 49:
```suggestion
    list_of_labels : List[str]
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+            Label = What labels are shown on the networkx graph
+            Atom_Name = Which sites will show these labels
+
+    matplotlib.pyplot.figure
+        The drawn networkx plot of the topology on a matplotlib editable figures
+
+    matplotlib.pyplot.axis
+        The axis information that corresponds to that figure. This output can be
+        shown using
+        matplotlib.pyplot.show()
```
umesh-timalsina, line 67: Are there return values justified here?

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+
+    Returns
+    -------
+    ipywidgets.Dropdown()
+        Three dropdown options, corresponding to:
+            Atom1 = The first site for which bonds will be listed. If all sites are selected for both
+                atoms, then only bonds with missing types will be shown.
+            Atom2 = A second site for which bonds will be listed. With both specified
+                only bonds between Atom1 and Atom2 will be shown.
+            list_of_bonds = The bonds that can be selected corresponding to the Atom1 and Atom2
```
umesh-timalsina, line 120: Same comments as above function.

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+def get_edges(networkx_graph, atom_name1, atom_name2):
+
+    # Create a list of the edges that have been selected corresponding to the selected atom_name1
+    # and atom_name2. If both are None, then select edges that are missing bond types.
+    list_of_edges = []
+    list_of_bonds = []
+    # Check for whether you want to plot bonds based on the atom types, or their node labels.
+    try:
+        [node.atom_type.name for node in networkx_graph.nodes]
+    except AttributeError:
```
umesh-timalsina, line 785: Two questions:
1. Why is this returning None?
2. Do you want to fail for even single `atom_type` missing?

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+        return print("Atomtypes are missing, so no bond types can be selected")
+    mia_bond_flag = 0
+    if atom_name1 and atom_name2:
+        for edge in list(networkx_graph.edges):
+            if (
+                edge[0].name == atom_name1
+                and edge[1].name == atom_name2
+                or edge[0].name == atom_name2
+                and edge[1].name == atom_name1
```
umesh-timalsina, line 794: Might be nicer to clarify precedence in this case not to get confused later. Although, and takes precedence over or in this case.
```suggestion
                (edge[0].name == atom_name1
                and edge[1].name == atom_name2)
                or (edge[0].name == atom_name2
                and edge[1].name == atom_name1)
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+        else:
+            selectable_list[label] = []
+            selectable_list[label].append(bond)
+
+    # turn the dic selectable list into a list of tuples.
+    list_of_edges = []
+    for key in selectable_list:
+        list_of_edges.append((key, selectable_list[key]))
+
+    return list_of_edges
```
umesh-timalsina, line 834:
```suggestion
    return tuple(selectable_list.items())
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+    labeled_bonds = zip(list_of_bonds, list_of_edges)
+
+    # create a dic so each selected bond selects the proper edges on the networkx.graph object.
+    selectable_list = {}
+    for label, bond in labeled_bonds:
+        if label in selectable_list.keys():
+            selectable_list[label].append(bond)
+        else:
+            selectable_list[label] = []
+            selectable_list[label].append(bond)
```
umesh-timalsina, line 827: Why not build this dictionary in the above loops?

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+        edge_colors=edge_colors,
+        list_of_labels=list_of_labels,
+        atom_name=atom_name1,
+    )
+    return fig, ax
+
+
+def report_bond_parameters(topology, edge):
+    # return nicely printed bond parameters for a given edge.
+    for bond in list(topology.bonds):
```
umesh-timalsina, line 871: Why is the casting important here? Its a tuple right?
```
    for bond in list(topology.bonds):
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+                if bond.connection_members == edge[0] or bond.connection_members == (
+                    edge[0][1],
+                    edge[0][0],
+                ):
+                    print(bond.bond_type.expression, "\n")
+                    print("{:<12} {:<15}".format("Parameter", "Value"))
+                    for k, v in bond.bond_type.parameters.items():
+                        print("{:<12} {:<15}".format(k, v))
+            except AttributeError:
+                print("This bond has no parameters associated with it")
```
umesh-timalsina, line 883: Could we have a more helpful error message here?
```suggestion
                print("This bond has no parameters associated with it")
```

gmso/formats/networkx.py
```diff
@@ -0,0 +1,953 @@
...
+        else:
+            print("This bond has no parameters associated with it")
```
umesh-timalsina, line 885: Same comment as above.

umesh-timalsina (changes requested):
Some comments regarding usage of the netowrkx graph/functions

gmso/formats/networkx.py
CalCraven: I wasn't quite sure how to handle this. It does plot the interactive ipywidget with these features, but doesn't explicitly return these objects. Is there a better way to annotate what running this function would give the user without using the return section of the docstring?

gmso/formats/networkx.py
CalCraven: I think that is the intended behavior. Since the main use case I was looking at was for reporting missing objects object parameters, I wanted these to be immediately reported. However, I could see a use where you're still trying to check a bond even when an atomtype is missing. Which wouldn't be possible with this line of code. I think in this case, I'll write a line of code where it only checks for missing atomtypes if that's the actual bond, angle, or dihedral you're looking for. In that case, it won't be able to show it because their is no name for that site.

gmso/formats/networkx.py
CalCraven: This more or less just takes the zipped lists of edges and the corresponding name of that edge, and just thins it out so repeats are not shown. Otherwise, each bond that had the same label would be shown in the dropdown selection, and would clutter it up with repeats. So the above loops generate the full list, and the second loop shown here goes through and finds only unique examples to put in the dictionary. I could simplify this using `numpy.unique()` I think.

gmso/formats/networkx.py
CalCraven: Fixed.

gmso/formats/networkx.py
CalCraven: Added information about the missing edge.

gmso/formats/networkx.py
CalCraven: Deleted this line as it is unnecessary. If the bonds are missing in the object, earlier functions will throw up errors.

gmso/formats/networkx.py
CalCraven: Gonna delete this try statement and check to see what happens if one of the nodes is missing an atomtype.

CalCraven: So now we just see a missing atomtype on the networkx plot for atomtypes, and when plotting bonds and angles, you can select a particular bond, but it is blank for that particular site.

jennyfothergill:
Great work :) I feel like this PR could be adapted to also address these issues:
https://github.com/mosdef-hub/foyer/issues/287 - visualize atomtyping
https://github.com/mosdef-hub/foyer/issues/258 - add readme image (your visualization would be great here, see mwthompsons draft image)
https://github.com/mosdef-hub/foyer/issues/169 - report unparameterized angles, etc
https://github.com/mosdef-hub/foyer/issues/323 - roughly the same as 169

CalCraven:
> [mosdef-hub/foyer#287](https://github.com/mosdef-hub/foyer/issues/287) - visualize atomtyping
I would say this PR covers this use case. Is there a way to close the issue by linking to this PR.
>   [mosdef-hub/foyer#258](https://github.com/mosdef-hub/foyer/issues/258) - add readme image (your visualization would be great here, see mwthompsons draft image)
I can take a sample system and add the visualization to the README for Foyer or GMSO. Honestly, I think it might be useful to have a single image that goes on both README's, that shows the connection between the two packages.
> [mosdef-hub/foyer#169](https://github.com/mosdef-hub/foyer/issues/169) - report unparameterized angles, etc
This also addresses this issue! If you plot using networkx bonds, angles, or dihedrals and don't specify which atoms you're interested in getting a list of parameters for, the unparameterized connections will be shown.
> [mosdef-hub/foyer#323](https://github.com/mosdef-hub/foyer/issues/323) - roughly the same as 169
Again, these needs should be addressed with this PR

CalCraven (commented):
Umesh reviews addressed

CalCraven (commented):
Applied Umesh suggestions.

PR #498, opened 2021-01-29
Make checking atom types optional in ForceField Creation

### LINKED ISSUE #497: Should checking atom type existence when creating a ForceField be Optional ?
It looks like a force field  can contain arbitrary `bond` / `angle` / `dihedral` types even if the atom `types`/`classes` defining the interaction between the `bonds` / `angles` / `dihedrals` do not exist in the `forcefield`. It might be better to get rid of this check or make this check optional at the very least.

Related Issues/ PRs:
Foyer XML Conversion: #402 
GMSO Integration with foyer: https://github.com/mosdef-hub/foyer/pull/358

umesh-timalsina:
Closes #497.

### DIFF
--- a/gmso/tests/test_forcefield_xml.py
+++ b/gmso/tests/test_forcefield_xml.py
@@ -8,7 +8,7 @@ from lxml.etree import DocumentInvalid
 from gmso.core.forcefield import ForceField
 from gmso.tests.utils import get_path
 from gmso.tests.base_test import BaseTest
-from gmso.exceptions import ForceFieldParseError
+from gmso.exceptions import ForceFieldParseError, MissingAtomTypesError
 
 
 class TestForceFieldFromXML(BaseTest):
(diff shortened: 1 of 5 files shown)


gmso/tests/test_forcefield_xml.py
```diff
@@ -200,3 +200,7 @@ def test_potential_types_by_expression(self, named_groups_ff):
         assert len(dihedral_types_grouped_by_expression['0.5*z*(r - r_eq)**2']) == 2
         assert len(improper_types_gropued_by_expression['0.5*z*(r - r_eq)**2']) == 1
 
+    def test_forcefield_missing_atom_types(self):
+        with pytest.raises(MissingAtomTypesError):
+            ff = ForceField(get_path(filename=get_path('ff_missing_atom_types.xml')))
```
daico007, line 205: can we add an exact same line but with `strict=False` just to make sure the forcefield will load.

daico007 (approved):
Just a tiny comment, otherwise LGTM!

gmso/tests/test_forcefield_xml.py
umesh-timalsina: I have added the test in 0a89771.

PR #500, opened 2021-02-02
Add doi conversion from foyer XML

daico007:
We are current not translating doi information when converting foyer XML to GMSO XML. This PR add that conversion.


umesh-timalsina (approved):
LGTM!!

PR #504, opened 2021-02-09
Fix error in mbuild conversion when compound.name is empty

### LINKED ISSUE #503: Broken `repr` when using `from_mbuild`
I used the following code to create a `gmso.Topology` and get the following error with the `repr`:

### Code:
```python
import mbuild
from gmso.external import from_mbuild

mol = mbuild.load("CCOC", smiles=True)
top = from_mbuild(mol)
print(top)
```

### Error:
```
  File "/Users/ryandefever/Desktop/break.py", line 7, in <module>
    print(top)
  File "/Users/ryandefever/repos/mosdef/gmso/gmso/core/topology.py", line 658, in __repr__
    descr.append(self.name + ' ')
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

rsdefever:
Closes #503 

The error arose because of the condition that accepted the `compound.name` if it was not the default value. I guess when creating a molecule from SMILES, the `compound.name` ends up being empty, i.e., `compound.name = ''`. 

I made a change to the `name.setter` in `gmso.Topology` so that if the string is empty or `None`, then the default name, `"Topology"` is used.

### DIFF
--- a/gmso/tests/test_convert_mbuild.py
+++ b/gmso/tests/test_convert_mbuild.py
@@ -165,5 +165,5 @@ class TestConvertMBuild(BaseTest):
     def test_empty_compound_name(self):
         compound = mb.load("CCOC", smiles=True)
         top = from_mbuild(compound)
-        print(top)
+        assert top.name is not None
 
--- a/gmso/tests/test_convert_mbuild.py
+++ b/gmso/tests/test_convert_mbuild.py
@@ -161,3 +161,9 @@ class TestConvertMBuild(BaseTest):
         top = from_mbuild(mb_ethane)
         assert_allclose_units(top.box.lengths,
                 (mb_ethane.boundingbox.lengths + [0.5, 0.5, 0.5]) * u.nm, rtol=1e-5, atol=1e-8)
+
+    def test_empty_compound_name(self):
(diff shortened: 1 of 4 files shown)


gmso/tests/test_convert_mbuild.py
```diff
@@ -161,3 +161,9 @@ def test_pass_box_bounding(self, mb_ethane):
         top = from_mbuild(mb_ethane)
         assert_allclose_units(top.box.lengths,
                 (mb_ethane.boundingbox.lengths + [0.5, 0.5, 0.5]) * u.nm, rtol=1e-5, atol=1e-8)
+
+    def test_empty_compound_name(self):
+        compound = mb.load("CCOC", smiles=True)
+        top = from_mbuild(compound)
+        print(top)
```
umesh-timalsina, line 168:
```suggestion
        print(top)
        assert top.name is not None
```

umesh-timalsina (approved):
LGTM!! Thanks @rsdefever.

daico007 (approved):
LGTM

justinGilmer (approved):
Reviewed with @umesh-timalsina and @daico007 , LGTM

PR #514, opened 2021-02-20
Fix typo in Periodic Torsion

rsdefever:
I believe there is a typo in the `PotentialTemplate` for periodic torsions. The value is squared when it should not be. See reference [here](https://manual.gromacs.org/current/reference-manual/functions/bonded-interactions.html#proper-dihedrals-periodic-type).


daico007 (approved):
LGTM! Thanks for the fix!

justinGilmer:
Thanks for catching this, @rsdefever ! LGTM!

justinGilmer (approved):

PR #516, opened 2021-02-21
Initial addition of scaling factors to topology

### LINKED ISSUE #363: Handling 1-4 scaling factors in Topology objects
This discussion started from #340. 

The question is how we should store 1-4 scaling factors in a `Topology`. For example, when writing an MCF file I need to know what the 1-4 scaling factors are. 

I'll start this thread by pasting the discussion so far.

Matt: 

> At different points in time we've passed around the idea of having a top.force_field attribute that just points to a ForceField object. It doesn't seem like this is implemented, and much of the important data is stored by parsing through the type objects directly, not by passing through a force field class. The two ideas I can come up with are

> 1. Add a top.force_field attribute that behaves as a pointer, so we could do things like top.force_field.scaling_factors['whatever']. This assumes, for now, that there is only one force field per topology, and also that the data contained in each should match.

> 2. Add the scaling factors to the Topology class directly, so that they can be accessed through top.scaling_factors['whatever']. This might be tricky to bookkeep without clear relationships between the classes, but can be done if we are careful

Justin: 

> Hmm that is a good point. Ive been thinking this over and I think I am a bit more on board with number 1. However, it is sort of the opposite of how we do things now. The sites and such have the atom types, once it is a typed topology. I feel like either the atomtypes should carry this, which is a bit redundant or it is carried at the top of the topology's hierarchy.

> Although I am not married to either idea yet. But this will be an issue we need to solve soon, since this information we will need when writing out typed systems to most engines.

> Maybe we do just carry a pointer to the forcefield, and if the system is typed, we should be able to use the scaling factor from that forcefield.

> Although this leads to another issue if you wanted to write out a topology object to a serializable format. Since the forcefield will not be following along (unless we decide to do that either by writing out the xml again, or including all the xml data in the json as well, or when loading in from a JSON file, we could specify if we want to bring the FF information contained in the json, or load a new ff with the structure.), if we are relying on the FF object for that information, we lose that.

Let's get some other opinions here!

rsdefever:
Initial changes to add scaling factors to `Topology`. Closes #363. 

If someone who is familiar with `parmed` and `openmm`, it would be nice to support translation of the scaling factors from `gmso` <--> `parmed` and `gmso` <--> `openmm`.

Note I included the `1-2` and `1-3` factors in addition to `1-4`. Though the `1-2` and `1-3` factors are often zero, this is more general than just specify the 1-4, and supported by [lammps](https://lammps.sandia.gov/doc/special_bonds.html) and [cassandra](https://cassandra-mc.readthedocs.io/en/latest/guides/input_files.html#intramolecular-scaling). I also used a more general term `vdw` rather than `lj`.

There is nothing preventing a user from adding additional scaling factors to the dictionary, though they might not be supported by the writers.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -46,6 +46,9 @@ class Topology(object):
     combining_rule : str, ['lorentz', 'geometric']
         The combining rule for the topology, can be either 'lorentz' or 'geometric'
 
+    scaling_factors : dict
+        A collection of scaling factors used in the forcefield
+
     n_sites : int
         Number of sites in the topology
 
@@ -147,6 +150,14 @@ class Topology(object):
         self._improper_types = {}
         self._improper_types_idx = {}
         self._combining_rule = 'lorentz'
+        self._scaling_factors = {
+            "vdw_12": 0.0,
+            "vdw_13": 0.0,
+            "vdw_14": 0.5,
+            "coul_12": 0.0,
+            "coul_13": 0.0,
+            "coul_14": 0.5,
+        }
         self._set_refs = {
             ATOM_TYPE_DICT: self._atom_types,
             BOND_TYPE_DICT: self._bond_types,
@@ -199,6 +210,24 @@ class Topology(object):
         if rule not in ['lorentz', 'geometric']:
             raise GMSOError('Combining rule must be `lorentz` or `geometric`')
         self._combining_rule = rule
+
+    @property
+    def scaling_factors(self):
+        return self._scaling_factors
+
+    @scaling_factors.setter
+    def scaling_factors(self, scaling_factors):
+        expected_items = ["vdw_12", "vdw_13", "vdw_14", "coul_12", "coul_13", "coul_14"]
+        if not isinstance(scaling_factors, dict):
+            raise GMSOError("Scaling factors should be a dictionary")
+        for item in expected_items:
+            if item not in scaling_factors.keys():
+                raise GMSOError(f"Expected {expected_items} as keys in the scaling factors")
+        for val in scaling_factors.values():
+            if val < 0.0 or val > 1.0:
+                raise GMSOError("Scaling factors should be between 0.0 and 1.0")
+
+        self._scaling_factors = scaling_factors
 
     @property
     def positions(self):
--- a/gmso/formats/mcf.py
+++ b/gmso/formats/mcf.py
@@ -5,6 +5,7 @@ import networkx as nx
 import sympy
 import unyt as u
 
+from gmso import __version__
 from gmso.core.topology import Topology
 from gmso.lib.potential_templates import PotentialTemplateLibrary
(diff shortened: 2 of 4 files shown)


justinGilmer:
I like the changes this includes, although since we will need to handle these filetype by filetype, we might need to include a "gmso warning" for scaling factor usage in these other files until we apply this to our various writers? Curious what others think about that.

rsdefever:
Yes, I agree that we need to be careful. But this is going to be an issue longer term as well. There are a million different ways to specify 1-4 scaling factors: are they in the topology file, or input file? Do you have the option to control 1-2 and 1-3 or only 1-4? What about charmm style, where, IIRC, they are specified per-1-4-pair? I think what we need to do is to be as flexible as possible in GMSO, and then explain _exactly_ how each writer handles all of these sorts of details in the docs. I.e., each type of file writer should have a detailed documentation page that explains exactly how all of these choices/details are handled.

bc118:
Here is an example of the pair style input for LAMMPS.  Which can account for different 1-n (up to 1-4 in LAMMPS) scaling factors, and different LJ/Non-bonded interactions.  IMPORTANT NOTE: I am unsure how many different hybrid pair styles are allowed, and it is not listed that I see.  There may be a better way to do this, but off-hand, it's the way I know.  Please let me know if there are any questions.

[LAMMPS_different_1_n_and_LJ_interactions.txt](https://github.com/mosdef-hub/gmso/files/6064103/LAMMPS_different_1_n_and_LJ_interactions.txt)

bc118:
Here is an example of a CHARMM style FF input for GOMC, where the LJ 1-4 interactions are rescaled to individual pairs in the NBFIX section (see bottom of FF file).  There should be no limit to the number of LJ 1-4 interactions in the CHARMM format.  

NOTE: the Coul 1-4 interactions are **not** currently able to be different in GOMC, as they are input as 1 constant in the GOMC control file.

bc118:
General comments:

- I'm currently not aware of FFs without zero's for the 1-2 or 1-3 interactions, but we should include them.  I'm sure as soon I say this someone will have an example.

- There are 1-5 and 1-6 scaling factors used in this paper (OPLS-AA FF), but I'm guessing this is not very common (
https://pubmed.ncbi.nlm.nih.gov/12370944/ ).  However, we may want to include them as optional but should fail to write the files they exist and the given engine does not support them.  

- I believe the 1-n interactions should be set to the common one for the given FF, and the specific interactions will have the exact 1-n parameters set for that pair.   This may allow fewer errors for the majority of the interactions, and the ability to easily see which pairs are special.  This structure would allow for the special LJ scaling to be paired alongside it easily, which will also be needed down the road.

bc118 (approved):
I had previously reviewed this code think that is accurate and accomplishes the goals set out.  I do not recommend any changes.  My comments are included in the PR as things to look for and consider in the future, and the existing code does not impede these additions in the future if they are needed.

PR #523, opened 2021-03-11
Abstract_Site_Representations

CalCraven:
# The goal of of this PR is to allow support of different site classes beyond the currently implemented `Atom` Class.


umesh-timalsina:
@chrisjonesBSU @jennyfothergill deferring this to you guys' expertise.

gmso/core/bead.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    mass_: Optional[Union[u.unyt_quantity, float]] = Field(
+        None,
+        description='Mass of the bead'
+    )
+
+    bead_type_: Optional[AtomType] = Field(
+        None,
+        description='AtomType associated with the atom'
```
jennyfothergill, line 53:
```suggestion
        description='AtomType associated with the bead'
```
should this be bead?

jennyfothergill (commented):
Forgive me if this review is not helpful--I'm not familiar with gmso.
These ideas are really interesting to me:
>Add a Virtual_Site class that is drawn from a grouping of Bead or Atom classes.
>Implement relevant usages for information drawn from Virtual_Site.
>Add tests for implementation of a topology that includes a virtual site.
I like the idea of using the `Bead` class to store a coarse grain mapping. (e.g., a thiophene moiety, smiles 'c1cscc1', is mapped to a single bead and maybe the `Bead` class could store that mapping.)

bc118:
For all GOMC and many other bio-type simulations the residues are required to conduct a standard simulation and analysis.  

**For GOMC the residue names are required for the following reasons:** 
- The residues are essential a molecule type, and the residues names are required to build the systems in MoSDeF.  Currently, in GOMC, A molecule with multiple residues will not work (i.e., a protein with all the amino acids named as different residues).
-  The special MC moves require the molecule to be specified by the residue names along with other molecule characteristics.  
- When using the CHARMM file writers in MoSDeF the FFs are specified on a per residue basis (i.e., per molecule as GOMC is currently designed)

**Other bio simulations in MoSDeFs future:** 
- There are multiple residues per molecules in a protein, with all the amino acids named as different residues.  This is the standard method for running these bio simulations.
- All the currently protein/bio analysis codes are specifically designed to select atom based on their residues.  therefore, without this attribute the bio people will in all likelihood avoid any code that does not include it.

jennyfothergill:
I wonder if residue/virtual site could be grouped into one attribute. It seems to me that they would both represent a grouping of particles and could contain which particles it contains, a type name, and an id/index? Is there anything further that a bio reside would need?

bc118:
As long as GOMC or bio engines can access this residue for building the engines input files (psf, pdb, and FF files) info we "should" be OK.  

There could be some bonding between the same type of residues in a protein, or if side group is added.  Therefore, all the residues may not be exactly the same molecularly speaking.   

Example: A residue at the beginning or end of a protein will be "capped" and have a slightly different structure that the same residue in the middle of the protein.  This is the same concept as building a polymer from monomers with the ends being different, except the monomer is the amino acids in the bio case.  

I just think we should make the residue information flexible as described above so bio application are easily accommodated.

gmso/core/bead.py
CalCraven: Yes, good catch @jennyfothergill. I'm going to go through and try to write some more implementation for the bead class. Based on the properties you mentioned above, looks like it will act relatively similarly to the current `site` class.

CalCraven:
> I would request that the residues be a part of the atom/bead in all cases, preferably as an attribute. I outline some of the major reasons below.

I agree with these reasons. I think particular for integration with commonly used bio simulation tools, such as [Chimera](https://www.cgl.ucsf.edu/chimera/) or [Avogadro](https://avogadro.cc/) that rely on residues for most of their tools and analysis, I think this makes sense to implement within this PR. 

> I wonder if residue/virtual site could be grouped into one attribute. It seems to me that they would both represent a grouping of particles and could contain which particles the residue/virtual site contains, a type name, and an id/index? Is there anything further that a bio reside would need?

I see what you're saying @jennyfothergill, but as far as I can tell there might be differences that are worth implementing them separately. I think it makes sense to include in each `atom` or `bead` class an attribute called Residue. In my experience, Gromacs uses these for a lot of the selection syntax, and it replaces the necessity of carrying around bulky `groups.ndx` as an extra specification to gmx grompp. Lammps also has a molecule-tag that could pull from the residue attribute, although it looks like it will only take integers for that specification. For engines that don't need residue information, that attribute just wouldn't be recorded in the output files.
However, I think virtual sites would be used on a less frequent basis. Maybe you want to keep track of the center of mass of a group of atoms, so you can specify a virtual site made up of that group. That may be from all atoms that make up a single residue, but I can see use cases where you want information from a more flexible grouping. I think including a secondary virtual_site class would be a good way to treat these situations. 

As far as actually creating this `virtual_sites` class, I think the most difficult thing will be making sure the group is properly translated to various engines, and will certainly require some software specific knowledge of how these are treated in the engines we want to support. That might have to be done in a separate PR for each engine. But I think some basic knowledge of what that would look like will help direct the way they are built in GMSO.

PR #534, opened 2021-04-25
Fix azure build 

daico007:
Attempt to fix the azure build (make it works with the new yml setup)


justinGilmer (approved):
Good catch! LGTM!

Issue #535, opened 2021-04-26
Update gmso docker to be more like mbuild (i.e., easier to use)


rsdefever:
See [mBuild #791](https://github.com/mosdef-hub/mbuild/pull/791) for details.

PR #537, opened 2021-04-27
Simplify Docker image

rmatsum836:
Addresses #535, implemented the changes made to the mBuild image.  Most of the changes to the documentation are copied from https://github.com/mosdef-hub/mbuild/pull/791.

I built and ran the docker image locally, everything was working for me.  I ran into an issue where I couldn't launch a Jupyter notebook because the package wasn't installed.  I fixed this by adding it to the `requirements-dev.txt` (I didn't push this change), although it seems like I shouldn't have to do this.

I also created a `devtools` directory, but I could also move `docker-entrypoints.sh` to the root directory if preferred.


daico007 (approved):

docs/docker.rst
```diff
@@ -14,51 +14,97 @@ After you have a working docker installation, please use the following command t
...
-    $ docker run -it --name gmso mosdef/gmso:latest
+    $ docker run -it --mount type=bind,source=$(pwd),target="/home/anaconda/data" mosdef/gmso:latest sh 
+    ~ $ source .profile
+    (gmso-dev) ~ $
 
-.. important::
+.. Warning::
 
-    The instructions above will start a docker container but containers by nature are ephemeral, so any filesystem changes (like adding a new notebook) you make will only persist till the end of the container's lifecycle. If the container is removed, any changes or code additions will not persist.
+Containers by nature are ephemeral, so filesystem changes (e.g.,m adding a new notebook) only persist until the end of the
```
umesh-timalsina, line 35:
```suggestion
Containers by nature are ephemeral, so filesystem changes (e.g., adding a new notebook) only persist until the end of the
```

umesh-timalsina (commented):
Simple comment on e.g. Also, I think the entrypoint file appears twice.

rmatsum836:
Would someone mind building locally and testing before merging?

devtools/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+. /opt/conda/etc/profile.d/conda.sh
+conda activate base
+conda activate gmso-dev
+
+if [ "$@" == "jupyter" ]; then
+      jupyter notebook --no-browser --notebook-dir /home/anaconda/data --ip="0.0.0.0"
```
justinGilmer, line 8: Just tested, this line fails since we do not have `jupyter` installed as a dependency. i can go into a shell session no problem though. The easiest fix is to probably include a line in the dockerfile RUN command to install jupyter after we create the environment.

justinGilmer (changes requested):
We need to include `jupyter` in the `RUN` command after the `gmso-dev` environment is created. This is most likely needed for mbuild as well.

justinGilmer:
Just pushed a commit to this dockerfile to see if it fixes the issue. Building locally.

justinGilmer:
Ok can confirm that the edits work successfully.

justinGilmer (approved):

Issue #521, opened 2021-03-05
Loading larger ForceField files is not performant


umesh-timalsina:
It might be nice to investigate why the creation of a `ForceField` for larger files takes a long time. To my knowledge, this is because of the long time it takes to build and validate potential expressions for the potential groups from the XML file. It might be nice to find and cache the potential expression validation function and/or do some smart creation of potentials while creating a `ForceField`. It might also be nice to consider some other avenues for optimizing reading forcefield XML files.

umesh-timalsina:
Loading in `oplsaa_gmso.xml`:
```py
>>> timeit.timeit("from gmso import ForceField;ForceField('oplsaa_gmso.xml', strict=False)", number=1, globals={'get_path': get_path})
41.384391662984854

```

rsdefever:
Yeah thats not going to fly. Most force fields are going to have the same potential expressions _over and over and over and over and over_. Since GMSO is supposed to be flexible we need to be able to handle the case when the potentials _are_ different. But I agree, some sort of caching or other optimization strategy would be helpful. This is such a simple solution that it is probably dumb, but could we just cache the exact strings used for the potential expression -- and if they are the same then call it a match and go on?

umesh-timalsina:
Yeah. We might need to resort to some delayed validation for the `PotentialExpression`. Basically, your XML file is already validated such that, your expressions and independent should match before you create a `PotentialExpression`. Regardless, the fist step I think here should be:
1. memoizing the `@staticmethods` [here](https://github.com/mosdef-hub/gmso/blob/fee3df261a74e39977a5f5d51ca4abd408aca2db/gmso/utils/expression.py#L211-L307). 
2. A delayed validation of `PotentialExpression` on demand.

mikemhenry:
@umesh-timalsina have you tried to profile the code to figure out where it is spending the most time? There could be some really low-hanging fruit.

umesh-timalsina:
@mikemhenry Yeah. I profiled the creation `opls` forcefield using `vprof`. The code heatmap shows that around 21% of the time is spent on validating expression with their parameters. Which doesn't need to happen for an already validated XML IMO. 
[ffcreate.json.zip](https://github.com/mosdef-hub/gmso/files/6093097/ffcreate.json.zip)
A lot of time is being spent in validation of `MetadataMixin` as well.

mikemhenry:
Cool, then an LRU cache should really speed that up

daico007:
[...]

Issue #473, opened 2020-10-15
Add unit test that confirms timing on large system


rsdefever:
(Somewhat abstract) idea: Add unit test(s) to make sure that new features don't negatively affect the performance of GMSO. I'm not exactly sure what this looks like, but something along the lines of building a large system and ensuring the time doesn't exceed some limit. `pytest` has timeout capabilities. I could use some brainstorming from others on what the specifics of the test should look like though.

rmatsum836:
We have something similar on foyer in `test_performance.py` where we load in a `mb.Compound`, atom type with foyer, and use `pytest.mark.timeout()`.

rmatsum836:
@rsdefever Here are several ideas for tests.  I think we'll eventually want to test atom-typing with GMSO once it's a supported feature.

- Writers
- Converter from mBuild
- Loading in a large FF XML (OPLS perhaps)

Also, just so we're on the same page, what would we consider to be a large system?

rsdefever:
@rmatsum836 I would say at least 100k atoms. If we really wanted to push it, then 500k atoms. But 100k atom systems aren't that "big" anymore.

rmatsum836:
Makes sense.  My concern is the wall time required for operations with 100k atoms.  Right now these unit tests would be quite long.  I wonder if it's possible to only run these tests on PR builds.

I'm not sure how much we've worried about optimization at this point.  I think it makes sense to leave this issue open and adjusts the tests once we're happy with performance.  Then the tests can ensure that new features aren't degrading performance.

rsdefever:
I'm fine with keeping the tests smaller for now, and then adding larger once as we get better performance.

PR #539, opened 2021-04-28
Add performance checks

rmatsum836:
Addresses #473.  The tests mostly check that writers are performant; the system size and timeout values could probably be tweaked, I based them off simple profiling I did with `time`.  

I also made a note to check loading a FF XML once #521 is addressed.  Should also add checks once foyer integrations is accomplished.

I'm open to any other suggestions for tests we could add.


jennyfothergill (approved):
This is awesome and I think we should implement something similar in other mosdef packages. :) 

I wonder if there is a way to implement something like [pytest-profiling](https://pypi.org/project/pytest-profiling/) to check if a test is slower/faster from one commit to the next? 🤔 Just a thought!

justinGilmer (approved):
Awesome idea! LGTM!

PR #543, opened 2021-05-18
Maint/pre commit

justinGilmer:
As is the case with foyer and mBuild, supporting autoformatting and various lint checks before committing makes reviewing pull requests much simpler for reviewers and contributors.

This PR is the initial setup for pre-commit in gmso.

However, one thing to note is that the documentation we get from pydantic does not 100% align with the numpydoc style that `pydocstyle` checks, which led to a lot of difficulty initially. @umesh-timalsina and I decided that we did not want to lose the useful features of the pydantic documentation we currently have set up. So a few of the main pydantic-heavy objects we use in gmso `gmso/core, gmso/abc` are excluded from the pydocstyle check. I added quite a few docstrings to these objects anyway, but it should not be required.


gmso/core/improper.py
```diff
@@ -31,80 +32,89 @@ class Improper(Connection):
     """
     connection_members_: Tuple[Atom, Atom, Atom, Atom] = Field(
         ...,
-        description='The 4 atoms of this improper. Central site first, '
-                    'then the three atoms connected to the central site.'
+        description="The 4 atoms of this improper. Central site first, "
```
umesh-timalsina, line 35:
```suggestion
        description="The 4 atoms of this improper. Central atom first, "
```

daico007 (approved):
LGTM!

umesh-timalsina:
pre-commit.ci run

umesh-timalsina (approved):
LGTM!!

PR #546, opened 2021-05-24
Update syntax with new `mb.Box`

daico007:
Update syntax that used `mb.Compound.boundingbox`. Will also pin `mbuild` version in a later commit (after it is released).

### DIFF
--- a/gmso/external/convert_mbuild.py
+++ b/gmso/external/convert_mbuild.py
@@ -118,7 +118,7 @@ def from_mbuild(compound, box=None, search_method=element_by_symbol):
     # if compound.periodicity is None and not box:
     else:
         if np.allclose(compound.periodicity, np.zeros(3)):
-            box = from_mbuild_box(compound.boundingbox)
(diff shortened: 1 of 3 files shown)


gmso/external/convert_mbuild.py
```diff
@@ -118,7 +118,7 @@ def from_mbuild(compound, box=None, search_method=element_by_symbol):
     # if compound.periodicity is None and not box:
     else:
         if np.allclose(compound.periodicity, np.zeros(3)):
-            box = from_mbuild_box(compound.boundingbox)
+            box = from_mbuild_box(compound.get_boundingbox())
```
justinGilmer, line 121:
```suggestion
            if not compound.box:
                box = from_mbuild_box(compound.get_boundingbox())
            else:
                box = from_mbuild_box(compound.box)
```

justinGilmer (approved):

umesh-timalsina (approved):

PR #549, opened 2021-06-09
Add member_classes attribute to parametric potentials. 

### LINKED ISSUE #548: ParametricPotentials should also have member classes?
I know we have discussed this issue before as well but never made a concrete decision on what needs to happen. But while working on #541, this issue rendered itself in a blocking fashion. 

## Some Context
Currently all the parametric potentials (`BondType`, `AngleType` etc...) in `gmso` have an attribute called `member_types`. However, in the ForceField representation and XML, we also allow Potentials to be defined based on `atomclass`. But while reading in XML file, the information on whether the `ParametriPotential` referred to `atomclass` or `atom_type` is lost. So, in a round-trip XML conversion an `atomclass` in the original XML will be converted to `atomtype`. 

## Resolution
For no conflict, I propose the following:
1. Allow `ParametricPotentials` to store `atomclass` information using `member_classes`attribute.
2. Don't allow criss-crossing between `atomclass` and `atom_type` while defining a `ForceField`.

umesh-timalsina:
Closes #548.

This PR add support for member_classes attribute for parametric potentials:

- [x] Add Attribute to potential classes
- [x] Don't allow criss-crossing between `type` and `class` 
- [ ] Tests
- [ ] ForceField creation changes

### DIFF
--- a/gmso/core/angle_type.py
+++ b/gmso/core/angle_type.py
@@ -26,7 +26,13 @@ class AngleType(ParametricPotential):
 
     member_types_: Optional[Tuple[str, str, str]] = Field(
         None,
-        description="List-like of gmso.AtomType.name or gmso.AtomType.atomclass "
+        description="List-like of gmso.AtomType.name "
+        "defining the members of this angle type",
+    )
+
+    member_classes_: Optional[Tuple[str, str, str]] = Field(
+        None,
+        description="List-like of gmso.AtomType.atomclass "
         "defining the members of this angle type",
     )
 
@@ -37,6 +43,7 @@ class AngleType(ParametricPotential):
         parameters=None,
         independent_variables=None,
         member_types=None,
+        member_classes=None,
         topology=None,
         tags=None,
     ):
@@ -55,6 +62,7 @@ class AngleType(ParametricPotential):
             independent_variables=independent_variables,
             topology=topology,
             member_types=member_types,
+            member_classes=member_classes,
             set_ref=ANGLE_TYPE_DICT,
             tags=tags,
         )
@@ -64,6 +72,12 @@ class AngleType(ParametricPotential):
         return self.__dict__.get("member_types_")
 
     class Config:
-        fields = {"member_types_": "member_types"}
+        fields = {
+            "member_types_": "member_types",
+            "member_classes_": "member_classes",
+        }
 
-        alias_to_fields = {"member_types": "member_types_"}
+        alias_to_fields = {
+            "member_types": "member_types_",
+            "member_classes_": "member_classes",
+        }
--- a/gmso/core/improper_type.py
+++ b/gmso/core/improper_type.py
@@ -38,7 +38,13 @@ class ImproperType(ParametricPotential):
 
     member_types_: Optional[Tuple[str, str, str, str]] = Field(
         None,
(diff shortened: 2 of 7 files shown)


gmso/core/improper_type.py
```diff
@@ -83,9 +91,19 @@ def member_types(self):
...
-        fields = {"member_types_": "member_types"}
+        fields = {
+            "member_types_": "member_types",
+            "member_classes_": "member_classes",
+        }
 
-        alias_to_fields = {"member_types": "member_types_"}
+        alias_to_fields = {
+            "member_types": "member_types_",
+            "member_classes_": "member_classes",
```
daico007, line 108:
```suggestion
            "member_classes": "member_classes_",
```

gmso/core/angle_type.py
```diff
@@ -69,7 +77,17 @@ def __init__(
...
-        fields = {"member_types_": "member_types"}
+        fields = {
+            "member_types_": "member_types",
+            "member_classes_": "member_classes",
+        }
 
-        alias_to_fields = {"member_types": "member_types_"}
+        alias_to_fields = {
+            "member_types": "member_types_",
+            "member_classes_": "member_classes",
```
daico007, line 92:
```suggestion
            "member_classes": "member_classes_",
```

bc118:
I started to review this.  From what I see the following is true:
- only classes or types can be used for the xml forcefields bonds, angles, dihedrals, and impropers, not both.  What about the case we discussed that a generic class will be overridden if there is a specific type provided as well?  Or am I misunderstanding this part?

- Only the selected type or class is provided in the output, not both (see my test output below).  I believe we need both the class and type output regardless whether only classes or types are used in the xml forcefield file, so the engine control file writer can parse it accordingly. Thought by others?

**Test member_type and member_class output from the current GOMC-gmso writer:**
bond_type_x = <BondType BondType1, expression: 0.5*k*(r - r_eq)**2, id: 140201461999552>
bond_type_x.member_types = ('o_spce', 'h_spce')
bond_type_x.member_classes = None
self.bond_types_dict_per_residue[res_x] None
angle_type_x = <AngleType AngleType-Harmonic-1, expression: k*(theta - theta_eq)**2, id: 140201441364384>
angle_type_x.member_types = None
angle_type_x.member_classes = ('CT', 'CT', 'HC')
angle_type_x = <AngleType AngleType-Harmonic-2, expression: k*(theta - theta_eq)**2, id: 140201441384576>
angle_type_x.member_types = None
angle_type_x.member_classes = ('HC', 'CT', 'HC')
angle_type_x = <AngleType AngleType1, expression: 0.5*k*(theta - theta_eq)**2, id: 140201478219856>
angle_type_x.member_types = ('h_spce', 'o_spce', 'h_spce')
angle_type_x.member_classes = None

bc118:
As discussed in the meeting we had today, I am OK with the current structure of this PR.

bc118 (changes requested):
I approve this PR, with the addition of a few test cases, which should include the class and type specified in the same bond, angle, dihedral, impropers.

justinGilmer (approved):

daico007 (approved):
LGTM! Will merge when all test pass.

PR #553, opened 2021-06-16
Strip white space when parsing atomtype overrides

daico007:
Strip white space when parsing atom_type overrides from XML. Fix bug mentioned in #551.

### DIFF
--- a/gmso/utils/ff_utils.py
+++ b/gmso/utils/ff_utils.py
@@ -318,10 +318,8 @@ def parse_ff_atomtypes(atomtypes_el, ff_meta):
             )
         if isinstance(ctor_kwargs["overrides"], str):
             ctor_kwargs["overrides"] = set(
-                [
-                    override.strip()
-                    for override in ctor_kwargs["overrides"].split(",")
-                ]
+                override.strip()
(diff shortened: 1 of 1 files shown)


gmso/utils/ff_utils.py
```diff
@@ -317,7 +317,12 @@ def parse_ff_atomtypes(atomtypes_el, ff_meta):
                 float(ctor_kwargs["mass"]), units_dict["mass"]
             )
         if isinstance(ctor_kwargs["overrides"], str):
-            ctor_kwargs["overrides"] = set(ctor_kwargs["overrides"].split(","))
+            ctor_kwargs["overrides"] = set(
+                [
+                    override.strip()
+                    for override in ctor_kwargs["overrides"].split(",")
+                ]
+            )
```
umesh-timalsina, line 325:
```suggestion
            ctor_kwargs["overrides"] = set(override.strip() for override in ctor_kwargs["overrides"].split(","))
```

umesh-timalsina (approved):
Minor suggestion on iteration. LGTM after that.

justinGilmer (approved):
Agree with umesh's suggestion

After that, its g2g

PR #554, opened 2021-06-17
Initial Attempt at Residue Support

umesh-timalsina:
- [x] Data Structure and Iteration
- [x] Functionality Discussion
- [x] Tests
- [x] Preview external and file writers discussion


CalCraven:
Related to #562, a mol2 file contains residue information that should be read in to the topology. For now, I plan to read that information into a label attribute, but once this PR is closer to merging, we should not forget to read in the residue information to this data structure.

bc118 (changes requested):
I used this PR to generate the GMSO-GOMC-MoSDeF coordinate (pdb) and connectivity (psf) files using residue names and indexes. This method works in GOMC now as GOMC's molecules currently need to have all the atoms labeled with the same residue name.  Currently, the residue names and indexes have to be iterated over and input with code after the topology file is created.  In the future, this may need to be automated in this function during the gmso.topology creation when importing a protein with many different residues in a molecule.  However, this method works in GOMC now, as GOMC's molecules currently only need to have all the atoms labeled as the same residue. 

My only other suggestion is to add a test case, but I am fine without it as well.

bc118 (approved):
Looks good, as my only real comment was to add test cases.

umesh-timalsina:
@daico007 There are the following methods in topology class. You can iterate though the topology's sites either by using a residue label or residue index (By using `iter_sites` method). 

https://github.com/mosdef-hub/gmso/blob/1a99c217ff17a344b0f2e7673db8f4c31eb38634/gmso/core/topology.py#L982-L1021

gmso/abc/abstract_site.py
```diff
@@ -45,6 +51,14 @@ class Site(GMSOBase):
 
     label_: str = Field("", description="Label to be assigned to the site")
 
+    residue_index_: Optional[StrictInt] = Field(
```
umesh-timalsina, line 54: Rename it to `residue_number` maybe (Suggested by @justinGilmer). What do you think @bc118?

gmso/abc/abstract_site.py
```diff
@@ -45,6 +51,14 @@ class Site(GMSOBase):
...
+        None, description="Residue index for the site"
+    )
+
+    residue_label_: Optional[StrictStr] = Field(
```
umesh-timalsina, line 58: Rename it to `residue_name` maybe (Suggested by @justinGilmer). What do you think @bc118?

gmso/core/topology.py
```diff
@@ -978,6 +979,57 @@ def _reindex_connection_types(self, ref):
...
+        if value is None:
+            raise ValueError(
+                "Expected `value` to be something other than None. Provided None."
+            )
+
+        for site in self.sites:
+            if getattr(site, key) == value:
+                yield site
+
+    def iter_sites_by_residue_label(self, label):
```
umesh-timalsina, line 1013: Update naming conventions after we figure out site residue specific attribute names.

gmso/abc/abstract_site.py
justinGilmer: Look up the PDB format naming convention for residues

bc118 (approved):
I approve of the latest changes:
- changed "residue_label" to "residue_name"
- changed "residue_index" to "residue_number"

PR #559, opened 2021-06-24
Fix inverted charges in `.top` and `.mcf` formats

### LINKED ISSUE #558: Bug: `gmso` charges are inverted
GMSO charges appear to have the sign flipped (i.e, a charge of +1 is changed to -1). I think the error comes from how we use the `unyt` package. E.g.,

```python
import unyt as u
charge = 0.1 * u.C
charge.in_units(u.charge_electron).value
```

Output:
```
array(-6.24150913e+17)
```

Here is a full example:

```python
import gmso
import mbuild
import foyer

from gmso.formats.top import write_top as gmso_write_top

mb_ethane = mbuild.load("CC", smiles=True)
ff = foyer.forcefields.load_OPLSAA()

pmd_ethane = ff.apply(mb_ethane)
top = gmso.external.from_parmed(pmd_ethane)
top.name = "ethane"

gmso_write_top(top, "test_gmso.top")
pmd_ethane.save("test_parmed.top", overwrite=True)
```

### Solution:
Use `u.charge_proton` rather than `u.charge_electron`.

rsdefever:
The `.top` and `.mcf` writers used `u.charge_electron` rather than `u.elementary_charge`. This inverted the sign of the charges.

Closes #558

### DIFF
--- a/gmso/formats/mcf.py
+++ b/gmso/formats/mcf.py
@@ -288,7 +288,7 @@ def _write_atom_information(mcf, top, in_ring):
                 types[idx],
                 names[idx],
                 site.mass.in_units(u.amu).value,
-                site.charge.in_units(u.charge_electron).value,
+                site.charge.in_units(u.elementary_charge).value,
             )
         )
         if vdw_style == "LJ":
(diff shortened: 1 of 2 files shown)


umesh-timalsina (approved):

justinGilmer (approved):
Makes sense, great catch! Thanks, @rsdefever !

PR #560, opened 2021-06-24
Make MCF writer feature complete

rsdefever:
GMSO has come far since my initial MCF writer PR. The goal of this PR is to make the MCF writer feature complete (i.e., support all the potential types we support with Cassandra).

Most of the features are implemented at this point, but I need to add unit tests.


daico007:
Replaced by #756

PR #562, opened 2021-06-28
Add mol2 parser within MoSDeF ecosystem

CalCraven:
# Overview
We have some support for mol2 file formats, but currently use the load_mol2 from [mdtraj](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/mol2.py). We have run into some issues because of various assumptions they make in their readers, as well as the variety implemented in the [*TRIPOS*](https://www.certara.com/) mol2 file format. See issue [#921](https://github.com/mosdef-hub/mbuild/issues/921) and [#893](https://github.com/mosdef-hub/mbuild/issues/893). This is also a relevant thread for the rdkit package's issues with their own implementation [PR#415](https://github.com/rdkit/rdkit/pull/415).

# Details
## Current Issues to Address:
* Replace mdtraj parser
* Support varying file formatting from different sources
* Support different system configurations (course-grained, bio-molecules, virtual sites)
* Support reading into various aspects of MoSDeF (mBuild `Compound`, mBuild `Lattice`, gmso `Topology`)
* Obvious warnings/errors when reading in nonstandardized formats
## Sources of mol2 file formats to act as references
(please note if any important ones are missing)
* [mdtraj - (read)](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/mol2.py)
* [mdanalysis - (read)](https://docs.mdanalysis.org/1.0.0/documentation_pages/coordinates/MOL2.html)
* [OpenBabel - (read/write)](https://github.com/openbabel/openbabel/blob/master/src/formats/mol2format.cpp)
* [AMBER - (read/write) antechamber ](https://pablito-playground.readthedocs.io/en/latest/tutorials/qmmm_amber_cpmd/mol2_file.html)
[rdkit - (read/write) ](https://github.com/rdkit/rdkit/pull/415)
* [PARMED - (read/write)](https://parmed.github.io/ParmEd/html/readwrite.html)
* [VMD - (read/write) Molefacture](https://www.ks.uiuc.edu/Research/vmd/plugins/molefacture/)
* [KNIME - (read/write)](https://hub.knime.com/inteligand/extensions/com.inteligand.knime/latest/com.inteligand.knime.nodes.io.mol2.Mol2ReaderNodeFactory)
* [Bio3d - (read/write)](https://www.rdocumentation.org/packages/bio3d/versions/2.4-2/topics/read.mol2)
* [DOCK - (write)](http://dock.compbio.ucsf.edu/Contributed_Code/sphere_select.htm)
### Can be supported with the current state of MoSDeF
Atom – mBuild compound
Bond – mBuild compound
CRYSIN – mBuild lattice
FF_PBC – mBuild compound box
FFCON_DISTANCE – topology 
FFCON_ANGLE – topology 
FFCON_TORSION – topology 
FFCON_RANGE – topology 
FFCON_MULTI – topology 
SUBSTRUCTURE – subtopology/separate compounds
### Could be supported in the future
CENTROID – virtual sites
CENTER_OF_MASS – virtual sites
EXTENSION_POINT – virtual sites
LINE – virtual sites
LSPLANE – virtual sites
NORMAL – virtual sites
MOLECULE – mBuild compound (residue)
### Probably will not support
ALT_TYPE
ANCHOR_ATOM
ASSOCIATED_ANNOTATION
COMMENT
DICT
DATA_FILE
RING_CLOSURE
ROTABLE_BOND
SEARCH_DIST – setting constraints
SEARCH_OPTIONS – Running a SYBYL sim
SET – Color schemes
U_FEAT - Unity support
UNITY_ATOM_ATTR - Unity support
UNITY_BOND_ATTR - Unity support
## Sample Tripos File Outline [Shortened from](http://dock.compbio.ucsf.edu/DOCK_6/tutorials/struct_prep/rec_charged.mol2)
@<TRIPOS>MOLECULE
1ABE.pdb
4651 4696 305 0 0
PROTEIN
AMBER99
@<TRIPOS>ATOM
      1 N          18.3590   26.8690   52.9550 N.pl3     1 GLY2       -0.4157
      2 CA         19.6350   26.6320   53.6710 C.3       1 GLY2       -0.0252
      3 C          19.9620   27.8960   54.5040 C.2       1 GLY2        0.5973
      4 O          19.2860   28.0330   55.5390 O.2       1 GLY2       -0.5679
      5 N          20.9160   28.7080   54.0680 N.am      2 LEU3       -0.4157
      6 CA         21.3040   29.9430   54.7530 C.3       2 LEU3       -0.0518
@<TRIPOS>BOND
     1    1 4650 1
     2    2    1 1
     3    2 4638 1
     4    2 4636 1
     5    3    2 1
     6    3    5 am
@<TRIPOS>SUBSTRUCTURE
     1 GLY2        2 RESIDUE           4 A     GLY     1 ROOT
     2 LEU3        6 RESIDUE           4 A     LEU     2
     3 LYS4       14 RESIDUE           4 A     LYS     2
     4 LEU5       23 RESIDUE           4 A     LEU     2

## Sample mol2 file in mBuild
[Betacristobalite](https://github.com/mosdef-hub/mbuild/blob/master/mbuild/lib/surfaces/beta-cristobalite-expanded.mol2)


justinGilmer:
Is this supposed to be a PR or an issue @CalCraven ?

CalCraven:
> Is this supposed to be a PR or an issue @CalCraven ?

@justinGilmer 
It's a PR. I just wrote up the discussion first to have it all in one place, I'm working on some code outline that will get pushed once the structure is right. But I figured this is a good place to put everything as I'm working on it, especially to document what we will and will not support.

CalCraven:
Just to update everyone on the status of this PR:

It should be working in a basic function. It reads in to `topology.sites`, `topology.bonds`, and `topology.box`. I think these are the most typical uses cases for a mol2 structure anyways, I haven't seen any places that write out parameterized information. A few basic tests have been created for mol2 files output from a few different sources, but if anyone has any further use cases that might not be covered, I'm happy to cover that as well. It's a pretty simple code so far.

Lastly, I think this PR is a good chance to also address issue #581 since this parser will heavily make use of that conversion. If anyone has any issues with that, let me know, otherwise I will move forwards with those changes as well.

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+                f, line, topology, site_type
+            )
+        elif line == "":
+            break
+        else:
+            # else, skip to next line
+            line = f.readline()
+    f.close()
+    # TODO: read in parameters to correct attribute as well. This can be saved in various rti sections.
+    return topology
```
umesh-timalsina, line 68: It might be nice to update the topology before returning it.

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+        >>> from gmso.external.convert_mol2 import from_mol2
+        >>> from gmso.external.convert_mbuild import to_mbuild
+        >>> top = from_mol2('myfile.mol2')
+        >>> mbuild_compound = to_mbuild(top)
+    """
+    msg = "Provided path to file that does not exist"
+    if not os.path.isfile(filename):
+        raise OSError(msg)
+    # Initialize topology
+    topology = Topology(name=os.path.splitext(os.path.basename(filename))[0])
```
umesh-timalsina, line 50:
```suggestion
    if not os.path.isfile(filename):
        raise OSError(msg)
    # Initialize topology
    topology = Topology(name=os.path.splitext(os.path.basename(filename))[0])
```

For these functions, I would recommend using `Pathlib`

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+    # save the name from the filename
+    f = open(filename, "r")
+    line = f.readline()
+    while f:
+        # check for header character in line
+        if line.startswith("@<TRIPOS>"):
+            # if header character in line, send to a function that will direct it properly
+            line, topology = parse_record_type_indicator(
```
umesh-timalsina, line 58: What is the motivation behind returning these lines? and the Topology?

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+
+from gmso import Atom, Bond, Box, Topology
+from gmso.core.element import element_by_name, element_by_symbol
+
+
+def from_mol2(
+    filename, site_type="Atom"
+):  # TODO add flags for information to return
+    """Read in a TRIPOS mol2 file format into a gmso topology object.
+
```
umesh-timalsina, line 16: This actual module might belong to `gmso/formats/mol2.py`

umesh-timalsina (commented):
General Comments on questions

umesh-timalsina (commented):
General Comments

bc118:
The mol2 reader function "convert_mol2" seems to work correctly in pulling atom coordinates and bonds.  It is only pulling a single bond for benzene when a double bond is existing, which I believe is the behavior we want.  I have attached the quick code I used to check the mol2 reader and the mol2 files.  

[gmso_mol2_loader.zip](https://github.com/mosdef-hub/gmso/files/6978892/gmso_mol2_loader.zip)

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+    Creates a Topology from a mol2 file structure. This will read in the
+    topological structure (sites, angles, and box) information into gmso.
```
bc118, line 18: should this line    "topological structure (sites, angles, and box) information into gmso."
be    "topological structure (sites, bonds, and box) information into gmso."

I did not see an angles determination.

bc118 (changes requested):
This looks pretty good, aside from my comments.

I suggest making the residue reader if at all possible be for merging, but this also means the current Residue PR needs to be merged first.  This will just make it future proof.  However, this is currently not a critical barrier for getting GOMC-MoSDeF working.

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+    Note that parameterized information can be found in these objects, but
+    will not be converted to the Topology.
+
+    Parameters
+    ----------
+    filename : string
+        path to the file where the mol2 file is stored.
+    site_type : string, optional, default='Atom'
```
bc118, line 26: consider listing the options ('Atom', 'lj') in this line, as I think it makes the options quickly clearer for the reader.

gmso/external/convert_mol2.py
CalCraven: Sounds good, added this.

gmso/external/convert_mol2.py
CalCraven: Used pathlib.Path to read in the filename and check that the path exists.

gmso/external/convert_mol2.py
CalCraven: Totally unnecessary, will remove passing both of these around since they're modified in place.

CalCraven: So I've been playing around with it, I think I need to keep passing the line around with the current usage. If I keep reading until the next rti, then I'll need to hold on to that line to see which function needs to draw that data from in the future. Topology passing is removed though.

gmso/external/convert_mol2.py
CalCraven: moved it into that formats from external! I put it in there to keep it close to the mBuild conversion initially, but this works too.

gmso/external/convert_mol2.py
```diff
@@ -0,0 +1,175 @@
...
+import warnings
+
+import unyt as u
+
+from gmso import Atom, Bond, Box, Topology
+from gmso.core.element import element_by_name, element_by_symbol
+
+
+def from_mol2(
+    filename, site_type="Atom"
```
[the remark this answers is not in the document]
CalCraven: Suggestion taken!

gmso/external/convert_mol2.py
CalCraven: Addressed!

gmso/external/convert_mol2.py
CalCraven: You read it right, it should be bonds and not angles.

gmso/tests/test_mol2.py
```diff
@@ -0,0 +1,73 @@
...
+        assert len(top.sites) == 0
+        assert len(top.bonds) == 0
+
+    def test_broken_files(self):
+        match = "The record type indicator @<TRIPOS>MOLECULE_extra_text\n is not supported"
+        with pytest.warns(UserWarning) as record:
+            from_mol2(get_fn("broken.mol2"))
+        assert record[0].message.args[0] == match
+        match = "This mol2 file has two boxes to be read in, only reading in one with dimensions Box(a=0.72 nm, b=0.9337999999999999 nm, c=0.6646 nm, alpha=90.0 degree, beta=90.0 degree, gamma=90.0 degree)"
+        assert record[90].message.args[0] == match
```
bc118, line 73: Unless I missed it, I would add at least 1 test with the test for the coordinates and in it as well.  This will ensure they are correct.

gmso/formats/mol2.py
```diff
@@ -0,0 +1,175 @@
...
+    Creates a Topology from a mol2 file structure. This will read in the
+    topological structure (sites, bonds, and box) information into gmso.
+    Note that parameterized information can be found in these objects, but
+    will not be converted to the Topology.
+
+    Parameters
+    ----------
+    filename : string
+        path to the file where the mol2 file is stored.
+    site_type : string, optional, options:(default:'atom','lj')
```
bc118, line 26: This line is a little confusing to me.  Maybe something like this is better?

site_type : string ('atom' or 'lj'), default='atom'

gmso/external/convert_mol2.py
bc118: This line is a little confusing to me. Maybe something like this is better?

site_type : string ('atom' or 'lj'), default='atom'

bc118:
Once we get the auto residue added and a test case for it, this should be in good shape.  I can look at it again quickly when this happens.

bc118 (changes requested):
Once the above comments are addresse and we get the auto residue added and a test case for it, this should be in good shape. I can look at it again quickly when this happens.

gmso/formats/mol2.py
bc118: looks good now!

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+    return line
+
+
+def parse_record_type_indicator(f, line, topology, site_type):
+    """Take a specific record type indicator from a mol2 file format and save to the proper attribute of a gmso topology.
+
+    Supported record type indicators include Atom, Bond, FF_PBC, and CRYSIN.
+    """
+    supported_rti = {
+        "@<TRIPOS>ATOM\n": load_top_sites,
```
umesh-timalsina, line 164: Assuming that these lines match exactly might not be the most robust. A suggestion here is to remove these `\n`. And change the conditional predicates to
`line.strip().startswith`

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+    # Initialize topology
+    topology = Topology(name=path.stem)
+    # save the name from the filename
+    f = open(path, "r")
+    line = f.readline()
+    while f:
+        # check for header character in line
+        if line.startswith("@<TRIPOS>"):
+            # if header character in line, send to a function that will direct it properly
+            line = parse_record_type_indicator(f, line, topology, site_type)
```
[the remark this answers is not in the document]
umesh-timalsina: https://github.com/mosdef-hub/gmso/pull/562#discussion_r687979852

Making sure I understand this correctly, sometimes you might get the splitted line (a list) as your input. Is that correct?

umesh-timalsina (commented):
General comments and some questions. Round 2.

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+    It may be common to want to create an mBuild compound from a mol2 file. This is possible
+    by installing [mBuild](https://mbuild.mosdef.org/en/stable/index.html)
+    and converting using the following python code:
+
+        >>> import gmso.Topology
+        >>> from gmso.external.convert_mbuild import to_mbuild
+        >>> top = Topology.load('myfile.mol2')
+        >>> mbuild_compound = to_mbuild(top)
+    """
+    msg = "Provided path to file that does not exist"
```
justinGilmer, line 46: This `msg` variable is only accessed within the `if` statement below on line 49. You can move this variable inside this conditional, or just manually pass in the name since this seems to be a one off thing.

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+    path = Path(filename)
+    if not path.exists():
+        raise OSError(msg)
+    # Initialize topology
+    topology = Topology(name=path.stem)
+    # save the name from the filename
+    f = open(path, "r")
+    line = f.readline()
+    while f:
```
justinGilmer, line 55: Also, if you dont want to have to manually open and then close the file, I recommend changing this into a `with` statement.

```py3
with open(path, "r") as f:
    line = f.readline()
    while f:
        ...
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+            # else, skip to next line
+            line = f.readline()
+    f.close()
+    topology.update_topology()
+    # TODO: read in parameters to correct attribute as well. This can be saved in various rti sections.
+    return topology
+
+
+def load_top_sites(f, topology, site_type="atom"):
+    """Take a mol2 file section with the heading '<TRIPOS>ATOM' and save to the topology.sites attribute."""
```
justinGilmer, line 72: Can you add a full docstring here?

Just so we can enumerate the possible values for "site_type" etc.

gmso/formats/mol2.py
justinGilmer: In favor of this too. If the specification does not allow anything else, then this is probably fine. Can people add comments or anything in this file? If so, what umesh suggests will make it robust enough

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+                lengths=[float(x) for x in line[0:3]] * u.Å,
+                angles=[float(x) for x in line[3:6]] * u.degree,
+            )
+        else:
+            break
+    return line
+
+
+def parse_record_type_indicator(f, line, topology, site_type):
+    """Take a specific record type indicator from a mol2 file format and save to the proper attribute of a gmso topology.
```
justinGilmer, line 159:
```suggestion
    """Take a specific record type indicator (RTI) from a mol2 file format and save to the proper attribute of a gmso topology.
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+        "@<TRIPOS>BOND\n": load_top_bonds,
+        "@<TRIPOS>CRYSIN\n": load_top_box,
+        "@<TRIPOS>FF_PBC\n": load_top_box,
+    }
+    # read in to atom attribute
+    try:
+        line = supported_rti[line](f, topology, site_type=site_type)
+    except KeyError:
+        warnings.warn(
+            "The record type indicator {} is not supported".format(line)
```
justinGilmer, line 174: So what does this do when it returns? Does it still process the lines after it has been returned? The behavior should be included in the warning message I think.

gmso/tests/test_mol2.py
```diff
@@ -0,0 +1,117 @@
...
+
+        top = Topology.load(get_fn("vmd.mol2"))
+        assert top.name == "vmd"
+        assert top.n_sites == 6
+        assert len(top.bonds) == 5
+        assert top.bonds[0].connection_members[0] == top.sites[0]
+        assert top.box == None
+
+        match = "No charges were detected for site C with index 1"
+        with pytest.warns(UserWarning) as record:
```
justinGilmer, line 58: You dont need to manually match the warning or error messages. You can do that with the `match` keyword arguments
```suggestion
        with pytest.warns(UserWarning, match=r"No charges were detected for site C with index 1"):
```

gmso/tests/test_mol2.py
```diff
@@ -0,0 +1,117 @@
...
+            top = Topology.load(get_fn("ethane.mol2"))
+        assert record[3].message.args[0] == match
+        assert list(top.sites)[0].charge is None
+
+        match = "No element detected for site C with index1, consider manually adding the element to the topology"
+        with pytest.warns(UserWarning) as record:
```
justinGilmer, line 64: same here, but the matches are usually done as a raw string. So you have to escape all special characters like commas.
```suggestion
        with pytest.warns(UserWarning, match=r"No element detected for site C with index1\, consider manually adding the element to the topology"):
```

gmso/tests/test_mol2.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+    def test_lj_system(self):
+        top = Topology.load(get_fn("methane.mol2"), site_type="lj")
+        assert np.all([site.element == None for site in top.sites])
+
+    def test_wrong_path(self):
+        with pytest.raises(OSError) as e:
+            Topology.load("not_a_file.mol2")
+        assert e.value.args[0] == "Provided path to file that does not exist"
```
justinGilmer, line 106: same here to use the `match` keyword argument

justinGilmer (changes requested):
Some changes along the way, but this is looking great @CalCraven !

gmso/formats/mol2.py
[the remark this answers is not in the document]
CalCraven: Good question. I have yet to see a mol2 file that is built with bonds before atoms. It would certainly make things difficult to build in gmso since you need to create bonds by having sites. If we wanted to support that, we would have to store the bond information somewhere separately until the site information was read in, and then go back and create the bonds. But it seems to me like a very unusual case since I've never seen it, although it doesn't explicitly specify that ordering in some of the documentation I've seen.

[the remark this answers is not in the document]
CalCraven: In response to the #562 comment:
I think you are correct. The function `parse_record_type_indicator` is going to need the splitted line as its input so it can decide where to send the next section of information. As of the current implementation, I continue trying to read a line until the end of the section is marked by a new line, a blank line, or a new rti. Which means that we don't care if there is a space or not at the end of the line. But if there is an rti right at the end of a section to indicate a new section, I need to keep passing that line info back into `from_mol2` to determine when the new section begins. So that's the motivation behind passing that back and forth, that way extra spaces at the end of sections are handled.

gmso/formats/mol2.py
```diff
@@ -0,0 +1,177 @@
...
+    topology = Topology(name=path.stem)
+    # save the name from the filename
+    f = open(path, "r")
+    line = f.readline()
+    while f:
+        # check for header character in line
+        if line.startswith("@<TRIPOS>"):
+            # if header character in line, send to a function that will direct it properly
+            line = parse_record_type_indicator(f, line, topology, site_type)
+        elif line == "":
```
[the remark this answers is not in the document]
CalCraven: This bit stops the function from freezing while parsing the end of the topology. There's probably a better way to do this any suggestions?

gmso/formats/mol2.py
CalCraven: I removed the \n for the record types and added strip() methods to try to isolate just the record type indicator titles. The specification is pretty exact on how these titles should be named, so we should be okay here I think.

As for comments, comments are absolutely fine! As long as they're not embedded in the data, they shouldn't be read in. See `gmso/utils/files/benzene_ua.mol2` for an example file that has a generated by VMD comment at the end.

gmso/core/topology.py
```diff
@@ -978,6 +979,47 @@ def _reindex_connection_types(self, ref):
...
+            The value that the given attribute should be equal to
+
+        Yields
+        ------
+        gmso.abc.abstract_site.Site
+            The site where getattr(site, key) == value
+        """
+        if key not in Site.__iterable_attributes__:
+
+            raise ValueError(
```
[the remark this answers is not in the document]
CalCraven: Uncertain. These are changes from the PR I pulled from  PR #578. So @umesh-timalsina might be better at answering that question.

gmso/formats/mol2.py
CalCraven: Good catch, I moved it into the if statement for clarity.

gmso/formats/mol2.py
CalCraven: Good point, will be added.

gmso/formats/mol2.py
CalCraven: It just spits out a warning and skips to the next RTI. 

Adding this to the warning: "Skipping current section and moving to the next RTI header."

bc118 (approved):
This looks good to me.  From what I see this should read in proteins, now.  Very excited!!

justinGilmer (commented):
Small question and once you resolve the merge conflicts, this should be g2g!

gmso/formats/mol2.py
```diff
@@ -43,36 +43,63 @@ def from_mol2(filename, site_type="atom"):
...
+                line = _parse_record_type_indicator(
+                    f, line, topology, site_type
+                )
+            elif line == "":
+                # check for the end of file
+                break
+            else:
+                # else, skip to next line
+                line = f.readline()
     topology.update_topology()
```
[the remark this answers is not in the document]
CalCraven: I tried my best to write something that could use a for loop instead of a while loop. It starts to get tricky again when it finds a header in the line. That passes it to the `_parse_record_type_indicator` function, which reads through the rest of that heading. In the case of a record type indicator (rti) starting a new section without a space after the previous section, that line must be passed back to the parser in order to continue forwards and identify which rti you are dealing with. With a for loop, every time through you must read the next line. In the instance as discussed above:
"""
@<TRIPOS>ATOM
Atom information
@<TRIPOS>BOND
bond information
"""

The next line cannot be read automatically because it will already be read when looking to find the end of the ATOM rti section. 

Does that make sense? There may be a way to accomplish this while still using a for loops, but I tried exactly what you suggested and I run into this error still.

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+
+        >>> import gmso.Topology
+        >>> from gmso.external.convert_mbuild import to_mbuild
+        >>> top = Topology.load('myfile.mol2')
+        >>> mbuild_compound = to_mbuild(top)
+    """
+    path = Path(filename)
+    if not path.exists():
+        msg = "Provided path to file that does not exist"
+        raise OSError(msg)
```
umesh-timalsina, line 49: A better/fitting exception should be [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError)
```suggestion
        raise FileNotFoundError(msg)
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+        warnings.warn(
+            "The record type indicator {} is not supported. Skipping current section and moving to the next RTI header.".format(
+                line
+            )
+        )
+        line = f.readline()
+    return line
+
+
+def _if_end_of_rti(line):
```
umesh-timalsina, line 209:
```suggestion
def _is_end_of_rti(line):
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+    """Check if line in an rti is at the end of the section."""
+    if (
+        "@" not in line
+        and not line == "\n"
+        and line
+        and not line.strip().startswith("#")
```
umesh-timalsina, line 215: What is the third check for?

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+    ):
+        return True
+    else:
+        return False
```
umesh-timalsina, line 219:
```suggestion
return (
            "@" not in line
            and not line == "\n"
            and line
            and not line.strip().startswith("#")
        )
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+    -------
+    top : gmso.Topology
+
+    Notes
+    -----
+    It may be common to want to create an mBuild compound from a mol2 file. This is possible
+    by installing [mBuild](https://mbuild.mosdef.org/en/stable/index.html)
+    and converting using the following python code:
+
+        >>> import gmso.Topology
```
umesh-timalsina, line 41:
```suggestion
        >>> from gmso import Topology
```

gmso/formats/mol2.py
```diff
@@ -0,0 +1,219 @@
...
+        >>> from gmso.external.convert_mbuild import to_mbuild
+        >>> top = Topology.load('myfile.mol2')
+        >>> mbuild_compound = to_mbuild(top)
+    """
+    path = Path(filename)
```
umesh-timalsina, line 46: Refactor the name path (path not a keyword but shadows the builtin module `os.path` and should be avoided if possible).

umesh-timalsina (changes requested):

gmso/formats/mol2.py
CalCraven: The third check will check for the end of the file (line will be None it you've reached that point in the readline).

gmso/formats/mol2.py
CalCraven: Changed to mol2path in 3f1dafb307428288ee67a80fe74c4a9e4f68c2cd

gmso/formats/mol2.py
CalCraven: Changed in 3f1dafb307428288ee67a80fe74c4a9e4f68c2cd.

gmso/formats/mol2.py
umesh-timalsina: If line is `None`, the first condition `'@' in line` will error out.

```py
>>> '@' in None
TypeError: argument of type 'NoneType' is not iterable

```
The way to get around this is to include the `NonType` check first, for short circuiting.
```py
>>> a = None
>>> bool(a and '@' in a)
False

```

CalCraven: fixed in commit 54e0b7763929045706cdb55984ea77d82e61f5d3

justinGilmer (approved):

PR #565, opened 2021-06-29
Wrap box with molbox

### LINKED ISSUE #563: Use molbox for box implmentation
We have refactored mbuild's box code contributed by @justinGilmer to a separate Python package (available [here](https://github.com/mosdef-hub/molbox)). It might be nice to extend the current Box implementation by wrapping the functionality using the `Box` class provided by `molbox`.

umesh-timalsina:
Closes #563.

### DIFF
--- a/gmso/core/box.py
+++ b/gmso/core/box.py
@@ -23,7 +23,6 @@ def _validate_lengths(lengths):
 
     lengths *= input_unit
     lengths.convert_to_units(u.nm)
-
     if np.any(
         np.less(
             lengths,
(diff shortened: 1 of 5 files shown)


gmso/core/box.py
```diff
@@ -117,30 +117,75 @@ class Box(object):
...
-    def __init__(self, lengths, angles=None):
+    def __init__(self, lengths, angles=None, precision=None):
         """Construct a `Box` based on lengths and angles."""
-        self._lengths = _validate_lengths(lengths)
-        self._angles = _validate_angles(angles)
+        print(lengths, angles)
+        lengths = _validate_lengths(lengths)
+        angles = _validate_angles(angles)
+        print(lengths, angles)
+        precision = int(precision or 6)
```
justinGilmer, line 126: Does this set precision to 6 if you cant typecast precision to an int?

umesh-timalsina: I think it does because you have handled that in Molbox.

justinGilmer: Ah true, that makes sense.

PR #567, opened 2021-07-07
Add serialization for Topology

umesh-timalsina:
This PR attempts to implement a full serialization for topology. Two methods are available:

```py
Topology.save(filename, overwrite=False, **kwargs)
Topology.load(cls, filename) 
```


umesh-timalsina:
This PR also brings in the concept of Registries to save/load files. Only the `json` methods using the registries but can be extended to other formats as well.

umesh-timalsina:
The following [gist](https://gist.github.com/umesh-timalsina/f127ad8e7e2eefc842c82128eb9abca5) contains simple test cases.

gmso/formats/json.py
```diff
@@ -0,0 +1,308 @@
...
+            target = targets[type(connection)]
+            connection_dict["connection_members"] = [
+                top.get_index(member)
+                for member in connection.connection_members
+            ]
+            target.append(connection_dict)
+            connection_type = getattr(connection, exclude_attr)
+            if types and connection_type:
+                connection_dict[exclude_attr] = id(connection_type)
+    if types:
```
umesh-timalsina, line 105: - Partially parameterized systems should warn the users if they are saving it with `types=True`.
- Fully/Partially parameterized systems should warn the users that type info will be lost during conversion with `types=False`.

gmso/formats/json.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Returns
+    -------
+    gmso.Topology
+        the equivalent Topology representation from the dictionary
+    """
+    from gmso.core.subtopology import SubTopology
+    from gmso.core.topology import Topology
+
+    # FixMe: DeepCopying a dictionary might not be the most efficient
```
umesh-timalsina, line 149: Add a comment here regarding the motivation for `deepcopy`.

gmso/tests/test_serialization.py
```diff
@@ -188,3 +190,45 @@ def test_include_and_exclude(self):
         assert "name" not in atom_json
         atom_json = atom.json(include={"mass_"})
         assert "name" not in atom_json
+
+    def test_full_serialization(
+        self,
+        typed_ethane,
+        are_equivalent_atoms,
+        are_equivalent_connections,
```
umesh-timalsina, line 198: Single fixture will work here.
```suggestion
```

justinGilmer:
Just tried this out but one thing i noticed is that the potential types are not being brought over in the json files. Is that expected?

an example:

```python3
from mbuild.lib.molecules import Ethane
from gmso.external import from_parmed
import foyer

oplsaa = foyer.forcefields.load_OPLSAA()
eth_typed = oplsaa.apply(Ethane())
topology_typed = from_parmed(topology_typed, refer_types=True)
topology_typed.save("eth.json", overwrite=True, indent=4)
```

umesh-timalsina:
Yeah this is expected. You have to explicitly specify `types=True` while saving it.

justinGilmer:
Very true! I forgot about that flag! Working well for me!

gmso/formats/json.py
umesh-timalsina: This has been added

justinGilmer (approved):
New changes look good. When all checks pass LGTM!

daico007 (approved):
Tested it out locally, LGTM!

Issue #570, opened 2021-07-13
There are 2 ethane fixture in base_test


daico007:
I just realized that they are two fixtures with the same name (i.e. `ethane`) in our `base_test.py`, this is lead to unexpected behavior when they are used in tests later (the second one is grabbed, but that one doesn't have any coords info).

Issue #566, opened 2021-07-01
to_mbuild conversion results in only the x dimension of positions being converted


CalCraven:
# Description
Converting a gmso `Topology` object to an mbuild `Compound` object works properly, except that only the x positions are converted over to the `Compound.pos` and `Compound.xyz` attributes. 

## Code to replicate results
```py
>>> import mbuild as mb
>>> import gmso
>>> from gmso import external
>>> import foyer
>>> compound = mb.load('CCC',smiles=True)
>>> opls = foyer.forcefield.Forcefield(name='oplsaa')
>>> pmd_compound = opls.apply(compound)
>>> gmso_cp = gmso.external.from_parmed(pmd_compound)
>>> mbuild_cp = gmso.external.to_mbuild(gmso_cp)
>>> for part in list(mbuild_cp.particles()):
>>>     print(part.pos)
>>> mbuild_cp.visualize()
```
Returns error
```
TypeError                                 Traceback (most recent call last)
<ipython-input-40-77fc9269429c> in <module>
     10 for part in list(mbuild_cp.particles()):
     11     print(part.pos)
---> 12 mbuild_cp.visualize()

~/miniconda3/envs/switchable38/lib/python3.8/site-packages/mbuild/compound.py in visualize(self, show_ports, backend, color_scheme)
   1341         if run_from_ipython():
   1342             if backend.lower() in viz_pkg:
-> 1343                 return viz_pkg[backend.lower()](
   1344                     show_ports=show_ports, color_scheme=color_scheme
   1345                 )

~/miniconda3/envs/switchable38/lib/python3.8/site-packages/mbuild/compound.py in _visualize_py3dmol(self, show_ports, color_scheme)
   1390                 particle.name = "UNK"
   1391         tmp_dir = tempfile.mkdtemp()
-> 1392         cloned.save(
   1393             os.path.join(tmp_dir, "tmp.mol2"),
   1394             show_ports=show_ports,

~/miniconda3/envs/switchable38/lib/python3.8/site-packages/mbuild/compound.py in save(self, filename, show_ports, forcefield_name, forcefield_files, forcefield_debug, box, overwrite, residues, combining_rule, foyer_kwargs, **kwargs)
   2027         formats.json_formats.compound_to_json : Write to a json file
   2028         """
-> 2029         conversion.save(
   2030             self,
   2031             filename,

[...]

PR #571, opened 2021-07-13
Fix bug when converting mbuild 

daico007:
Fix bug related to #566 and #570. Rename one of the ethane fixture and fix mbuild position conversion issue.


gmso/tests/test_convert_mbuild.py
```diff
@@ -70,6 +71,10 @@ def test_full_conversion(self, ethane):
 
         assert new.n_particles == 8
         assert new.n_bonds == 7
+        for i in range(new.n_particles):
+            print(new[i].xyz)
```
justinGilmer, line 75: Extraneous print statements.

justinGilmer (changes requested):
Extra print statements, but looks good. That was the fix I was thinking for that section. 

Real quick, do we need to make any edits to the others tests for those that might have been using `ethane_from_scratch`

daico007:
The only thing is I don't which test that actually need to use that `ethane_from_scratch`, my guess would be something in `test_topology.py` (I can sit down and go through it sometime later today).

justinGilmer (approved):

PR #572, opened 2021-07-13
Fix default expression for `PairPotentialType`

umesh-timalsina:
#484 introduced a bug where the default value of expression for `PairPotentialType` has one extra bracket. This PR attempts to fix it.

### DIFF
--- a/gmso/core/pairpotential_type.py
+++ b/gmso/core/pairpotential_type.py
@@ -35,7 +35,7 @@ class PairPotentialType(ParametricPotential):
     def __init__(
         self,
         name="PairPotentialType",
-        expression="4 * eps * (sigma / r)**12 - (sigma / r)**6)",
+        expression="4 * eps * (sigma / r)**12 - (sigma / r)**6",
         parameters=None,
         independent_variables=None,
         member_types=None,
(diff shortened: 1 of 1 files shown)


daico007 (approved):

gmso/core/pairpotential_type.py
```diff
@@ -35,7 +35,7 @@ class PairPotentialType(ParametricPotential):
     def __init__(
         self,
         name="PairPotentialType",
-        expression="4 * eps * (sigma / r)**12 - (sigma / r)**6)",
+        expression="4 * eps * (sigma / r)**12 - (sigma / r)**6",
```
daico007, line 38:
```suggestion
        expression="4 * eps * ((sigma / r)**12 - (sigma / r)**6)",
```

daico007 (changes requested):

gmso/core/pairpotential_type.py
justinGilmer, line 38: I dont think the order of operations is correct here:

The LJ potential should be:
this
```suggestion
        expression="4 * eps * ((sigma / r)**12 - (sigma / r)**6)",
```

umesh-timalsina: Thanks. Fixed.

umesh-timalsina: Thanks. Fixed.

justinGilmer (approved):

PR #578, opened 2021-07-27
Feat/writer registries

justinGilmer:
After PR #567, which brought in the addition of concept of Registries to
save/load files; Only the json methods are currently using the registries but can be extended to other formats as well.

This PR introduces the overhaul to the `xyz` file format, one of the
simpler cases. So far, this seems to have worked quite well.

This currently depends on PR #567, so this should not be merged until 567 is merged.


gmso/formats/formats_registry.py
```diff
@@ -15,7 +15,7 @@ def _assert_can_process(self, extension):
         if extension not in self.handlers:
             raise UnsupportedFileFormatError(
                 f"Extension {extension} cannot be processed as no utility "
-                f"is defined in the current API to handle {extension} files."
+                f" is defined in the current API to handle {extension} files."
```
umesh-timalsina, line 18: Extra whitespace

gmso/formats/gro.py
```diff
@@ -116,77 +120,92 @@ def write_gro(top, filename):
...
             "Topology contains some negative positions. Translating "
             "in order to ensure all coordinates are non-negative."
         )
-
+    min_xyz = np.min(top.positions, axis=0)
+    for i, minimum in enumerate(min_xyz):
+        if minimum < 0.0:
+            for site in top.sites:
+                site.position[i] = site.position[i] - minimum
     return top
```
daico007, line 153: For this method, do we just want to work on a clone of the `top` and not the `top` itself? We might bump into the issue with LAMMPS writer in mbuild if we are modifying the original object.

umesh-timalsina (approved):

PR #593, opened 2021-09-21
Fix bug in `_get_improper_type` in `ForceField`

umesh-timalsina:
There was bug in `get_improper_type` for the `ForceField` class, where in the matches were searched incorrectly using the dihedral types of the `ForceField`. This PR attempts to fix it.


daico007 (approved):
LGTM!

PR #594, opened 2021-09-22
Update docker image and re-add anaconda user

justinGilmer:
Previously, the latest version of miniconda3 alpine docker image was incorrectly built as a debian image. This has now been fixed.

However, in between the previous version we have been using, the `anaconda` user was not added by default. I added the necessary `adduser` commands to fix this.


daico007 (approved):
LGTM!

umesh-timalsina (approved):

PR #601, opened 2021-10-19
Properly populate `tags` in `kwargs` for `Potentials`

### LINKED ISSUE #600: Tests are Failing with gmso 0.7.0
**Bug summary**

Since the release of gmso 0.7.0, tests are failing because of a bug where gmso's `AbstractPotential` is not saving tags for the given atom types. 

**Code to reproduce the behavior**
An example of this failure is given below:
https://github.com/mosdef-hub/foyer/blob/5679d7068e99832b93b96c306a110d9377359339/foyer/tests/test_gmso_forcefield.py#L233-L256

This fails because we rely on `atom_type.tags` to get element information:
https://github.com/mosdef-hub/foyer/blob/5679d7068e99832b93b96c306a110d9377359339/foyer/general_forcefield.py#L156-L164

And the tags are empty because of a refactor for the metadata: **Curplit : https://github.com/mosdef-hub/gmso/pull/589**

Since this originated in foyer, I am opening the issue here, but will be transferred to gmso.

umesh-timalsina:
See #600, For a summary of the bug. This PR fixes #600.

### DIFF
--- a/gmso/tests/test_atom_type.py
+++ b/gmso/tests/test_atom_type.py
@@ -16,6 +16,12 @@ class TestAtomType(BaseTest):
     def atomtype_metadata(self):
         return AtomType()
 
+    def test_atom_type_tag_kwarg(self):
+        at = AtomType(
+            tags={"element": "Li", "comesFrom": "ForceFieldExperiments"}
+        )
+        assert at.tag_names == ["element", "comesFrom"]
+
     def test_new_atom_type(self, charge, mass):
         new_type = AtomType(
             name="mytype",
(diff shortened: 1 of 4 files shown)


gmso/tests/test_atom_type.py
```diff
@@ -16,6 +16,12 @@ class TestAtomType(BaseTest):
     def atomtype_metadata(self):
         return AtomType()
 
+    def test_atom_type_tag_kwarg(self):
+        at = AtomType(
+            tags={"element": "Li", "comesFrom": "ForceFieldExperiments"}
+        )
+        assert at.tag_names == ["element", "comesFrom"]
+
```
justinGilmer, line 24: Should we also add a test that ensures that the tags, if empty, return an empty list?

justinGilmer (approved):
Small comment, but LGTM

daico007 (approved):
LGTM!

PR #609, opened 2021-11-30
fix bug when convert position in to_mbuild

daico007:
Found a bug in `to_mbuild` method when converting positions (did not do unit conversion + only end up with one of the x position). This PR will address that and add relevant tests.


daico007:
Pending tests. Will be needed for https://github.com/mosdef-hub/mbuild/pull/971

PR #611, opened 2021-12-01
Add mol2 to formarts __init__.py

daico007:
Add mol2 loader import to format's `__init__.py`. Missing the line doesn't cause any issue with GMSO tests, but it failed when I tried to use it for a different PR in mbuild (https://github.com/mosdef-hub/mbuild/pull/971)

### DIFF
--- a/gmso/formats/__init__.py
+++ b/gmso/formats/__init__.py
@@ -6,6 +6,7 @@ from .gro import read_gro, write_gro
 from .gsd import write_gsd
 from .json import save_json
 from .lammpsdata import write_lammpsdata
+from .mcf import write_mcf
 from .mol2 import from_mol2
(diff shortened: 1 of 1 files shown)


justinGilmer:
Are there any other formats that we have not included here yet? If so, lets add them, and then this LGTM.

daico007:
Based on the files in the `formats`, it seems like the `__init__.py` already get everything but the `networkx` and `mcf` conversion. But I don't think those work readily with the new registries, so might need to be handled in a different PR?

justinGilmer:
MCF is supported with the new registry format, so that can be added. But its only for writing MCF

daico007:
I can add that one in then

justinGilmer (approved):

PR #613, opened 2021-12-08
Update parmed converter to use new residue infrastructure

daico007:
Update parmed conversion method to utilize the new `site.residue_name` and `site.residue_number`.

### DIFF
--- a/gmso/external/convert_parmed.py
+++ b/gmso/external/convert_parmed.py
@@ -456,43 +456,32 @@ def to_parmed(top, refer_type=True):
     dihedral_map = dict()  # Map top's dihedral to structure's dihedral
 
     # Set up unparametrized system
-    # Build subtop_map (site -> top)
-    default_residue = pmd.Residue("RES")
-    for subtop in top.subtops:
-        for site in subtop.sites:
-            subtop_map[site] = subtop
-
     # Build up atom
+    residue_map = dict()
     for site in top.sites:
-        if site in subtop_map:
-            residue = subtop_map[site].name
-            residue_name = residue[: residue.find("[")]
-            residue_idx = int(
-                residue[residue.find("[") + 1 : residue.find("]")]
-            )
-            # since subtop contains information needed to build residue
-        else:
-            residue = default_residue
-        # Check element
         if site.element:
             atomic_number = site.element.atomic_number
             charge = site.element.charge
         else:
             atomic_number = 0
             charge = 0
-
         pmd_atom = pmd.Atom(
             atomic_number=atomic_number,
             name=site.name,
-            mass=site.mass,
-            charge=site.charge,
+            mass=site.mass.to(u.amu).value,
+            charge=site.charge.to(u.elementary_charge).value,
         )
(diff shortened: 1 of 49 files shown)


daico007:
Need to fix the issue when parmed periodic improper converted into gmso proper dihedral

gmso/external/convert_parmed.py
```diff
@@ -452,43 +456,32 @@ def to_parmed(top, refer_type=True):
...
         ).value
 
         # Add atom to structure
-        structure.add_atom(pmd_atom, resname=residue_name, resnum=residue_idx)
+        if site.residue_name:
+            structure.add_atom(
+                pmd_atom, resname=site.residue_name, resnum=site.residue_number
+            )
+        else:
+            structure.add_atom(pmd_atom, resname="RES", resnum=0)
```
CalCraven, line 484: Do we want to default our resnum to -1 like [parmed does?](https://github.com/ParmEd/ParmEd/blob/5e48857403ce8fd7848abed4c58fbf677781f97b/parmed/topologyobjects.py#L4230)

gmso/external/convert_parmed.py
```diff
@@ -452,43 +456,32 @@ def to_parmed(top, refer_type=True):
...
 
     # Set up unparametrized system
-    # Build subtop_map (site -> top)
-    default_residue = pmd.Residue("RES")
-    for subtop in top.subtops:
-        for site in subtop.sites:
-            subtop_map[site] = subtop
-
     # Build up atom
+    residue_map = dict()
```
CalCraven, line 460: Do we use this residue map anywhere?

gmso/external/convert_parmed.py
```diff
@@ -452,43 +456,32 @@ def to_parmed(top, refer_type=True):
...
+            mass=site.mass.to(u.amu).value,
+            charge=site.charge.to(u.elementary_charge).value,
         )
         pmd_atom.xx, pmd_atom.xy, pmd_atom.xz = site.position.to(
             "angstrom"
         ).value
 
         # Add atom to structure
-        structure.add_atom(pmd_atom, resname=residue_name, resnum=residue_idx)
+        if site.residue_name:
```
CalCraven, line 479: So does this also append the atom to the correct [ResidueList](https://github.com/ParmEd/ParmEd/blob/5e48857403ce8fd7848abed4c58fbf677781f97b/parmed/topologyobjects.py#L4479) in the parmed structure? Or does this just tag the atom with the residue information? Might want to add a unit test to check that.

CalCraven (changes requested):
Just needs unit tests and it's good to go!

gmso/external/convert_parmed.py
daico007: This is handled a few lines below in `structure.rsidues.claim()` (so the residue list is updated)

gmso/external/convert_parmed.py
daico007: changed as requested.

justinGilmer (approved):

PR #614, opened 2021-12-14
Fix XML unit conversion

daico007:
When converting from foyer XML to GMSO XML, we forgot the per mol for all `k` (force constant) of bond, angle, etc. This PR fixes that.


CalCraven (approved):
LGTM!

justinGilmer (approved):

PR #615, opened 2021-12-15
Modified RB and Periodic torsions, and added new Mie n-6 potential

bc118:
- The RB torsion should just use C0 and not C0 * cos(phi)**0

- The periodic torsion should just be similar to RB and go to the 5th power series (i.e., cos(5x) term)

- Added the n-6 Mie potential as it is the one GOMC supports.  Also, from my understanding changing both the power series ones can be counterproductive in the force field fitting, but I could be wrong.

### DIFF
--- a/gmso/lib/jsons/PeriodicTorsionPotential.json
+++ b/gmso/lib/jsons/PeriodicTorsionPotential.json
@@ -1,5 +1,5 @@
 {
   "name": "PeriodicTorsionPotential",
-  "expression": "k * (1 + cos(n * phi - phi_eq))",
+  "expression": "k0 + k1 * (1 + cos(1 * phi - phi1_eq)) + k2 * (1 + cos(2 * phi - phi2_eq)) + k3 * (1 + cos(3 * phi - phi3_eq)) + k4 * (1 + cos(4 * phi - phi4_eq)) + k5 * (1 + cos(5 * phi - phi5_eq))",
   "independent_variables": "phi"
 }
--- a/gmso/lib/jsons/MiePotential_n_6.json
+++ b/gmso/lib/jsons/MiePotential_n_6.json
@@ -1,0 +1,5 @@
+{
+  "name": "MiePotential_n_6",
+  "expression": "(n/(n-6)) * (n/6)**(6/(n-6)) * epsilon * ((sigma/r)**n - (sigma/r)**6)",
+  "independent_variables": "r"
+}
--- a/gmso/lib/jsons/RyckaertBellemansTorsionPotential.json
+++ b/gmso/lib/jsons/RyckaertBellemansTorsionPotential.json
@@ -1,5 +1,5 @@
 {
   "name": "RyckaertBellemansTorsionPotential",
-  "expression": "c0 * cos(phi)**0 + c1 * cos(phi)**1 + c2 * cos(phi)**2 + c3 * cos(phi)**3 + c4 * cos(phi)**4 + c5 * cos(phi)**5",
+  "expression": "c0 + c1 * cos(phi)**1 + c2 * cos(phi)**2 + c3 * cos(phi)**3 + c4 * cos(phi)**4 + c5 * cos(phi)**5",
   "independent_variables": "phi"
 }


bc118:
We need to ask if we want to use the Phi variable for RB torsions, as it is used for OPLS and Periodic torsions.  RB torsion Phi is not the same as OPLS and Periodic:

RB-torsion-angle = OPLS-torsion-angle - Pi
RB-torsion-angle = Periodic -torsion-angle - Pi

Maybe we call the RB-torsion-angle Psi like the GROMACS manual.  If everyone agrees I will change it.

bc118:
Are we naming the HarmonicTorsionPotential.json potential correctly? It seems like we should change the name or equation, or add both and change them accordingly.  Thoughts?

It seems LAMMPS calls this a quadratic (https://docs.lammps.org/dihedral_quadratic.html#dihedral-style-quadratic-command)

LAMMPS calls this a harmonic (https://docs.lammps.org/dihedral_harmonic.html)

CalCraven:
> Are we naming the HarmonicTorsionPotential.json potential correctly? It seems like we should change the name or equation, or add both and change them accordingly. Thoughts?
> 
> It seems LAMMPS calls this a quadratic (https://docs.lammps.org/dihedral_quadratic.html#dihedral-style-quadratic-command)
> 
> LAMMPS calls this a harmonic (https://docs.lammps.org/dihedral_harmonic.html)
 
Gromacs cites this equation as a [harmonic improper dihedral](https://manual.gromacs.org/current/reference-manual/functions/bonded-interactions.html#improper-dihedrals-harmonic-type). So that might be where the current naming convention comes from.

bc118:
It is the form of an improper but it says dihedral/torsion.  So at minimum, this should be renamed.

CalCraven:
> We need to ask if we want to use the Phi variable for RB torsions, as it is used for OPLS and Periodic torsions. RB torsion Phi is not the same as OPLS and Periodic:
> 
> RB-torsion-angle = OPLS-torsion-angle - Pi = Phi - PI RB-torsion-angle = Periodic -torsion-angle - Pi = Phi - PI
> 
> Maybe we call the RB-torsion-angle Psi like the GROMACS manual (https://manual.gromacs.org/documentation/2019/reference-manual/functions/bonded-interactions.html). I changed it in this PR to Phi for the main equation.
> 
> If everyone agrees, I will change the other files accordingly. Thoughts? Comments? OK with this?

I agree with this point, being clear and giving these parameters different names hopefully will reduce errors from users missing this distinction. Along the same lines, if the above harmonic torsion potential is changed to an a harmonic improper, the independent variable should be &#958;.

I also think we need the proper dihedral form of the function, so I'm in favor of changing the name of this to a quadratic torsion and keeping the current form.

Side note, should be citing in these `.json` files some indication of where the functional form was pulled from?

gmso/tests/files/carbon.xml
```diff
@@ -15,32 +15,48 @@
...
+		<Parameter name="r_eq" value="0.1324"/>
+                <Parameter name="k"value="493460.96"/>
             </Parameters>
         </BondType>
     </BondTypes>
     <AngleTypes expression="0.5 * k * (theta-theta_eq)**2">
         <ParametersUnitDef parameter="theta_eq" unit="radian"/>
 	<ParametersUnitDef parameter="k" unit="kJ/(mol*radian**2)"/>
-        <AngleType name="AngleType1" type1='C' type2='C' type3="C">
+        <AngleType name="AngleType1" type1="'C" type2="'C'" type3="C">
```
daico007, line 28:
```suggestion
        <AngleType name="AngleType1" type1="C" type2="C" type3="C">
```

gmso/lib/jsons/PeriodicTorsionPotential.json
```diff
@@ -1,5 +1,5 @@
 {
   "name": "PeriodicTorsionPotential",
-  "expression": "k * (1 + cos(n * phi - phi_eq))",
+  "expression": "k0 + k1 * (1 + cos(1 * phi - phi_eq1)) + k2 * (1 + cos(2 * phi - phi_eq2)) + k3 * (1 + cos(3 * phi - phi_eq3)) + k4 * (1 + cos(4 * phi - phi_eq4)) + k5 * (1 + cos(5 * phi - phi_eq5))",
```
daico007, line 3: After spending some more thought on this, we might want to be careful with this eqn, especially regarding how many terms we want to support. As in, can we keep the old untouched and create a different/derivative for 5 terms, 4 terms, etc. More ideas/thoughts are welcome

bc118: Yeah, I hear what you mean.  I think, If you do not supported fixed terms it may be harder to parse when you don’t know what is there for the engines.  I believe We should be consistent, meaning if RB and OPLS is fixed this should be too.

bc118: I also don’t believe I saw anything ever past n=5 or power=5,  which are both there converting to one or the other.  The fixed term way in my mind is consistent for programming and the readers.

bc118: After additional though, if we decide to use the original periodic form, we should have additional check for the input variables. For example:

- The valuables must me k1, … kx, n1, …, nx or phi1, …, phix.  This will ensure the FFs are shareable.  

- k1 and phi1 must have n1=1 .  Kx and phix must have nx =x

Issue #624, opened 2021-12-21
The Trappe-UA and OPLSAA FF foyer to GMSO conversion seems to be missing the 1/2 factor for bonds and angles


bc118:
The Trappe-UA and OPLSAA FF foyer to GMSO conversion seem to be missing the 1/2 factor for bonds and angles.  

Basically, the Trappe-UA FF foyer to GMSO conversion K's are 2x what they should be for the bonds and angles.

Issue #617, opened 2021-12-16
1-4 bonding constants and combining rules may be outputting wrong


bc118:
- The 1-4 bonding constants (vdw and electro) seem to always output 0.5, when using:
new_structure_iteration.scaling_factors["vdw_14"]
new_structure_iteration.scaling_factors["coul_14"]

- The combining rules seem to be only outputting geometric.  Is this the default, and if so, where is it set in the GMSO XML?

I could also be doing something wrong, so please correct me if I am.

daico007:
For this issue, the 1-4 bonding constants should be fixed by https://github.com/mosdef-hub/foyer/pull/489 and the `combining_rule` issue should be fixed by https://github.com/mosdef-hub/gmso/pull/629

PR #629, opened 2022-01-19
Fix bugs in convert_foyer_xml

### LINKED ISSUE #628: Wrong expression in convert to foyer xml
The SymPy expressions are not consistent with foyer forcefield file.  E.g.,:

            "expression": "epsilon * ((sigma/r)**12 - (sigma/r)**6)",

https://github.com/mosdef-hub/gmso/blob/cc0f42d8cfda9b4236c0f1f4b724dc99036427f3/gmso/external/convert_foyer_xml.py#L249

These needs the factor of 4 out front.  Similarly harmonic potentials need the 1/2:

            "expression": "k * (r-r_eq)**2",

https://github.com/mosdef-hub/gmso/blob/cc0f42d8cfda9b4236c0f1f4b724dc99036427f3/gmso/external/convert_foyer_xml.py#L293

The output from, e.g., the LAMMPS writer is still correct, because it assumes the expression consistent with the original foyer XML file and does not directly examine the SymPy expression.

daico007:
Fix issues in `convert_foyer_xml` per discussion with @jbg, @bc118, @CalCraven  and @chrisiacovella. Fix bug where `combining_rule` info is not properly transfered. Fix wrong potential expression for nonbonded, harmonic bonds, and harmonic angles. Should fix #628, #624 and #617 (partially).

### DIFF
--- a/gmso/external/convert_foyer_xml.py
+++ b/gmso/external/convert_foyer_xml.py
@@ -252,7 +252,6 @@ def _write_nbforces(forcefield, ff_kwargs):
         forcefield,
         "AtomTypes",
         attrib_dict={
-
             "expression": "4 * epsilon * ((sigma/r)**12 - (sigma/r)**6)",
         },
     )
--- a/gmso/external/convert_foyer_xml.py
+++ b/gmso/external/convert_foyer_xml.py
@@ -246,7 +246,7 @@ def _write_nbforces(forcefield, ff_kwargs):
         forcefield,
         "AtomTypes",
         attrib_dict={
-            "expression": "epsilon * ((sigma/r)**12 - (sigma/r)**6)",
+            "expression": "4.0 * epsilon * ((sigma/r)**12 - (sigma/r)**6)",
         },
     )
     parameters_units = {"epsilon": "kJ/mol", "sigma": "nm"}
@@ -290,7 +290,7 @@ def _write_harmonic_bonds(forcefield, ff_kwargs):
         forcefield,
         "BondTypes",
         attrib_dict={
-            "expression": "k * (r-r_eq)**2",
+            "expression": "0.5 * k * (r-r_eq)**2",
         },
     )
 
@@ -323,7 +323,7 @@ def _write_harmonic_angles(forcefield, ff_kwargs):
         forcefield,
         "AngleTypes",
         attrib_dict={
-            "expression": "k * (theta - theta_eq)**2",
+            "expression": "0.5 * k * (theta - theta_eq)**2",
         },
     )
 
--- a/gmso/external/convert_foyer_xml.py
+++ b/gmso/external/convert_foyer_xml.py
@@ -134,12 +134,12 @@ def _write_gmso_xml(gmso_xml, **kwargs):
     """Given the set of keyword arguments, write a gmso Forcefield xml file."""
     forcefield = etree.Element("ForceField")
 
-    if kwargs.get("name"):
+    if kwargs.get("name") is not None:
         forcefield.attrib["name"] = kwargs.get("name")
     else:
         forcefield.attrib["name"] = pathlib.Path(gmso_xml).stem
 
-    if kwargs.get("version"):
+    if kwargs.get("version") is not None:
         forcefield.attrib["version"] = kwargs.get("version")
     else:
         forcefield.attrib["version"] = "0.0.1"
--- a/gmso/external/convert_foyer_xml.py
+++ b/gmso/external/convert_foyer_xml.py
@@ -363,7 +363,7 @@ def _write_ub_angles(forcefield, ff_kwargs):
         forcefield,
         "AngleTypes",
         attrib_dict={
-            "expression": "0.5 * k * (w - w_0) ** 2",
+            "expression": "1/2 * k * (w - w_0) ** 2",
         },
     )
 
--- a/gmso/external/convert_foyer_xml.py
+++ b/gmso/external/convert_foyer_xml.py
@@ -65,9 +65,11 @@ def from_foyer_xml(
     ff_root = foyer_xml_tree.getroot()
     name = ff_root.attrib.get("name")
     version = ff_root.attrib.get("version")
+    combining_rule = ff_root.attrib.get("combining_rule")
     f_kwargs = {
         "name": name,
         "version": version,
+        "combining_rule": combining_rule,
         "coulomb14scale": [],
         "lj14scale": [],
         "atom_types": [],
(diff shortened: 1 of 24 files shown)


daico007:
Pending unit tests

chrisiacovella:
Regarding the function expressions: Do we want to also change the Urey-Bradley terms to have a prefactor of 1/2 for consistency with how we define harmonic potentials in the Foyer XML?  

The standard CHARMM forcefield definition does not include the 1/2 term for UB terms (the factor of 1/2 is dropped from all harmonic terms...so NOT having 1/2 would be consistent with CHARMM, but not consistent with how you'd have to define the bonds and angles anyway).

daico007:
Good catch, I will udpate the U-B expression

bc118:
The current equation compare functions I have (not in PR yet) should scale any different input scaling factor the user inputs.  The only standard required should really be the equation, with any scaling factor.

chrisiacovella:
So this is converting from the old foyer xml files where there is an implicitly defined functional form (not explicitly stated).   But yeah, more generally, the writers need to be doing these checks for equivalence and doing the appropriate scaling (I'm working on that right now for the lammps writer).

bc118:
I would use my equation compare functions which will be in a PR soon to scale the K constants for the potentials.  They are here for now (https://github.com/bc118/mbuild/blob/gmso_gomc/mbuild/utils/gmso_equation_compare.py)

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                                / 2.0,
+                                bond_type.parameters["r_eq"]
+                                .in_units(u.Unit("angstrom"))
+                                .value,
+                            )
+                        )
+                    elif (
+                        scaling_factor := simplify(bond_expression)
+                        / simplify(bond_type.expression)
+                    ).is_real:
```
daico007, line 262: isn't the middle `elif` already covered by the next `elif` (where the scaling factor is calculated)

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                            .in_units(u.angstrom)
                             .value,
                         )
                     )
-
-            if topology.angles:
-                data.write("\nAngle Coeffs\n\n")
-                for idx, angle_type in enumerate(topology.angle_types):
+                elif (
+                    scaling_factor := simplify(lj_expression)
```
daico007, line 202: also, the walrus operator (`:=`) is a new thing in python >= 3.8, but since we are still trying to support python 3.7, this may be in appropriate. 
```suggestion
                    scaling_factor = simplify(lj_expression)
```

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                                .in_units(u.Unit("kcal/mol/angstrom**2"))
+                                .value
+                                / 2.0,
+                                bond_type.parameters["r_eq"]
+                                .in_units(u.Unit("angstrom"))
+                                .value,
+                            )
+                        )
+                    elif (
+                        scaling_factor := simplify(bond_expression)
```
daico007, line 260:
```suggestion
                        scaling_factor = simplify(bond_expression)
```

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                                .in_units(u.Unit("kcal/mol/radian**2"))
+                                .value
+                                / 2.0,
+                                angle_type.parameters["theta_eq"]
+                                .in_units(u.Unit("degree"))
+                                .value,
+                            )
+                        )
+                    elif (
+                        scaling_factor := simplify(angle_expression)
```
daico007, line 318:
```suggestion
                        scaling_factor = simplify(angle_expression)
```

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                                angle_type.parameters["k"]
+                                .in_units(u.Unit("kcal/mol/radian**2"))
+                                .value
+                                / 2.0,
+                                angle_type.parameters["theta_eq"]
+                                .in_units(u.Unit("degree"))
+                                .value,
+                            )
+                        )
+                    elif (
```
daico007, line 317: Similar to the comments above, the third `elif` should have covered the second `elif`

gmso/formats/lammpsdata.py
```diff
@@ -183,48 +183,159 @@ def write_lammpsdata(topology, filename, atom_style="full"):
...
+                    ).is_real:
+                        data.write(
+                            "{}\t{:.5f}\t{:.5f}\n".format(
+                                idx + 1,
+                                angle_type.parameters["k"]
+                                .in_units(u.Unit("kcal/mol/radian**2"))
+                                .value
+                                / scaling_factor,
+                                angle_type.parameters["theta_eq"]
+                                .in_units(u.Unit("degree"))
```
daico007, line 329: Need to double check the unit. Per the lammps docs (https://docs.lammps.org/angle_harmonic.html), it mentioned the angles are converted to radians internally, so i am not sure if K also need to be specified in kcal/mol/degree^2 (and the conversion will happen in lammps).

gmso/formats/lammpsdata.py
chrisiacovella: Yes, it is covered.  The scaling_factor calculation would work for all cases (I did the explicit way initially, then did the more general form in case someone defined these parameters in a non-standard, but otherwise equivalent form).  We could just switch completely to the scaling_factor.

gmso/formats/lammpsdata.py
chrisiacovella: Theta is degrees and is converted to radians after being read in by lamps; k is not converted in the code and needs to be per radian squared.

https://github.com/CFDEMproject/LAMMPS/blob/a92e8c271bcb7add8245f5a32f2920fbe35bacc9/src/MOLECULE/angle_harmonic.cpp#L178-L184

daico007: Good to know!

gmso/formats/lammpsdata.py
chrisiacovella: I removed this and just switched to using the scaling_factor expression.

gmso/formats/lammpsdata.py
chrisiacovella: same as above

chrisiacovella (approved):

gmso/external/convert_foyer_xml.py
```diff
@@ -357,7 +363,7 @@ def _write_ub_angles(forcefield, ff_kwargs):
         forcefield,
         "AngleTypes",
         attrib_dict={
-            "expression": "k * (w - w_0) ** 2",
+            "expression": "0.5 * k * (w - w_0) ** 2",
```
justinGilmer, line 366: We should be consistent with the above or we go with 0.5 everywhere
```suggestion
            "expression": "1/2 * k * (w - w_0) ** 2",
```

gmso/external/convert_foyer_xml.py
```diff
@@ -132,17 +134,21 @@ def _write_gmso_xml(gmso_xml, **kwargs):
     """Given the set of keyword arguments, write a gmso Forcefield xml file."""
     forcefield = etree.Element("ForceField")
 
-    if kwargs.get("name") is not None:
+    if kwargs.get("name"):
```
justinGilmer, line 137: Lets revert this back so we dont get burned by "falsy" values

daico007: done

bc118 (approved):
This was tested with GOMC-GMSO-MoSDeF and all but the combining rule are passing.  I am approving this as I created a new issue for the combining rules (https://github.com/mosdef-hub/gmso/issues/631).

justinGilmer (approved):

PR #635, opened 2022-02-03
Prevent accidental defaults values while creating Potentials

### LINKED ISSUE #634: Prevent accidental defaults while creating potentials
Let's say I want to create an `AtomType`, with the current setup/expression verification logic, the following statement is completely valid:

```py
>>> from gmso import AtomType
>>> import unyt as u
>>> at = AtomType()
>>> at.expression
4*epsilon*(-sigma**6/r**6 + sigma**12/r**12)
>>> at.parameters
{'sigma': unyt_quantity(0.3, 'nm'), 'epsilon': unyt_quantity(0.3, 'kJ')}
>>> at.independent_variables
{r}
```

The above might be a great default to have. However, problem arises when you want to do something like:
```py
at = AtomType(expression="epsilon*sigma+r")
>>> at.parameters
{'sigma': unyt_quantity(0.3, 'nm'), 'epsilon': unyt_quantity(0.3, 'kJ')}

```
This problem is simple yet tedious to handle. On the one hand these defaults provide a handy way to create new Objects while sacrificing one important issue: `unintended defaults`. So, while if any the `parameters`/`expression`/`independent_variables` provided are not-inclined, proper checks happen and an error is raised, my concern is for those where these values matchup. I would argue that we should remove ability to provide an expression without providing both `independent_variables` as well as `parameters`.

umesh-timalsina:
Closes #634.

### DIFF
--- a/gmso/abc/abstract_potential.py
+++ b/gmso/abc/abstract_potential.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, Iterator, List
 from pydantic import Field, validator
 
 from gmso.abc.gmso_base import GMSOBase
-from gmso.utils.expression import _PotentialExpression
+from gmso.utils.expression import PotentialExpression
 
 
 class AbstractPotential(GMSOBase):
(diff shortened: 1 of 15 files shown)


daico007 (approved):
LGTM!

justinGilmer (approved):

PR #636, opened 2022-02-11
fix typo when converting foyer XML

daico007:
The naming mismatch (`combining_rule` vs `combiningRule`) causing issues converting combining rule information from foyer XML to gmso XML. This PR fix this by sticking to `combiningRule`.


daico007:
This should fix the second portion of #617

justinGilmer (changes requested):
Can you add a test to ensure that the combining rule is being found properly now?

After that, it should be g2g

daico007:
unit test added

justinGilmer (approved):

PR #644, opened 2022-04-14
Atomtype parameterization methods for parameterizing a topology in GMSO.

CalCraven:
This PR will add the atomtyping module to GMSO for passing a GMSO.Forcefield
and a GMSO.Topology and match atomtype using foyer as the backend. Then,
the corresponding connection types will be found in the Forcefield and
applied to the connections in the topology.

* Create parameterize.py, which has the apply function which can take a
topology, and a gmso forcefield to apply to it. This can use subgraph
isomorphism to identify molecular structures in the topology through
the bondgraph and bin those into unique molecules that match a specified
forcefield. This apply function can also do the standard atomtyping of
the entire topology in one step.
* Create isomorph.py which uses networkx graphs to identify disconnected
components and isomorphism to identify repeated structures.

~~Checklist:~~
~~- [x] Use site residues to type repeated structures~~
~~- [ ] Write a function that will add residue information to specified sites~~
~~- [ ] Evaluate docstring clarity~~
~~- [ ] Write unit tests~~ 
~~- [x] Match pre-commit style~~

## Updated Checklist:
- [x] Feature Parity with Parmed
- [x] Multi--Forcefield application
- [x] Isomorphic Test


gmso/atomtyping/isomorph.py
```diff
@@ -0,0 +1,59 @@
+"""TopologyGraph Functions that identify molecules from isomorphism."""
+from collections import deque
+
+import networkx as nx
+
+
+def partition_isomorphic_topology_graphs(graph):
+    """Return a dictionary of subgraphs that are partitioned by isomorphism.
```
umesh-timalsina, line 8:
```suggestion
    """Return a collection of isomorphic sets of subgraph of the Topology Graph.
```

gmso/atomtyping/isomorph.py
```diff
@@ -0,0 +1,59 @@
...
+    first_mismatch = None
+    while graph_queue:
+        if graph_queue[0] == first_mismatch:
+            count = 0
+            graph_of_interest = graph_queue.popleft()
+            isomorphic_elements[graph_of_interest] = []
+        if graph_queue:
+            graph = graph_queue.popleft()
+            matcher = nx.algorithms.isomorphism.GraphMatcher(
+                graph, graph_of_interest
```
umesh-timalsina, line 48: I think we need to add a node_match function (based on the element of the node) IIRC.

umesh-timalsina, line 48: I think we need to add a node_match function (based on the element of the node) IIRC.

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+        bonds.
+    assert_angle_params : bool, optional, default=True
+        If True, Foyer will exit if parameters are not found for all system
+        angles.
+    assert_dihedral_params : bool, optional, default=True
+        If True, Foyer will exit if parameters are not found for all system
+        proper dihedrals.
+    assert_improper_params : bool, optional, default=False
+        If True, Foyer will exit if parameters are not found for all system
+        improper dihedrals.
```
umesh-timalsina, line 53:
```suggestion
    assert_bond_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        bonds.
    assert_angle_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        angles.
    assert_dihedral_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        proper dihedrals.
    assert_improper_params : bool, optional, default=False
        If True, an error is raised if parameters are not found for all system
        improper dihedrals.
```

umesh-timalsina, line 53:
```suggestion
    assert_bond_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        bonds.
    assert_angle_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        angles.
    assert_dihedral_params : bool, optional, default=True
        If True, an error is raised if parameters are not found for all system
        proper dihedrals.
    assert_improper_params : bool, optional, default=False
        If True, an error is raised if parameters are not found for all system
        improper dihedrals.
```

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    top.identify_connections()
```
umesh-timalsina, line 55: Any thoughts on adding a flag for this operation?

umesh-timalsina, line 55: Any thoughts on adding a flag for this operation?

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+    top_graph = get_modified_topology_graph(top)
+    if identify_connected_components:  # True, (True,False)
+        # populate molecule information
+        isomorphic_top_graphs = partition_isomorphic_topology_graphs(top_graph)
+        apply_atomtypes_from_isomorphic(
+            top, isomorphic_top_graphs, forcefields
+        )  # what if only one forcefield?, traversing done in this function
```
umesh-timalsina, line 62: Can you elaborate on the question posed in this comment?

umesh-timalsina, line 62: Can you elaborate on the question posed in this comment?

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+    elif not use_residue_info:  # False, False
+        # get entire topgraph
+        topgraph = get_modified_topology_graph(top)
+        # get rules provider from gmso_ff
+        # Assumes forcefields is just one forcefield
+        at_rules_provider = get_atomtyping_rules_provider(gmso_ff=forcefields)
+        # create a typemap of that topgraph
+        typemap = find_atomtypes(topgraph, at_rules_provider)
+        # apply typemap
+        traverse_typemap(top, typemap, forcefields)
```
umesh-timalsina, line 72: I don't think this will work for a dictionary of forcefields with more than one key right?

umesh-timalsina, line 72: I don't think this will work for a dictionary of forcefields with more than one key right?

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+    connect_members = {
+        "bond": [0, 1],
+        "angle": [0, 1, 2],
+        "dihedral": [0, 1, 2, 3],
+        "improper": [0, 1, 2, 3],
+    }  # TODO validate improper order
+    for connect in getattr(top, "_" + parameter + "s"):
+        sub_top_label = connect.connection_members[0].label_
+        ff_of_interest = forcefields[sub_top_label]
+        type_getter = attrgetter("atom_type.atomclass")
```
umesh-timalsina, line 267: Is a None-Type check prudent here? I think foyer will raise an error if it cannot find an atomtype right?

umesh-timalsina, line 267: Is a None-Type check prudent here? I think foyer will raise an error if it cannot find an atomtype right?

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,275 @@
...
+    top_graph = TopologyGraph()
+    for atom in gmso_topology.sites:
+        if isinstance(atom, Atom):
+            if atom.name.startswith("_"):
+                top_graph.add_atom(
+                    name=atom.name,
+                    index=gmso_topology.get_index(atom),
+                    atomic_number=None,
+                    element=atom.name,
+                    subtopology_label=atom.label,
```
umesh-timalsina, line 128: Not sure where we want to address this but we need a better identifier than `atom.label`.

umesh-timalsina, line 128: Not sure where we want to address this but we need a better identifier than `atom.label`.

umesh-timalsina (commented):
This is shaping up nicely @CalCraven. I have added some comments by skimming through it.

gmso/atomtyping/parameterize.py
CalCraven: So only identify connections if those connections were missing? Makes sense to me, do you want it default flagged on or off?

I do start to get worried about the number of flags that are starting to get passed around.

gmso/atomtyping/parameterize.py
CalCraven: This question was a note to myself to handle the case that only one forcefield is passed, without specifying the molecules it should be attached to. Currently, that will just spit out an error since it expects `forcefields` to be of type dict. So we should either tell users in this case to use identify_connected_components=False, or just automatically use the single forcefield that was passed.

gmso/atomtyping/parameterize.py
CalCraven: I agree. I will think about this, and see if there is something better we can be using.

bc118: Should not the atom be the atom number in the topology?

gmso/atomtyping/isomorph.py
CalCraven: What exactly would a node_match function do? I'm not entirely familiar, is it something passed to the GraphMatcher to do the node matching?

daico007: just tested this offline, I think the current use of GraphMatcher only concern about the shape of the graph but not about the nodes involved. So this could cause issue with molecules with same shape but different element would still be recognized as isomorphic (so lumped into the same group).

daico007:
```python3 
ethanol = mb.load("CCO", smiles=True)
ethanol.name = "Ethanol"

eth1 = mb.clone(ethanol)
eth2 = mb.clone(ethanol)

eth1_top = mb_convert(eth1)
eth1_top.sites[0].name = "K"
eth1_top.sites[0].element = element_by_symbol("K")
eth2_top = mb_convert(eth2)
```
Note below I am using the `get_modified_topology_graph` from the example notebook.
```python3
eth1_grp = get_modified_topology_graph(eth1_top)
eth2_grp = get_modified_topology_graph(eth2_top)
matcher = nx.isomorphism.GraphMatcher(eth1_grp, eth2_grp)
```
Then 
`matcher.is_isomorphic()` will still return `True`

daico007: on a different note, I think we can use this `GraphMatcher` for the issue with the assumed site order in a molecule (or subtopology) (also this https://github.com/mosdef-hub/foyer/issues/504). Will have to sacrifice some performance for checking but the number of checks performed will still be fewer that going the completely splitting the topology, so should still be a bit faster.

gmso/atomtyping/parameterize.py
daico007: I think we will need a flag for this operation if we want to keep it here. But, I am thinking, should we ask the user to perform the `identify_connection` before arriving at this method? Since the `identify_connection` itself may need some adjustment (like do we want to identify improper dihedrals or not).

gmso/atomtyping/isomorph.py
CalCraven: Okay this makes more sense to me now, thanks @daico007. I was a little hazy on the exact implementation of GraphMatcher. I think having a few different options for matching makes sense, which trade of increasing performance for increasing matching precision. This reminds me a bit of the filters used in PR #649, which can be used to specify how you want to filter unique parameter types. I think it makes sense to at least try a basic one that checks at least for elements in the correct positions, and allow that to be a built in option.

gmso/atomtyping/parameterize.py
CalCraven: The index is the atom_number in the topology. The subtopology_label should be the string that we use to apply a given forcefield to the sites with this label.

CalCraven: According to PR #638, this info will be stored in site.molecule. I think whatever convention we decide to use there can be the reference label we use to identify the molecule name, and therefore the identifier we use for the forcefield

gmso/atomtyping/parameterize.py
CalCraven: Added an error message in e1f3d4e.

gmso/atomtyping/parameterize.py
CalCraven: Fixed in e1f3d4e.

gmso/atomtyping/isomorph.py
CalCraven: Fixed in e1f3d4e.

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    connect_members = {
+        "bond": [0, 1],
+        "angle": [0, 1, 2],
+        "dihedral": [0, 1, 2, 3],
+        "improper": [0, 1, 2, 3],
+    }  # TODO validate improper order
+    molecule_label = "label_"  # place to grab info referencing molecule name
+    for connect in getattr(top, "_" + parameter + "s"):
+        sub_top_label = connect.connection_members[0].get(molecule_label)
```
daico007, line 269:
```suggestion
        sub_top_label = connect.connection_members[0].__dict__.get(molecule_label)
```
or 
```suggestion
        sub_top_label = connect.connection_members[0].molecule_label
```

gmso/atomtyping/parameterize.py
```diff
@@ -0,0 +1,284 @@
...
+        if sub_top_label is None:
```
daico007, line 270:
```suggestion
        if not sub_top_label:
```
I think this would also catch the case where the `sub_top_label` is empty string.

CalCraven:
Rename module to to parameterization, not atomtyping

gmso/parameterization/subtopology_utils.py
```diff
@@ -0,0 +1,50 @@
+"""Utilities for application of a particular forcefield to a subtopology."""
+
+
+def _members_in_subtop(connection, subtop):
+    """Check if all the members in a connection belong to a subtopology."""
+    return all(site in subtop.sites for site in connection.connection_members)
+
+
+def _subtop_connections(subtop, attr):
```
bc118, line 9: I like this a lot, a much-needed function.  Awesome.

gmso/tests/parameterization/test_subtopology_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+            expected_members = {
+                ("C", "C", "H", "H"),
+                ("C", "H", "H", "C"),
+                ("C", "H", "H", "H"),
+                ("C", "H", "C", "H"),
+            }
+            assert all(
+                a_member in expected_members for a_member in improper_members
+            )
```
bc118, line 110: We should consider adding a test remove all the non-parameterized impropers, because that will be done for nearly every engine

bc118 (commented):
This seems to work on my few mosdef-gomc-gmso test cases.  We still need to get it working for beads, but outside this PR.  

Please see and consider the comments listed, but otherwise LGTM.

bc118 (approved):

gmso/parameterization/foyer_utils.py
```diff
@@ -0,0 +1,106 @@
...
+    gmso_topology: gmso.Topology-like
+        The GMSO Topology
+
+    atomdata_populator: callable, default=None
+        A function that will be called with the following arguments `gmso_topology` as well as `atom` to pass extra
+        arguments to the foyer.topology_graph.AtomData object
+
+    Notes
+    -----
+    The gmso topology here is a duck type.
```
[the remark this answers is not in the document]
daico007: ```Duck typing is a concept related to dynamic typing, where the type or the class of an object is less important than the methods it defines. When you use duck typing, you do not check types at all. Instead, you check for the presence of a given method or attribute.``` ([Duck Typing - Real Pythonhttps://realpython.com › lessons › duck-typing](https://realpython.com/lessons/duck-typing/#:~:text=Duck%20typing%20is%20a%20concept,a%20given%20method%20or%20attribute.)) 

TIL

PR #645, opened 2022-04-18
Update method to load in mol2 file

daico007:
We encountered a flaky test over mbuild when reading in the [beta-cristabolite mol2](https://github.com/mosdef-hub/mbuild/blob/master/mbuild/lib/surfaces/beta-cristobalite-expanded.mol2). The issue happens when parsing the bond section, which sometime misses a bond, which we are suspecting due to some racing condition (?). I was able to reproduce this issue when reading the file in a loop `for i in range(100)`, and assert the number of bond. This PR update the method to read in the mol2 file to parse the whole file at the beginning to the list and parse each section accordingly.

### DIFF
--- a/gmso/formats/mol2.py
+++ b/gmso/formats/mol2.py
@@ -33,6 +33,7 @@ def from_mol2(filename, site_type="atom"):
 
     Notes
     -----
+    The position of atom in the mol2 file is assumed to be in Angstrom.
     It may be common to want to create an mBuild compound from a mol2 file. This is possible
     by installing [mBuild](https://mbuild.mosdef.org/en/stable/index.html)
     and converting using the following python code:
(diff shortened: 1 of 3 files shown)


gmso/formats/mol2.py
```diff
@@ -51,165 +51,88 @@ def from_mol2(filename, site_type="atom"):
...
-        while f:
-            # check for header character in line
-            if line.strip().startswith("@<TRIPOS>"):
-                # if header character in line, send to a function that will direct it properly
-                line = _parse_record_type_indicator(
-                    f, line, topology, site_type
+        with open(filename, "r") as f:
+            fcontents = f.readlines()
+        sections = dict()
+        key = "meta"
```
CalCraven, line 57: I think this line needs to be 
section_key = "meta"

gmso/formats/mol2.py
```diff
@@ -51,165 +51,88 @@ def from_mol2(filename, site_type="atom"):
...
+    parse_ele = {
+        "lj": lambda ele: None,
+        "atom": lambda ele: element_by_symbol(ele)
+        if element_by_symbol(ele)
+        else element_by_name(ele),
+    }
+    for line in section:
+        cont = line.split()
+        position = [float(x) for x in cont[2:5]] * u.Å
+        element = parse_ele[site_type](cont[5])
```
CalCraven, line 97: Maybe worthwhile to put a note at the top that mol2 positions are assumed to be in angstroms

gmso/formats/mol2.py
```diff
@@ -51,165 +51,88 @@ def from_mol2(filename, site_type="atom"):
...
-                )
-                charge = None
-            atom = Atom(
-                name=line[1],
-                position=position.to("nm"),
-                charge=charge,
-                element=element,
-                residue_name=line[7],
-                residue_number=int(line[6]),
+def _parse_atom(top, section, site_type="atom"):
```
CalCraven, line 86: Instead of passing around kwargs to every function that doesn't need this site_type argument, it might be better to have a function named _parse_lj_site(top, section) that is separate from the _parse_atom function. Then we can add that to the supported_rti as:

```python
        site_parser = _parse_lj if atom_style == "lj" else _parse_atom
        supported_rti = {
            "@<TRIPOS>ATOM": site_parser,
            "@<TRIPOS>BOND": _parse_bond,
            "@<TRIPOS>CRYSIN": _parse_box,
            "@<TRIPOS>FF_PBC": _parse_box,
        }
```

CalCraven (changes requested):
This looks much nicer. Just need to get tests passing.

gmso/formats/mol2.py
daico007: good catch!

gmso/formats/mol2.py
daico007: added separate method to parse `lj`

gmso/formats/mol2.py
daico007: added to comment

gmso/formats/mol2.py
```diff
@@ -49,52 +50,76 @@ def from_mol2(filename, site_type="atom"):
...
+        "@<TRIPOS>FF_PBC": _parse_box,
+    }
+    for section in sections:
+        if section not in supported_rti:
+            warnings.warn(
+                f"The record type indicator {section} is not supported."
+            )
+        else:
+            supported_rti[section](
+                topology, sections[section], site_type="atom"
```
CalCraven, line 77: Just need to remove site_type argument now to this function call. LGTM after this.

CalCraven (approved):

PR #649, opened 2022-04-27
Add view classes for handling potentials in a topology

umesh-timalsina:
To Do:
- [x] Topology Changes and gmso (formats, external) module checks.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -329,9 +329,9 @@ class Topology(object):
         Notes
         -----
         This returns a TopologyPotentialView object which can be used as
-        an iterator. By default this will return a view with all the atom_types
-        in the topology (repeats even if multiple sites point to the same atomtype
-        object. Use, different filters(builtin or custom) to suit your needs.
+        an iterator. By default, this will return a view with all the atom_types
+        in the topology (if multiple sites point to the same atom_type, only a
+        single reference is returned/iterated upon). Use, different filters(builtin or custom) to suit your needs.
         See examples below.
 
         Examples
@@ -347,9 +347,9 @@ class Topology(object):
         ...     atom.atom_type = atom_type
         ...     top.add_site(atom)
         >>> len(top.atom_types)
-        100
-        >>> len(top.atom_types(filter_by=PotentialFilters.UNIQUE_ID))
         1
+        >>> len(top.atom_types(filter_by=PotentialFilters.REPEAT_DUPLICATES))
+        100
         >>> len(top.atom_types(filter_by=PotentialFilters.UNIQUE_NAME_CLASS))
         1
 
@@ -397,9 +397,9 @@ class Topology(object):
         Notes
         -----
         This returns a TopologyPotentialView object which can be used as
-        an iterator.By default this will return a view with all the bond_types
-        in the topology (repeats even if multiple bonds point to the same bond_type
-        object). Use, different filters(builtin or custom) to suit your needs.
+        an iterator.By default, this will return a view with all the bond_types
+        in the topology (if multiple bonds point to the same bond_type, only a
+        single reference is returned/iterated upon). Use, different filters(builtin or custom) to suit your needs.
         See examples below.
 
         Examples
@@ -446,9 +446,9 @@ class Topology(object):
         Notes
         -----
         This returns a TopologyPotentialView object which can be used as
-        an iterator. By default this will return a view with all the angle_types
-        in the topology (repeats even if multiple angles point to the same
-        angle_type object). Use, different filters(builtin or custom) to suit
+        an iterator. By default, this will return a view with all the angle_types
+        in the topology (if multiple angles point to the same angle_type, only a
+        single reference is returned/iterated upon). Use, different filters(builtin or custom) to suit
         your needs. See examples below.
 
         Examples
@@ -496,10 +496,10 @@ class Topology(object):
         Notes
         -----
         This returns a TopologyPotentialView object which can be used as
-        an iterator. By default this will return a view with all the dihedral_types
-        in the topology (repeats even if multiple dihedrals point to the same
-        dihedral_type object). Use, different filters(builtin or custom) to suit
-        your needs. See examples below.
+        an iterator. By default, this will return a view with all the dihedral_types
+        in the topology (if multiple dihedrals point to the same dihedral types, only a
+        single reference is returned/iterated upon). Use, different filters(builtin or custom)
+        to suit your needs. See examples below.
 
         Examples
         --------
@@ -549,10 +549,10 @@ class Topology(object):
         Notes
         -----
         This returns a TopologyPotentialView object which can be used as
-        an iterator. By default this will return a view with all the improper_types
-        in the topology (repeats even if multiple impropers point to the same
-        improper_type object). Use, different filters(builtin or custom) to suit
-        your needs. See examples below.
+        an iterator. By default, this will return a view with all the improper_types
+        in the topology (if multiple impropers point to the same improper_type, only a
+        single reference is returned/iterated upon). Use, different filters(builtin or custom) to
+        suit your needs. See examples below.
 
         Examples
         --------
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -320,7 +320,7 @@ class Topology(object):
     @property
     def impropers(self):
         """Return all impropers in the topology."""
-        return TopologyPotentialView(self._impropers)
+        return self._impropers
 
     @property
     def atom_types(self):
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1,4 +1,5 @@
 """Base data structure for GMSO chemical systems."""
+import itertools
 import warnings
 from pathlib import Path
 
@@ -18,7 +19,7 @@ from gmso.core.dihedral_type import DihedralType
 from gmso.core.improper import Improper
 from gmso.core.improper_type import ImproperType
 from gmso.core.pairpotential_type import PairPotentialType
-from gmso.core.parametric_potential import ParametricPotential
+from gmso.core.views import TopologyPotentialView
 from gmso.exceptions import GMSOError
 from gmso.utils.connectivity import (
     identify_connections as _identify_connections,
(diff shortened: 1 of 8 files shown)


gmso/core/topology.py
```diff
@@ -630,17 +581,16 @@ def add_subtopology(self, subtop, update=True):
...
+            next(self.atom_types)
+        except StopIteration:
+            try:
+                next(self.connection_types)
+            except StopIteration:
+                return False
+        return True
+
+    def is_fully_typed(self, group="topology"):
         """Check if the topology or a specifc group of objects that make up the topology are fully typed
```
CalCraven, line 594: """Check if the topology or a specific group of objects that make up the topology are fully typed

CalCraven (approved):
This looks pretty awesome. I'm going to play around with it some to see how long it takes to generate unique types, and try out some different filters for unique potentials.

gmso/core/views.py
```diff
@@ -0,0 +1,143 @@
...
+    for potential_type in potential_types:
+        identifier = get_identifier(potential_type)
+        if identifier not in visited[type(potential_type)]:
+            visited[type(potential_type)].add(identifier)
+
+            yield potential_type
+
+
+class PotentialFilters:
+    UNIQUE_NAME_CLASS = "unique_name_class"
```
CalCraven, line 43: Can we make other potential filters called "UNIQUE_EXPRESSION_CLASS" and "UNIQUE_PARAMETER_CLASS"?

```python
def get_expression(potential):
    """Get sympy expression for a topology potential."""

    return potential.expression

def unique_potentials_by_expression(potential_types):
    """Filter unique potentials based on pre-defined sympy expressions."""
    visited = defaultdict(set)
    for potential_type in potential_types:
        identifier = get_expression(potential_type)
        if identifier not in visited[type(potential_type)]:
            visited[type(potential_type)].add(identifier)

            yield potential_type

def get_parameters(potential):
    """Get sympy expression for a topology potential."""

    return (
        tuple(potential.get_parameters().keys()), 
        tuple(map(lambda x: x.to_value(), my_ex.get_parameters().values()))
    )

def unique_potentials_by_parameters(potential_types):
    """Filter unique potentials based on pre-defined sympy expressions."""
    visited = defaultdict(set)
    for potential_type in potential_types:
        identifier = get_parameters(potential_type)
        if identifier not in visited[type(potential_type)]:
            visited[type(potential_type)].add(identifier)

            yield potential_type
```

gmso/core/views.py
```diff
@@ -0,0 +1,143 @@
...
+
+def get_identifier(potential):
+    """Get identifier for a topology potential."""
+    if isinstance(potential, AtomType):
+        return potential.name
+    if isinstance(potential, (BondType, AngleType, DihedralType, ImproperType)):
+        return potential.member_types or potential.member_classes
+
+
+def unique_potentials(potential_types):
```
CalCraven, line 31: I would rename this function to unique_potentials_by_name, to be clear that's what you're considering a unique_potential.

gmso/core/views.py
```diff
@@ -0,0 +1,143 @@
...
+potential_attribute_map = {
+    Atom: "atom_type",
+    Bond: "bond_type",
+    Angle: "angle_type",
+    Dihedral: "dihedral_type",
+    Improper: "improper_type",
+}
+
+
+def get_identifier(potential):
```
CalCraven, line 23: We can potential condense some of these identifiers (see my other suggested ones below) by adding some flags to this function. Or we can rename this function to "get_parameter_names" which may be more indicative of what this function is doing.

gmso/core/views.py
```diff
@@ -0,0 +1,143 @@
...
+    """Get identifier for a topology potential."""
+    if isinstance(potential, AtomType):
+        return potential.name
+    if isinstance(potential, (BondType, AngleType, DihedralType, ImproperType)):
+        return potential.member_types or potential.member_classes
+
+
+def unique_potentials(potential_types):
+    """Filter unique potentials based on pre-defiend identifiers."""
```
CalCraven, line 32: typo "pre-defined"

gmso/core/views.py
```diff
@@ -0,0 +1,143 @@
...
+    Parameters
+    ----------
+    iterator: typing.Iterator, required=True
+        An iterator of either topology sites or connections from which to extract potentials from
+
+    Parameters
+    ----------
+    filter_by: str or func, default=None
+        If provided, filter the collected potentials by some function
+        see, default_filters for names of the default potential filters
```
CalCraven, line 61: I think we should show a case of using a filter_by, just to be clear where you might import these default_filter from. People with less python familiarity won't be totally clear how this filter is being applied, so showing the basic one PotentialFilters.UNIQUE_NAME_CLASS is probably the best bet.

gmso/tests/test_views.py
```diff
@@ -0,0 +1,12 @@
+from gmso.core.views import PotentialFilters
+from gmso.tests.base_test import BaseTest
+
+
+class TestViews(BaseTest):
+    def test_view_atom_types_typed_ar_system(self, n_typed_ar_system):
+        atom_types = n_typed_ar_system().atom_types(repeat=True)
+        assert len(atom_types) == 100
+        atom_types_unique = n_typed_ar_system().atom_types(
```
CalCraven, line 9: Add this same test, but with repeat=True and a filter. The filter should cover this. What if filter=False and a filter_by method is passed?

CalCraven (changes requested):
Looking great. I tested this on 1000 ethanol molecules. It takes under a hundredth of a second to get the unique atom_types, which is crazy compared to what it was before. Will be cool to provide a few additional filters for people, and some docs on how to use them. 

Last thing, when I run 
```
from gmso.external import from_parmed
applied_top = from_parmed(pmd_ethanol)
```

I get a bunch of `True` (12 to be exact) printed to my output. This might be an extra print statement somewhere, but I don't think this is intended behavior.

A script to replicate this behavior:
```python
import mbuild as mb
import foyer
from gmso.external import from_parmed

mol = mb.load("CCO", smiles=True)
filled_box = mb.packing.fill_box(mol, 1, density=0.8)
ff = foyer.Forcefield(name="oplsaa")
pmd_ethanol = ff.apply(filled_box)
applied_top = from_parmed(pmd_ethanol)
```

gmso/core/views.py
umesh-timalsina: I agree. But I think even better place to put these examples will be in the property decorator's docstring in topology class.

gmso/core/views.py
umesh-timalsina: Yeah, having a good filteration system is crucial. I can add these methods, with a more sophisticated way of handling filters.

umesh-timalsina: Added this

daico007 (approved):
This looks great!

gmso/core/views.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __iter__(self):
+        yield from self.yield_view()
+
+    def index(self, item):
+        for j, potential in enumerate(self.yield_view()):
+            if potential is item:
+                return j
+
+    def _collect_potentials(self):
```
CalCraven, line 131: I think the point was to use the filters to get the number of unique potentials. But if you just looked at len(gmso.Topology.atom_types()), you would get the number of potentials with no filtering applies. See the code below as an example.

```python
import mbuild as mb
from foyer import Forcefield
from gmso.external import from_parmed
ff = Forcefield(name="oplsaa")
mol = mb.load("CCO", smiles=True)
filled_box = mb.packing.fill_box(mol, 1, density=0.8)
applied_top = ff.apply(filled_box)
applied_top = from_parmed(applied_top)
print(len(applied_top.atom_types))

5
```

But the number of atomtypes should be 9, one for each sites. This is akin to repeats=False essentially.

gmso/core/views.py
```diff
@@ -0,0 +1,180 @@
...
+        """Collect potentials from the iterator"""
+        visited = set()
+        for item in self.iterator:
+            potential = getattr(
+                item, potential_attribute_map[type(item)]
+            )  # Since this use is internal, KeyErrors N/A
+            if potential and id(potential) not in visited:
```
CalCraven, line 138: For the above behavior, I think you would just get rid of this `if id(potential) not in visited` clause.

gmso/core/views.py
daico007: I think this can be achieved by `applied_top.atom_types(filter_by=PotentialFilters.UNIQUE_NAME_CLASS`
So I think we just need to clarify this in the docstring?

umesh-timalsina: Yeah. This has been an issue. However, a repeat behavior is akin the following. Essentially, should an atomtype that is associated with multiple sites, contribute to overall length
```python
atom_types_repeated = filter(lambda p: p is not None, map(lambda s: s.atom_type, top._sites))
```

CalCraven: @bc118 do you have any thoughts here? I think the more common thing would be to just be worried about unique potentials. But currently, someone has to use the above workaround @umesh-timalsina posted to get all of the potentials. So if that's behavior users may need, it may be at least useful write that in. But I'm also okay with the way it's written currently, and if down the line someone wants that feature, it's easy to write a filter that will return that info.

bc118: Many users will want only the unique force field parameters, and not care about the repeating ones (GOMC, LAMMPS...).  I'm sure someone will also want all of them though, at some point.  I am personally, more concerned about the former.

umesh-timalsina: @bc118. I definitely agree to and second your points. However, uniqueness here is engine/context and use-case specific. So, if we give it some thought, the idea behind these filters makes intuitive sense atleast to me. With this, we are essentially deferring the use of these filters to our writers or any end user using this package. With a reasonable default and some good builtin filters.

```py
class PotentialFilters:
    UNIQUE_NAME_CLASS = "unique_name_class"  # Potential Name/Class (Uniqueness as incorporated in our workflows)
    UNIQUE_EXPRESSION = "unique_expression"  # No Clear path but can be leveraged in more rigorous sympy checks for uniquness
    UNIQUE_PARAMETERS = "unique_parameters" # No Clear path but can be leveraged in more rigorous sympy checks for uniquness
    UNIQUE_ID = "unique_id"  # Default behavior (I.e. provide all the ones with unique object ids)
    REPEAT_DUPLICATES = "repeat_duplicates" # Last use case for duplcation

    @staticmethod
    def all():
        return set(
            f"{PotentialFilters.__name__}.{k}"
            for k, v in PotentialFilters.__dict__.items()
            if not k.startswith("__") and not callable(v)
        )


```

CalCraven (approved):
LGTM!

bc118:
This seems to be working to pull only the unique bond, angle, dihderal, and improper FF parameters and info, when using
filter_by=PotentialFilters.UNIQUE_NAME_CLASS.

PR #658, opened 2022-05-24
Include ImproperType reading in "from_parmed"

CalCraven:
This PR is branched from PR #644. The parmed conversion to GMSO
topology was not able to read in various Improper connections to
GMSO. It would aggregate them with the Dihedral connections, from
when GMSO did not have explicit treatment of impropers, similar to
ParmEd. This PR aims to check a ParmEd object for structure.impropers
or structure.dihedrals with the dihedral.impropers flag set to
True and convert those to `gmso.Impropers`. Likewise, it looks to
read in periodic or harmonic impropers and generate the correct
`gmso.ImproperType` parametric potential for that type, if the flag
`refer_types` is set to True in the `from_parmed` call.

* Add pmd_improper_types_map function to convert_parmed
* Generate a parametric potential with harmonic type
expression from pmd.impropers
* Generate a parametric potential with periodic type
expression from pmd.dihedrals and dihedral.improper=True
* Validate testing with foyer opls_validation when using
functions in PR #644

Harmonic type expression for impropers is taken from
https://manual.gromacs.org/current/reference-manual/functions/bonded-interactions.html#improper-dihedrals-harmonic-type
This uses Xi as the independent variable. If someone testing this PR can
validate that this is the formalism we want to use when reading in
ImproperTypes from ParmEd improper_types, that would be a good sanity
check.
The periodic type expression is taken from
https://manual.gromacs.org/current/reference-manual/functions/bonded-interactions.html#equation-eqnperiodicpropdihedral
with phi as the independent variable.


daico007 (approved):
I think this PR looks good! I believe the new features is tested implicitly in the `test_opls_gmso.py` but I think it may worth making a few tests to explicitly tests the new features (look at one system, and make sure all the impropers converted make sense).

bc118 (approved):
This looks like it is working with a few test cases for dihedrals.  However, we need to build improper test cases. This will be done for mosdef-gomc in the next few weeks.  Therefore, We should approve it to merge into the dev branch for not so other can use this to get other Prs started.

gmso/external/convert_parmed.py
```diff
@@ -426,6 +462,65 @@ def _dihedral_types_from_pmd(structure, dihedral_types_member_map=None):
...
+
+    for impropertype in structure.impropers:
+        improper_params = {
+            "xi_k": (impropertype.psi_k * u.Unit("kcal/mol")),
+            "xi_eq": (impropertype.psi_eq * u.Unit("kcal/mol")),
+        }
+
+        member_types = improper_types_member_map.get(id(impropertype))
+
+        top_impropertype = gmso.ImproperType(
```
umesh-timalsina, line 514: One quick thought or question/thought here. Can we use the same/uniform expression throughout the conversion?

Here's what we have defined for a standard HarmonicImproperPotential to look like.
https://github.com/mosdef-hub/gmso/blob/3a387c707109e64b5f178591249db94a134d2ac4/gmso/lib/jsons/HarmonicImproperPotential.json#L1-L5

gmso/external/convert_parmed.py
```diff
@@ -158,41 +170,65 @@ def from_parmed(structure, refer_type=True):
...
-        # Generate dihedral parameters for DihedralType that gets passed
-        # to Dihedral
+        # Generate parameters for ImproperType or DihedralType that gets passed
+        # to corresponding Dihedral or Improper
         # These all follow periodic torsions functions
-        # (even if they are improper dihedrals)
         # Which are the default expression in top.DihedralType
         # These periodic torsion dihedrals get stored in top.dihedrals
+        # and periodic torsion impropers get stored in top.impropers
         if dihedral.improper:
```
umesh-timalsina, line 179: The parmed documentation suggests that atom3 is the central atom if this flag is set for a dihedral. Have you found that to be the case?

https://parmed.github.io/ParmEd/html/topobj/parmed.topologyobjects.Dihedral.html#parmed.topologyobjects.Dihedral

gmso/external/convert_parmed.py
```diff
@@ -158,41 +170,65 @@ def from_parmed(structure, refer_type=True):
...
             )
-        # No bond parameters, make Connection with no connection_type
+            if refer_type and isinstance(dihedral.type, pmd.DihedralType):
+                top_connection = gmso.Improper(
+                    connection_members=[
+                        site_map[dihedral.atom1],
+                        site_map[dihedral.atom2],
+                        site_map[dihedral.atom3],
+                        site_map[dihedral.atom4],
+                    ],
```
umesh-timalsina, line 193: See the above comment.

gmso/external/convert_parmed.py
daico007: I spot checked the top file of `124-trimethylbenzene` in Foyer OPLSAA test and its impropers have the order of `[Central, 1, 2, 3]`, and this ordered is the same when reading in using parmed (just `parmed.load_file()`). Found a comment (https://github.com/ParmEd/ParmEd/blob/b9f2955f792c03394686824519cabe030d5342b4/parmed/gromacs/gromacstop.py#L1557-L1568) so this sounds like a known bug for them and it has not been fixed. The order of atoms you found is probably what it should be if things flow downstream (you create a system and save it out to anything but a top format). I think this means the order of atom in their dihedral when `dihedral.improper == True` is just not reliable and we might need to test for when creating our own improper object (either here or in a different PR, also checking for the validity of other connection object like angle and proper dihedral). Here, I think we should make a note here about the parmed issue so we can come back to it later.

gmso/external/convert_parmed.py
CalCraven: After chatting with @bc118, I think this makes sense instead of creating a new custom potential.

umesh-timalsina:
After the tests pass, LGTM!!.

PR #661, opened 2022-06-01
Expanded search for elements when reading mol2

umesh-timalsina:
Currently, when loading in a mol2 file, we only consider column5 for the element guessing. This PR attempts to guess the element by looking at both column5 and column1 of the mol2 file.


daico007 (approved):
LGTM!

PR #669, opened 2022-06-14
Modify scaling factors to a numpy array

### LINKED ISSUE #668: Allow multiple set of scaling factors and mixing rule
Per discussion with @daico007, @bc118, @umesh-timalsina, and @CalCraven, it appears there might be occasions where we may want to allow multiple sets of scaling factors and mixing rules (attached to molecule level).  For example, AMBER FFs have different 1-4 interactions scaling between them and are allowed to be simulated together.  We need to allow these different 1-4 scaling factors (1-2, 1-3, 1-4,...) on the GMSO side, and let the engine or file writers filter out what is acceptable at that level.  The best way to access the data may be something like this molecule_name -> FF.xml -> dict/list of scaling factors.

We probably still want to store this information at the Topology level (or molecule level). We should be able to find the 1-4 interactions; however, storing this information at the atom level is error-prone.  For example, errors arise when using >1 FF if the atom_type is the same between 2 FFs and the scaling factor is different (how do we sort this mess unless renaming the atomType like atomType_molecule name to ensure it is unique to that molecule?). There are likely other reasons, but this is one of them.  

For the mixing rules, these should also be stored at the Topology level (or molecule level).  While using different mixing rules is technically wrong, some people are doing it for various systems/reasons.  Therefore, to make GMSO support more supportive, we should consider adding an "expert_mode" override (to be added); however, I would likely still throw a warning that this is happening because it is technically wrong. This should be done the same as the 1-4 scaling factors: molecule_name -> FF.xml -> dict/list of mixing rule names, allowing the engine/file writers to handle the issues however the user programs it.

umesh-timalsina:
Closes #668. This PR converts the scaling factors ina topology to a numpy array. Currently, this is an 2 by 3 numpy array with `12`, `13` and `14` interactions defined as follows:
```py
[
   [lj12, lj13, lj14,],
   [electrostatics12, electrostatics13, electrostatics14,]
]
```

To that end the following methods are added for making working the `scaling_factors` easier.

```py
>>> from gmso.core.topology import Topology
>>> top = Topology()
>>> top.scaling_factors
{'Topology': array([[0. , 0. , 0.5],
       [0. , 0. , 0.5]])}]
top.lj_scale(interaction='14'), top.lj_scale(interaction='12'), top.lj_scale(interaction='13')
(0.5, 0.0, 0.0)
>>> top.lj_scale(interaction='16')
  File "/home/umesh/krow/mosdef/gmso/gmso/core/topology.py", line 619, in lj_scale
    return self._get_scaling_factor(molecule_id, interaction, "lj_scale", 0)
  File "/home/umesh/krow/mosdef/gmso/gmso/core/topology.py", line 655, in _get_scaling_factor
    raise GMSOError(f"Unknown {name} interaction {interaction}")
gmso.exceptions.GMSOError: Unknown lj_scale interaction 16
>>> top.set_lj_scale([0.2], interaction='14')
>>> top.lj_scale()
array([0. , 0. , 0.2])
>>> top.set_lj_scale([0.1, 0.2, 0.4])
>>> top.lj_scale()
array([0.1, 0.2, 0.4])
>>> top.set_lj_scale([0.1, 0.2, 0.3], molecule_id='my_subtopology_name')
>>> top.lj_scale(molecule_id='my_subtopology_name', interaction='14')
0.3

```

Same applies for `electrostatics_scale` as well.

### DIFF
--- a/gmso/parameterization/topology_parameterizer.py
+++ b/gmso/parameterization/topology_parameterizer.py
@@ -181,40 +181,57 @@ class TopologyParameterizer(GMSOBase):
             subtop_or_top, forcefield, is_subtop=is_subtop
         )
 
-    def _verify_forcefields_metadata(self):
-        """Verify all the provided forcefields have the same scaling factors and combining rule."""
+    def _set_combining_rule(self):
+        """Verify all the provided forcefields have the same combining rule and set it for the Topology."""
         if isinstance(self.forcefields, dict):
-            ffs = list(self.forcefields.values())
-            init_scaling_factors = ffs[0].scaling_factors
-            init_combining_rule = ffs[0].combining_rule
-            for ff in ffs[1:]:
-                if ff.scaling_factors != init_scaling_factors:
-                    raise ParameterizationError(
-                        "Scaling factors of the provided forcefields do not"
-                        "match, please provide forcefields with same scaling"
-                        "factors that apply to a Topology"
-                    )
-
-                if ff.combining_rule != init_combining_rule:
-                    raise ParameterizationError(
-                        "Combining rules of the provided forcefields do not"
-                        "match, please provide forcefields with same scaling"
-                        "factors that apply to a Topology"
-                    )
-            return init_scaling_factors, init_combining_rule
+            all_comb_rules = set(
+                ff.combining_rule for ff in self.forcefields.values()
+            )
         else:
-            return (
-                self.forcefields.scaling_factors,
-                self.forcefields.combining_rule,
+            all_comb_rules = {self.forcefields.combining_rule}
+
+        if not len(all_comb_rules) == 1:
+            raise ParameterizationError(
+                "Combining rules of the provided forcefields do not"
+                "match, please provide forcefields with same scaling"
+                "factors that apply to a Topology"
+            )
+        self.topology.combining_rule = all_comb_rules.pop()
+
+    def _set_scaling_factors(self):
+        """Set either per-molecule or global scaling factors for the topology based on the forcefields provided."""
+        # ToDo: Set other scaling factors by extending the forcefield schema
+        # ToDo: What to do when all the scaling factors matchup? Should we promote them to be global?
+        if isinstance(self.forcefields, Dict):
+            for subtop_id, ff in self.forcefields.items():
+                self.topology.set_lj_scale(
+                    ff.scaling_factors["nonBonded14Scale"],
+                    interaction="14",
+                    molecule_id=subtop_id,
+                )
+                self.topology.set_electrostatics_scale(
+                    ff.scaling_factors["electrostatics14Scale"],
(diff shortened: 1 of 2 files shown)


gmso/tests/test_topology.py
```diff
@@ -630,38 +631,52 @@ def test_topology_get_dihedrals_for(self, typed_methylnitroaniline):
...
+        )
+        typed_methylnitroaniline.set_lj_scale(1.0, interaction="12")
+        assert np.allclose(typed_methylnitroaniline.lj_scale(), [1.0, 0.5, 1.0])
 
     def test_topology_invalid_scaling_factors(self, typed_methylnitroaniline):
         with pytest.raises(GMSOError):
-            typed_methylnitroaniline.scaling_factors = (0.5, 1.0)
-        with pytest.raises(GMSOError):
-            typed_methylnitroaniline.scaling_factors = {"lj_12": 0.0}
+            typed_methylnitroaniline.lj_scale(interaction="16")
```
CalCraven, line 680:
```suggestion
        with pytest.raises(GMSOError):
            typed_methylnitroaniline.get_scaling_factors(molecule_id="Missing Name")
            
```

gmso/tests/test_topology.py
```diff
@@ -630,38 +631,52 @@ def test_topology_get_dihedrals_for(self, typed_methylnitroaniline):
...
+            typed_methylnitroaniline.electrostatics_scale(interaction="12"), 1.0
+        )
+        assert np.allclose(
+            typed_methylnitroaniline.electrostatics_scale(interaction="13"), 1.0
+        )
+        assert np.allclose(
+            typed_methylnitroaniline.electrostatics_scale(interaction="14"), 1.0
+        )
+        typed_methylnitroaniline.set_lj_scale(1.0, interaction="12")
+        assert np.allclose(typed_methylnitroaniline.lj_scale(), [1.0, 0.5, 1.0])
```
CalCraven, line 676:
```suggestion
        typed_methylnitroaniline.set_lj_scale(1.0, interaction="12", molecule_id="Molecule")
        assert np.allclose(typed_methylnitroaniline.lj_scale(molecule_id="Molecule"), [1.0, 0.0, 0.0])

```

CalCraven (approved):
Additional tests. Do we want to change it to `get_lj_scale`, instead of just `lj_scale`? Otherwise looks good to me!

daico007 (approved):
I think this looks good. I don't know why but it seems like this branched is being compared to the `main` branch? (or one that doesn't have the new gmso.parameterization` module). Will merge this in tmr morning and consolidate with #638 so Brad can test everything. After we might need to see if we can reduce the number of method relating to set/get scaling factors (right now there are 10, including helper methods and and @property, which seems a bit too many, I am worrying that users may be confused which one to used)

PR #676, opened 2022-07-21
pin unyt to version that supports 3.7

daico007:
It seems like unyt decided to drop support for python 3.6 and 3.7 (https://pypi.org/project/unyt/), so this PR will pin the unyt package to those that still support py3.7. We will need to make a decision soon about whether it's time to drop support for py3.7 too.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - numpy
   - sympy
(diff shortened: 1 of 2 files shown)


daico007:
didn't realize test on main doesn't fail with last push, so gonna hold this off til any funky thing happens.

Issue #678, opened 2022-07-22
`write_gsd` doesn't work when no mass is set


chrisjonesBSU:
`write_gsd` doesn't work if masses aren't set in the topology. 

```
129 masses = np.array([site.mass for site in top.sites])
    130 masses[masses == 0] = 1.0
--> 131 gsd_snapshot.particles.mass = masses / ref_mass
```

It looks like it tries to correct for this case (line 130), but the problem is that `top.site.mass` defaults to `None` if it hasn't been set.

This is also related to #664 

I think a couple fixes are:

1. Set default `site.mass` to 0 instead of None
2. Add another line `masses[masses == None] = 1.0` to correct for None masses
3. Go through our external `from_*` functions and try to populate `site.mass` correctly when the information is available
4. Check for `None` masses and throw an error message instructing the user to populate the site masses

chrisjonesBSU:
It also fails due to `site.charge` defaulting to `None` see #679

CalCraven:
Makes sense to me. I think the preferred behavior would be to check for `mass == None` values, since the `None` input gives us a bit more information than the 0. A `None` just means no mass has been associated with the site, and that allows some of the different automated ways to grab the mass to kick in, such as from the `atomtype` or from`ele` interpreting the element from the name. Therefore I think leaving None as the default is good. Honestly, auto setting it to 1 seems like bad practice because you could miss that the masses are not saved and then they get overwritten with that 1 value. 

I almost think this issue necessitates some hierarchy of mass treatment. We could make a module which gives the many ways to check the mass for a site, and we can specify a given order for the specific writer instance. 
1. from site.atomtype.mass
2. from site.mass
3. from ele.element_from_name(site.name)
4. from site_max_sigma.sigma / site.sigma * site_max_sigma.mass (the ratio of the current site's sigma to the max sigma in the system)
5. Raise Warning that can be turned off in expert mode

chrisjonesBSU:
I agree. I'll just add an extra line that corrects for `None` masses and charges.

CalCraven:
I think that's the easiest fix. Where are you setting the mass from in this instance?

chrisjonesBSU:
I came across this issue going form mbuild --> gmso --> gsd file. Ultimatley, using it as a replacement for the gsd writer in mBuild (which uses Parmed) and doesn't populate angles and dihedrals.

daico007:
[...]

Issue #663, opened 2022-06-08
Module gmso.core.unit_systems


umesh-timalsina:
It might be nice to define a default unyt system for GMSO such that there's uniformity in files written in and out of GMSO. This offers several significant advantages:
1. Uniformity in different files written in and out of GMSO, even the possibility of stripping the unyt storage.
2. At each writer/external package level, we will have the ability to define different unit systems as was mentioned in #656. 
3. Missing Units from XML or JSON can be handled by default capabilities.

daico007:
https://github.com/mosdef-hub/gmso/issues/315

PR #680, opened 2022-07-22
Fix mass and charge in GSD writer; Add angles and dihedrals to GSD writer

chrisjonesBSU:
This helps address #678 where the gsd writer was trying to perform math operations with `NoneType` objects. 

Still need to add a unit test

Edit: Some other errors popped up, and I ended up totally revamping the gsd writer.  It is writing out angles and dihedrals now.

Edit: I edited the title again to better reflect the changes by the PR. After talking with Cal in #678 there aren't any changes to default mass and charge values.

### DIFF
--- a/gmso/formats/gsd.py
+++ b/gmso/formats/gsd.py
@@ -85,21 +85,21 @@ def write_gsd(
     gsd_snapshot.configuration.box = np.array([lx, ly, lz, xy, xz, yz])
 
     warnings.warn(
-        "Only writing particle, bond, and angle information."
-        " Dihedrals are not currently written to GSD files",
+        "Only writing particle, bond, angle, and dihedral information."
+        " Special pairs are not currently written to GSD files",
         NotYetImplementedWarning,
     )
     _write_particle_information(
         gsd_snapshot, top, xyz, ref_distance, ref_mass, ref_energy, rigid_bodies
     )
-    # if write_special_pairs:
-    #    _write_pair_information(gsd_snapshot, top)
     if top.n_bonds > 0:
         _write_bond_information(gsd_snapshot, top)
     if top.n_angles > 0:
         _write_angle_information(gsd_snapshot, top)
     if top.n_dihedrals > 0:
         _write_dihedral_information(gsd_snapshot, top)
+    # if write_special_pairs:
+    #    _write_pair_information(gsd_snapshot, top)
 
     with gsd.hoomd.open(filename, mode="wb") as gsd_file:
         gsd_file.append(gsd_snapshot)
@@ -272,10 +272,8 @@ def _write_dihedral_information(gsd_snapshot, top):
     ----------
     gsd_snapshot :
         The file object of the GSD file being written
-
-    Warnings
-    --------
-    Not yet implemented for gmso.core.topology objects
+    top : gmso.Topology
+        Topology object holding system information
 
     """
     gsd_snapshot.dihedrals.N = top.n_dihedrals
@@ -296,7 +294,11 @@ def _write_dihedral_information(gsd_snapshot, top):
             _t2 = t2.name
             _t3 = t3.name
 
-        dihedral_type = ('-'.join((_t1, _t2, _t3, _t4)))
+        if [_t2, _t3] == sorted([_t2, _t3], key=natural_sort):
+            dihedral_type = ('-'.join((_t1, _t2, _t3, _t4)))
+        else:
+            dihedral_type = ('-'.join((_t4, _t3, _t2, _t1)))
+
         unique_dihedral_types.add(dihedral_type)
         dihedral_typeids.append(
                 list(unique_dihedral_types).index(dihedral_type)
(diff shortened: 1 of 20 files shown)


chrisjonesBSU:
This also contains another fix where the gsd writer was trying to call on `t.atom_type` and `t.name` when `t1` and `t2` were already the types or names.

chrisjonesBSU:
So, it looks like f-string formatting doesn't work when using it with `warnings.warn`?

daico007:
hmm, that should not be the case, I used f-string with `warnings.warn` in multiple occasions before. What's the error message are you getting?

chrisjonesBSU:
Before we merge this, let me test it out with some more compounds from mBuild and make sure the information the gsd file is correct and makes sense.

chrisjonesBSU:
The workflow in the latest unit test I added also solves https://github.com/mosdef-hub/mbuild/issues/1025

gmso/formats/gsd.py
```diff
@@ -121,16 +121,18 @@ def _write_particle_information(
...
     gsd_snapshot.particles.types = unique_types
-    warnings.warn("{} unique particle types detected".format(len(unique_types)))
+    warnings.warn(f"{len(unique_types)} unique particle types detected")
 
     typeids = np.array([unique_types.index(t) for t in types])
     gsd_snapshot.particles.typeid = typeids
 
     masses = np.array([site.mass for site in top.sites])
     masses[masses == 0] = 1.0
+    masses[masses == None] = 1.0
```
daico007, line 131: Do you think we should either raise a warning here (that there is mass of 0/None which would be set to 1) or even raise and Error (that particle mass has not been initiated)?

daico007: that's a good idea! I think we could have a method for Topology, just to scan through and make sure all necessary properties have been initiated, all objects have been parameterized, etc. This will not need to be explicitly called by the user.

daico007: I guess we can have the warning/error here temporarily before the validating method is added to the main gmso.Topology

umesh-timalsina: Putting my two cents here, but I think we could do this with a custom unyt system per writer. Since masses are by default assumed to be in `gram/mol` in gmso. I don't like the conversion to numpy array here because we don't enforce any constraints on the units of a site mass.


```py

>>> from gmso import Atom, Topology
>>> import unyt as u
>>> import numpy as np
>>> top = Topology(name='mass demonstration')
>>> atom1 = Atom(mass=2.3 * u.gram / u.mol)
>>> atom2 = Atom(mass=2.4 * u.mg / u.mol)
>>> atom2.mass
unyt_quantity(2.4, 'mg/mol')
>>> atom1.mass
unyt_quantity(2.3, 'g/mol')
>>> top.add_site(atom1)
>>> top.add_site(atom2)
>>> top.update_topology()
>>> top
<Topology mass demonstration, 2 sites,
 0 connections,
 0 potentials,
 id: 140549244341696>
>>> mass_array = np.array([site.mass for site in top.sites])
>>> mass_array
array([2.3, 2.4])
```

Clearly, this is not correct. I think a common validation/filtration/customization  layer from between `gmso.core` and `gmso.formats` that can be leveraged by the writers can be developed (starting with the gsd writer). The idea is to define useful unit systems for hoomd/gsd at the writer level, however have these common utilities either in the validation/filteration layer or the gmso topology itself.

umesh-timalsina: Currently, I have been thinking of these writer/conversion, etc... as a series of dictionary transformations with differing levels of hierarchy per package/simulation engine.

umesh-timalsina: See https://github.com/mosdef-hub/gmso/issues/663 for a discussion on custom unyt systems

CalCraven: These are good points @umesh-timalsina. Can I suggest that in this unyt validator, we can iterate through all properties with unyts attached to them and make sure they are equivalent? This would include the unyts in multiple different potential expressions attached to different typed objects across the topology. What we can do is raise an error if there are any differences, and then suggest a method available through the topology which converts all parameters following one of our available sets of base units. These same libraries than can be leveraged by particular writers for their necessary output styles.

umesh-timalsina: This is already the case with the potential expressions' expected dimensions in our potential template library. We had this idea of dropping unit object internally. However, this can be fleshed out further. This will take a substantial refactor of the existing codebase (internal part of the core API) however, what we need is a good set of default units to carry and some effort on the plumbing to make it seamless to the end user.

chrisjonesBSU: >  The idea is to define useful unit systems for hoomd/gsd at the writer level

I think we should avoid defining writer specific unit systems. I think that adds unnecessary complications and potentially confusion for the end user.  For example, hoomd doesn't have a standard set of [units](https://hoomd-blue.readthedocs.io/en/v3.3.0/units.html). So in that case, if the user is using g/mol or mg/mol, or angstroms or nm in their topology, it doesn't matter. As long as the defined units are consistent; which I believe is what Cal is getting at.

gmso/formats/gsd.py
```diff
@@ -85,21 +85,21 @@ def write_gsd(
     gsd_snapshot.configuration.box = np.array([lx, ly, lz, xy, xz, yz])
 
     warnings.warn(
-        "Only writing particle and bond information."
-        " Angle and dihedral is not currently written to GSD files",
+        "Only writing particle, bond, angle, and dihedral information."
+        " Special pairs are not currently written to GSD files",
```
daico007, line 89:
```suggestion
        "Impropers and special pairs are not currently written to GSD files",
```

daico007 (approved):
Slight update to docstring, the rest LGTM!

CalCraven (approved):
LGTM! Would like some testing for more ways to write out GSD files and testing them, but until we have a more extensive set of things we want to be sure to support, I think we're good.

PR #684, opened 2022-08-02
Populate both member types and classes

daico007:
Populate both `member_types` or `member_classes` of a connection potential if that field is empty in the forcefield. Also reorder the `connection.connection_members` based on the order of the matched connection potential


daico007:
Have to revert the reordering of `connection.connection_members` since that prevent us from doing `assert_same_connection_params` check (for oplsaa tests specifically)

bc118:
I ran all mosdef-gomc tests 50 times, and 100 times with the improper being order specific.  It all passes, soI would say this is working properly!

bc118 (approved):
I ran all mosdef-gomc tests 50 times, and 100 times with the improper being order specific.  It all passes, soI would say this is working properly!

PR #693, opened 2022-09-07
Drop py37

daico007:
To coordinate with foyer, this PR drop support for py37. Also fix a bug coming from a arg name change in foyer (i.e., this PR should only pass the bleeding test and none of the other). This PR can wait until release foyer get released.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -68,6 +68,11 @@ jobs:
           extra-specs: |
             python=3.8
 
+      - name: Clone mbuild and foyer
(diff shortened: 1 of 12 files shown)


CalCraven (approved):
Looks like there are still some failing tests in test_impropers_parameterization.py due to the symbol vs element changes done here to align with Foyer. So either we need to wait until a foyer release to merge this, or pin the testing to the develop version of Foyer.

daico007:
Bleeding test passed and everything else (as expected), so I am gonna merge this PR

PR #696, opened 2022-10-05
Add HOOMD Supports

daico007:
[WIP] Start adding supports for HOOMD. Goal is to convert a typed GMSO Topology to a HOOMD state (gsd snapshot) and all the required forces.


gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
...
+            "parser": _parse_lj,
+        },
+        "BuckinghamPotential": {
+            "container": hoomd.md.pair.Buckingham,
+            "parser": _parse_buckingham,
+        },
+        "MiePotential": {"container": hoomd.md.pair.Mie, "parser": _parse_mie},
+    }
+
+    nlist = hoomd.md.nlist.Cell(exclusions=["bond", "1-3"], buffer=nlist_buffer)
```
chrisjonesBSU, line 566: Do we want to include the option to add "1-4" to the nlist exclusions? IMO that would be useful.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
...
+    )
+    groups = dict()
+    for agtype in unique_agtypes:
+        group = potential_types[agtype]
+        if group not in groups:
+            groups[group] = [agtype]
+        else:
+            groups[group].append(agtype)
+
+    base_units["angle"] = u.degree
```
chrisjonesBSU, line 651: Does this mean the equilibrium angles will be in degrees? Hoomd uses radians for angles and dihedrals.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
...
+        expected_units_dim = potential_refs[group][
+            "expected_parameters_dimensions"
+        ]
+        groups[group] = _convert_params_units(
+            groups[group], expected_units_dim, base_units
+        )
+
+    dtype_group_map = {
+        "PeriodicTorsionPotential": {
+            "container": hoomd.md.dihedral.Harmonic,  # Should this be periodic, ask Josh
```
chrisjonesBSU, line 743: As of Hoomd 3.8, `hoomd.md.dihedral.Harmonic` is deprecated and I think we should use `hoomd.md.dihedral.Periodic`  However, `hoomd.md.dihedral.Periodic` isn't available in versions < 3.8.  Do we need a hoomd version check somewhere?

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
...
+        expected_units_dim = potential_refs[group][
+            "expected_parameters_dimensions"
+        ]
+        groups[group] = _convert_params_units(
+            groups[group], expected_units_dim, base_units
+        )
+
+    itype_group_map = {
+        "HarmonicImproperPotenial": {
+            "container": hoomd.md.dihedral.Harmonic,  # Should this be periodic, ask Josh
```
chrisjonesBSU, line 800: See the previous comment about `hoomd.md.dihedral.Periodic`

chrisjonesBSU (changes requested):
I commented a couple questions, and some things to watch out for.  I haven't tried using this yet, I'll be sure to do that ASAP.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
...
+            agtype_group_map[group]["parser"](
+                container=agtype_group_map[group]["container"](),
+                agtypes=groups[group],
+            )
+        )
+
+    return angle_forces
+
+
+def _parse_dihedral_forces(top, potential_types, potential_refs, base_units):
```
chrisjonesBSU, line 678: In this funciton, do we need to make sure any values for angles are in the correct units, similar to you're doing in the `_parse_angles` function?

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,881 @@
+"""Convert GMSO Topology to GSD snapshot."""
+from __future__ import division
+
+import itertools
+import json
+import re
+import statistics
+import warnings
+
+import hoomd
```
chrisjonesBSU, line 10: We either should add hoomd to the environment file for gmso, or raise an exception when the user is trying to use `convert_hoomd.py` and does not have hoomd in their environment.

gmso/external/convert_hoomd.py
chrisjonesBSU: Looks like we are getting the angles in degrees.

Here is what I get for a methane molecule using oplsaa

```
opls_140-opls_138-opls_140
_HOOMDDict{'t0': 107.80000000000646, 'k': 0.08411827276958087}
```

Hoomd is agnostic to units for distance and energy, but angles should definitely be in radians.

chrisjonesBSU:
I tested this out in a notebook. It seems to be working well for methane. See my comments above about making sure we use radians.

I also tried it out with ethane so we have dihedrals, and the performance suffers quite a bit.


```
import mbuild as mb
import gmso
import forcefield_utilities as ffutils
from gmso.external.convert_hoomd import to_hoomd_forcefield 
from gmso.external.convert_mbuild import from_mbuild

mol = mb.load("CC", smiles=True)
box = mb.packing.fill_box(compound=mol, n_compounds=40, box=[2,2,2])
oplsaa = ffutils.FoyerFFs().load('oplsaa').to_gmso_ff()
gmso_box = from_mbuild(box)
apply(top=gmso_box, forcefields=oplsaa, identify_connections=True)

forcefield = to_hoomd_forcefield(top=gmso_box)
```

When I use 40 methane molecules the `to_hoomd_forcefield` call takes about 200ms to run

When I use 20 ethane molecules the `to_hoomd_forcefield` call takes takes over 4 minutes to run

The slow down must be somewhere in parsing dihedrals, since methane doesn't contain any.

daico007:
Good catch! I also just realized the performance issue last night but haven't pinpoint where the bottleneck is. From your comment, I will start with the dihedral type parsing.

gmso/external/convert_hoomd.py
daico007: Now included (dynamically added from top.scaling_factors)

gmso/external/convert_hoomd.py
daico007: Good call! I will fix it in the next commit.

gmso/external/convert_hoomd.py
daico007: Note, I think we will only try to support HOOMD 3 for GMSO (I will have a version check somewhere)

chrisjonesBSU:
The latest changes definitely improves the performance.  Using the same example above, 40 ethane molecules takes about 600ms while the 40 methanes still take about 200ms.

gmso/external/convert_hoomd.py
daico007: Added the has_hoomd check

chrisjonesBSU:
I think we should implement `auto_scale` in the same way it's done in `create_hoomd_forcefield` in mBuild.  Has there been any discussion about this that I missed, or is this something that is potentially handled elsewhere in GMSO?

chrisjonesBSU:
> I think we should implement `auto_scale` in the same way it's done in `create_hoomd_forcefield` in mBuild. Has there been any discussion about this that I missed, or is this something that is potentially handled elsewhere in GMSO?

Also, I can plan on contributing this part of the PR if it is something we want in `to_hoomd_forcefield`

daico007:
@chrisjonesBSU please do! I don't think we make any decision about the `autoscale` stuff and I just forgot about that altogether

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1498 @@
...
+        all reference scaling values is set to 1.
+        A dictionary specifying the referenced scaling values may also be provided for this argument.
+
+    Returns
+    -------
+    forces : dict
+        Each entry
+    """
+    potential_types = _validate_compatibility(top)
+    base_units = _validate_base_units(
```
chrisjonesBSU, line 607: It might be helpful to return base_units as well, similar to what we do in `create_hoomd_forcefield`

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1498 @@
...
+            if site.mass
+            else 1 * base_units["mass"]
+        )
+        charges.append(site.charge if site.charge else 0 * u.elementary_charge)
+
+    """
+    Permittivity of free space = 2.39725e-4 e^2/((kcal/mol)(angstrom)),
+    where e is the elementary charge
+    """
+    # e0 = u.physical_constants.eps_0.in_units(
```
chrisjonesBSU, line 285: More commented out lines that can probably be removed.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1498 @@
...
+
+
+def _parse_particle_information(
+    snapshot,
+    top,
+    base_units,
+    rigid_bodies,
+    shift_coords,
+    box_lengths,
+    # ref_values,
```
chrisjonesBSU, line 236: Can probably remove this, and update doc strings

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1498 @@
...
+        A dictionary specifying the referenced scaling values may also be provided for this argument.
+
+    Notes
+    -----
+    Force field parameters are not written to the GSD file and must be included
+    manually in a HOOMD input script. Work on a HOOMD plugin is underway to
+    read force field parameters from a Foyer XML file.
+    """
+    base_units = _validate_base_units(base_units, top, auto_scale)
+    # ref_values = _infer_ref_values(top, auto_scale, base_units)
```
chrisjonesBSU, line 185: Remove this line?

chrisjonesBSU (commented):
Looks good. Auto scale seems to be working as expected. What do you think about returning the `base_units` dict along with the forcefield? This is basically what we do in mbuild's `create_hoomd_forcefield`.

Also, looks like there are a few commented out lines from when you were troubleshooting things. 

I think this is ready after we decide whether or not to include `base_units` in the return statement.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1498 @@
...
+        fudged 1,4 interactions in HOOMD.
+
+    Notes
+    -----
+    Force field parameters are not written to the GSD file and must be included
+    manually in a HOOMD input script. Work on a HOOMD plugin is underway to
+    read force field parameters from a Foyer XML file.
+    """
+    base_units = _validate_base_units(base_units, top, auto_scale)
+    # ref_values = _infer_ref_values(top, auto_scale, base_units)
```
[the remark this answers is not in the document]
daico007: good catch, I will remove all these lines

gmso/external/convert_hoomd.py
daico007: `base_units` is now returned along with the forces/snapshot

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1439 @@
...
+from gmso.utils.geometry import coord_shift
+from gmso.utils.io import has_gsd, has_hoomd
+from gmso.utils.sorting import (
+    natural_sort,
+    sort_connection_members,
+    sort_member_types,
+)
+
+if has_gsd:
+    import gsd
```
chrisjonesBSU, line 30: I think this should be `import gsd.hoomd`  I'm getting an attribute error if we only import gsd then try calling `gsd.hoomd`.

chrisjonesBSU (commented):
One note to fix the import, I'll push this change real quick, just wanted to document it.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1439 @@
...
+        xy = u_vectors[1][0]
+        xz = u_vectors[2][0]
+        yz = u_vectors[2][1]
+    return lx, ly, lz, xy, xz, yz
+
+
+def to_hoomd_forcefield(
+    top,
+    nlist_buffer=0.4,
+    r_cut=0,
```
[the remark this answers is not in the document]
daico007: That's a good idea. I can make it so user have to explicitly enter r_cut.

chrisjonesBSU (approved):
LGTM! It seems to be working great.  Thanks for working on this @daico007.

daico007:
Cool, thank you for reviewing @chrisjonesBSU! I will let this sit a bit in case anyone want to drop a comment/review. I will merge this PR early next week.

gmso/external/convert_hoomd.py
```diff
@@ -0,0 +1,1441 @@
...
+                )
+            else:
+                msg = "{key} is in wrong unit dimension"
+                if isinstance(base_units[key], u.Unit):
+                    assert base_units[key].dimensions == ref[key], msg
+                elif isinstance(base_units[key], u.Unit):
+                    assert base_units[key].units.dimensions == ref[key], msg
+                else:
+                    raise TypeError(
+                        f"Base unit of {key} must be of type u.Unit or u.unyt_quantity."
```
chrisjonesBSU, line 1349: @daico007 Quick quesiton, if I try using `unyt.amu` for the mass base units, this exception is being raised.

I'm not very familiar with Unyt, but for some reason `unyt.amu` has a different type than `unyt.g` (is there a reason for this that you know of?)

`unyt.amu` is of type `unyt.array.unyt_quantity`. For comparison, `unyt.g`  and `unyt.g/unyt.mol` are type `unyt.unit_object.Unit`

Do we want to allow someone to use a mass unit of `amu` here. I know it's the same as as `g/mol`, which works, but some people might default to using `amu`; then get confused about this error message.

daico007: Good point! I didn't consider the amu when doing the parser. I will at a special treatment for `u.amu`.

daico007: I think I found the bug for this, working on a fix now.

PR #700, opened 2022-11-22
Speed up check_compatibility for relevant writer method

daico007:
With the new change to Topology's potential types, the performance of the writing step may be affected by the `check_compatiblity` step (due to a lot of `sympy.sympify` call). A `simplify_check` option has been added to speed up the process (started with GROMACS top writer). This PR will attempt to add the `simplify_check` option to other relevant writer and add documentation at the `Topology.save` method.


daico007:
On second, I think the simplify check behavior should be the default behavior for the compatibility check step. I will try to modify the method such that I will speed up in most case and only slow down/perform detail check when encounter hard case.

CalCraven:
@daico007 I've pretty much dismantled the lammpswriter in the PR I'm currently working on. So maybe we should limit this to just the speedups checks instead of also including these changes directly to the lammpswriter. Let me pull the PR with what I have currently and maybe we can discuss what needs to be included between them. I mostly followed the way it was done with the `.top` writer.

CalCraven:
Briefly comparing the two, it looks like I didn't actually make the connection maps you generate in this code addition, but otherwise I've covered the rest of these changes. We can discuss the best way to iterate through the connection types in that PR.

PR #703, opened 2022-12-06
turn off fail-fast & update readme

daico007:
Right now, if what of the check failed, all the other unfinished checks will get canceled, which miss the point of testing on different OS. This PR turns that off. Similar to https://github.com/mosdef-hub/mbuild/pull/1067


daico007:
Will also try to update the readme to reflect recent changes

Issue #709, opened 2023-01-24
Gromacs writer truncates residue and atom type names too short.


chrisiacovella:
In the current code, the residue name and atom type are both truncated to 3 characters.  5 characters are supported for both these fields.  Truncation of atom type is particularly problematic as 4 character names are typical in charmm and also may be needed for various OPLS parameters (even chopping off the"opls_" from the atom type name, there are a handful of parameters, including those for silica and alkylsilanes we have added, that have numbers 1000 and greater in our force field fields e.g. opls_1002 ). 

This should also include checks to throw an error if the residue number exceeds 99999 (as it is limited to 5 positions as well).

https://manual.gromacs.org/documentation/current/reference-manual/file-formats.html#gro

chrisiacovella:
Related: If I use the "to_gmso" function to go from an mbuild compound that contains copies of multiple different types of molecules (e.g., 5 FFA molecules and 5 CER molecules), the first 5 FFA molecules will be residues 1-5, then the numbers will start again once we encounter the CERs, as this info is inferred from site. molecule information. 

Gromacs doesn't actually us the number in the file to assign the internal residue number, but it uses changes in residue number in the file to trigger an update to the current residue number.   

So in the case described above, this will be fine, because as song as we start reading in the first CER molecule, it will see that the new res_num (1)  != old_res_num ( 5).  If the file were initialized, as FFA1, CER1, FFA2, CER2, , FFA3, CER3, etc... this would not treat FFA and CER differently given the same molecule number. 

While this could be addressed in the write, change the conversion tool (where molecule number is set) would be better, since if you loop over a topology, site.molecule.number indexing should be unique to the topology, not unique to a molecule type in the topology.

chrisiacovella:
Summary of discussions with @daico007 and @CalCraven ; there likely is utility to having the site.molecule.number be unique to each molecule type, rather than between types.  We could easily add in additional quantities that provides a unique molecule number.  Even with a unique molecule number, logic needs to be a bit different in .gro files because of the support for residues.

PR #710, opened 2023-01-25
Gromacs .gro writer

chrisiacovella:
This relates to issue #709 .  Initially, residue name and atom_type name were truncated to be 3 characters, when the file format itself allows for 5 (these longer representations are needed for forcefields like charmm).   This has been changed.

Additionally, this adds a sorting function to ensure unique a unique residue id is set for each molecule and /or residue in the system.  These residue ids for each site are set in the label field (non destructively, anything in the field is added on), to allow us to see the association between a given site/molecule and the renumbered output file (I imagine primarily useful for testing purposes).   

Additionally, resid and atom id numbers are reset to 1 if/when they exceed 5 digits,(necessary since this is a fixed width format). 

While the code is in place and should work for when residues are explicitly defined (rather than just using the residue id as a convenient way to tag entire molecules), this is untested at the moment.   New tests still needed to be added to testing this writer.


CalCraven:
Could be useful to add a test that would generate a structure with an odd ordering of the molecules in the system:

```python
import mbuild as mb
mol1 = mb.load("CC", smiles=True)
mol2 = mb.load("CO", smiles=True)
cpd = mb.Compound()
for i in range 5:
    if i%2:
        cpd.add(mol1)
    else:
        cpd.add(mol2)

import gmso
top = gmso.external.from_mbuild(cpd)
top.save("top.gro")

assert compare_expected_order_of_molecules_in_gro("top.gro")
```

I almost think that these tests might be useful across all engines for writing some interestingly generated topologies. I may address this in a new PR as I'm creating the set of tests for lammps.

chrisiacovella:
I agree.  I'm still working on adding in new tests.  I was definitely going to include testing residue numbering for mixed molecules systems. I need to looking for closely into testing the writing of residues (specifically into whether these are being set correctly when converting/loading).  That may need to be a separate PR depending on how things go.

chrisiacovella:
#711 Has now been addressed in this PR as well (misspelling of residue in `iter_sites` in topology. )

gmso/formats/gro.py
```diff
@@ -166,32 +166,119 @@ def _prepare_atoms(top, updated_positions, precision):
...
+                site_res_id[idx] = res_nr
+                last_res_id = site.residue.number
+                last_name = site.residue.name
+        else:
+            if idx == 0:
+                unaffiliated = 1
+                site_res_id[idx] = 1
+            elif unaffiliated == -1:
+                res_nr = res_nr + 1
+                unaffiliated = res_nr
```
daico007, line 237:
```suggestion

    # we need to sort through the sites to provide a unique number
    # for each molecule and residue
    # we will store the unique id in dictionary
    # where the key is the idx

    site_res_id = dict()
seen = dict()
for idx, site in enumerate(top.sites):
    if site.molecule: 
        if site.molecule not in seen: 
            seen[site.molecule] = len(seen) + 1
        site_res_id[idx] = seen[site.molecule]
    elif site.residue:
        if site.residue not in seen:
            seen[site.residue] = len(seen) + 1
        site_res_id[idx] = seen[site.residue]
    else:
        if "MOL" not in seen:
            seen["MOL"] = len(site_res_id) + 1
        site_res_id[idx] = seen["MOL"]
```

daico007: There is no issue with the old logic, I just clean up the code a little bit and do away with juggling too many temp variables.

gmso/formats/gro.py
```diff
@@ -166,32 +166,119 @@ def _prepare_atoms(top, updated_positions, precision):
...
+            res_name = (
+                site.residue.name
+                if len(site.residue.name) <= 5
+                else site.residue.name[:5]
+            )
+            site.label = f"res_id: {res_id}, " + site.label
         else:
-            res_id = 1
+            res_id = unaffiliated
             res_name = "MOL"
```
daico007, line 259:
```suggestion
            res_id = site_res_id[idx]
            res_name = "MOL"
```

daico007: to match with the change proposed above

daico007 (commented):
Some small proposed change to make a code a bit easier to read, all others stuff LGTM!

daico007 (approved):
LGTM! All unit tests passed so I gonna merge this PR shortly.

PR #714, opened 2023-03-22
writing out pairs section of top file

daico007:
I just realized the `top.py` is not writing out the `[ pairs ]` sections in the `.top` file, which result in the loss of 1-4 interactions (vdw and electrostatics). This PR adds worker method to write out that section.


chrisiacovella:
So it looks like you just need to update the testing comparison files to be the ones that now include pairs (and some minor changes to formatting re: tabs).  I did some hand validation of benzene and it looks correct. 

On a similar note, this only works with gen-pairs = yes .  We should probably have a separate PR later that allows these to be explicitly written out (i.e., with parameters as well).   If we are using something like CHARMM, there are cases where we will use different parameters explicitly for certain 1-4 parameters.  I assume this would require adding in a new data structure in topology.

daico007:
I will update the tests at some point this week! Also agree with you on `gen-pairs`, that it needed to be handled in a different PR. We actually already have a special structure for special pair potential (https://github.com/mosdef-hub/gmso/blob/main/gmso/core/pairpotential_type.py), but it still needed to be tested more and implemented in these writers.

daico007:
Test files have been updated and all the tests is passing now.

chrisiacovella:
This looks good!  We should probably have another discussion about what is the best way to check the output files so that we can have consistency across the writer tests.  This could potentially also be related to the stress testing suite we mentioned, where we have some scripts for the engines that actually try to run what was generated.

chrisiacovella:
It looks like the generate_pairs function doesn't have a direct test (only indirect via what is written to file).  It would probably be good to add that before we merge.

daico007:
I have some additional tests for this PR (not pushed yet), basically writing out and comparing the new files against some reference files.

daico007:
@chrisiacovella, I added more unit tests (include the `_generate_pairs`) and all of them seems to pass now.

chrisiacovella (approved):
This looks good!

Issue #717, opened 2023-04-25
Performance issue when converting typed topology to parmed 


daico007:
For large system, conversion going from a typed topology to parmed can be quite slow. A quick profiling determine that this has to do with a check we perform to make ensure that the converted type is supported via parmed. The check uses sympy to parse the expression and is the bottle neck of the conversion. Replacing sympy with symengine may mitigate the issue.

chrisjonesBSU:
@daico007 can we close this since you merged pr #719 ?

PR #719, opened 2023-04-26
Improve convert to parmed performance

daico007:
Address #717. Replacing `sympy` `parse_expr` with `symengine.expand` for all the equation checking/assertion.


daico007:
Yeah, I can do some detail profiling and see how much speed up does symengine offer and see if it makes sense for use to replace sympy completely/

Issue #718, opened 2023-04-25
GROMACS pairs edge case 


daico007:
When I tried to use parmed to load in top file written out using GMSO, I bumped into an error stating that the number of pairs is more than the true 1-4 pairs. Further inspection determined that this was due to edge cases that were not caught during previous PR, i.e., 4-member ring and 5-member ring. Currently, pairs were determined from dihedrals (grabbing the first and last connection members) and perform sorting/pruning to avoid duplicating. However, it didn't take care of 4 and 5-member rings, where dihedral's first and last connection members were connected by a path < 3 (hence should be excluded in the pairs section). Potential solution include using networkx (our conversion from topology to networkx graph) to do a check (i.e., calculating the shortest path length) to exclude such cases. From local testing, the algorithm to determine `shortest_path_length` is sufficiently fast (taking ~0.2s to check 33350 pairs), so this can be a feasible solution.

PR #720, opened 2023-04-26
Additional checks when generate pair lists 

daico007:
Address #718. Added checks for `shortest_path_length` when generate pair lists from dihedral. Added unit test for the 4 and 5-ring member. Added options to create a lighter weight networkx graph (not populating angles and dihedrals for each node).


CalCraven (approved):
LGTM!

Issue #716, opened 2023-04-20
Issue in conversion from parmed


CalCraven:
The from_parmed method in GMSO wants to grab bond types if they exist from the structure.bond_types `TrackedList`. However, that is not all unique types, which is the assumption in that code. Working on a fix for this now.

Code to reproduce issue:
```python
import mbuild as mb
import foyer
compound = mb.load("CC1=C(C=CC(=C1)[N+](=O)[O-])N", smiles=True)
oplsaa = foyer.Forcefield(name="oplsaa")
pmd_structure = oplsaa.apply(compound)

from gmso.external import from_parmed
from gmso.core.views import PotentialFilters as pfilters
top = from_parmed(pmd_structure)
pmd_unique_bonds = set(map(lambda x: x.bond_type, pmd_structure.bonds))
assert len(top.bond_types) == len(pmd_unique_bonds
```

PR #721, opened 2023-04-26
Parameterized Parmed Conversions

CalCraven:
This PR should fix minor issues #716 which didn't account for symmetries in conversions from parmed structures. This checks that the expected number of potentials are found, which were only being checked over simple molecules such as ethane, and as such missing that some connection types were repeated.

Additional tests were done to check for urey bradleys, harmonic and periodic impropers, and periodic propers.

### DIFF
--- a/gmso/external/convert_parmed.py
+++ b/gmso/external/convert_parmed.py
@@ -1,5 +1,6 @@
 """Module support for converting to/from ParmEd objects."""
 import warnings
+from operator import attrgetter
 
 import numpy as np
 import unyt as u
@@ -15,6 +16,14 @@ if has_parmed:
     pmd = import_("parmed")
 
 lib = PotentialTemplateLibrary()
+# function to check reversibility of dihedral type
+rev_dih_order = lambda x: x.atom2.type > x.atom3.type or (
+    x.atom2.type == x.atom3.type and x.atom1.type > x.atom4.type
+)
+# function to get the names for a given atomtype
+atomtype_namegetter = lambda x: (
+    attrgetter("atom1.type", "atom2.type", "atom3.type", "atom4.type")(x)
+)
 
 
 def from_parmed(structure, refer_type=True):
@@ -64,7 +73,6 @@ def from_parmed(structure, refer_type=True):
             structure, angle_types_member_map=angle_types_map
         )
         # Consolidate parmed dihedraltypes and relate to topology dihedraltypes
-        # TODO: CCC seperate structure dihedrals.improper = False
         dihedral_types_map = _get_types_map(
             structure, "dihedrals", impropers=False
         )
@@ -72,13 +80,18 @@ def from_parmed(structure, refer_type=True):
         pmd_top_dihedraltypes = _dihedral_types_from_pmd(
             structure, dihedral_types_member_map=dihedral_types_map
         )
+        pmd_top_rbtorsiontypes = _rbtorsion_types_from_pmd(
+            structure, dihedral_types_member_map=dihedral_types_map
+        )
         # Consolidate parmed dihedral/impropertypes and relate to topology impropertypes
-        # TODO: CCC seperate structure dihedrals.improper = True
         improper_types_map = _get_types_map(structure, "impropers")
         improper_types_map.update(
             _get_types_map(structure, "dihedrals"), impropers=True
         )
-        pmd_top_impropertypes = _improper_types_from_pmd(
+        pmd_top_periodic_impropertypes = _improper_types_periodic_from_pmd(
+            structure, improper_types_member_map=improper_types_map
+        )
+        pmd_top_harmonic_impropertypes = _improper_types_harmonic_from_pmd(
             structure, improper_types_member_map=improper_types_map
         )
 
@@ -113,7 +126,12 @@ def from_parmed(structure, refer_type=True):
             connection_members=[site_map[bond.atom1], site_map[bond.atom2]]
         )
         if refer_type and isinstance(bond.type, pmd.BondType):
-            top_connection.bond_type = pmd_top_bondtypes[bond.type]
+            key = (
+                bond.type.k,
+                bond.type.req,
+                tuple(sorted((bond.atom1.type, bond.atom2.type))),
+            )
+            top_connection.bond_type = pmd_top_bondtypes[key]
         top.add_connection(top_connection, update_types=False)
 
     for angle in structure.angles:
@@ -127,7 +145,12 @@ def from_parmed(structure, refer_type=True):
             ]
         )
         if refer_type and isinstance(angle.type, pmd.AngleType):
-            top_connection.angle_type = pmd_top_angletypes[angle.type]
+            member_types = (
+                angle.atom2.type,
+                *sorted((angle.atom1.type, angle.atom3.type)),
+            )
+            key = (angle.type.k, angle.type.theteq, member_types)
+            top_connection.angle_type = pmd_top_angletypes[key]
         top.add_connection(top_connection, update_types=False)
 
     for dihedral in structure.dihedrals:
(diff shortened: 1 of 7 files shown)


CalCraven:
I tested this with two different numbers of atoms:
2000 atoms -> 2.24 +- 0.09 s
5000 atoms -> 5.80  +- 0.21 s

So I think we're looking pretty much good to merge this.

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
         # k in (kcal/(angstrom**2 * mol))
         # r_eq in angstrom
-        member_types = angle_types_member_map.get(id(angletype))
         top_angletype = gmso.AngleType(
-            potential_expression=expr, member_types=member_types
+            name=name,
+            parameters=angle_params,
+            expression=expression,
+            independent_variables=variables,
+            member_types=angletype[2],
```
daico007, line 425: Based on how atoms order defined above, wouldn't the `angletype[2]` have the order of `(middle-point, edge-point-1, edge-point-2)`, should that be reordered here to conform with the angle convention where the middle atom(type) is in the middle?

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
-    improper_types_member_map: optional, dict, default=None
-        The member types (atomtype string) for each atom associated with the improper_types the structure
 
     Returns
     -------
     pmd_top_impropertypes : dict
-        A dictionary linking a pmd.ImproperType or pmd.DihedralType
-        object to its corresponding GMSO.ImproperType object.
+        A dictionary linking a pmd.improperType
+        object to its corresponding GMSO.improperType object.
```
daico007, line 581:
```suggestion
        object to its corresponding GMSO.ImproperType object.
```

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
         Parmed Structure that needed to be converted.
-    improper_types_member_map: optional, dict, default=None
-        The member types (atomtype string) for each atom associated with the improper_types the structure
 
     Returns
     -------
     pmd_top_impropertypes : dict
-        A dictionary linking a pmd.ImproperType or pmd.DihedralType
-        object to its corresponding GMSO.ImproperType object.
+        A dictionary linking a pmd.improperType
```
daico007, line 580:
```suggestion
        A dictionary linking a pmd.ImproperType
```

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
-        pmd_top_dihedraltypes[id(dihedraltype)] = top_dihedraltype
-    return pmd_top_dihedraltypes
+        pmd_top_rbtorsiontypes[dihedraltype] = top_dihedraltype
+    return pmd_top_rbtorsiontypes
 
 
-def _improper_types_from_pmd(structure, improper_types_member_map=None):
-    """Convert ParmEd improper types to GMSO ImproperType.
+def _improper_types_periodic_from_pmd(structure):
+    """Convert ParmEd DihedralTypes to GMSO improperType.
```
daico007, line 566:
```suggestion
    """Convert ParmEd DihedralTypes to GMSO ImproperType.
```

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
+    if len(improper_dihedralsList) == 0 or not structure.dihedral_types:
+        return pmd_top_impropertypes
+    unique_improper_types = list(
+        dict.fromkeys(
+            [
+                (
+                    improper.type.phi_k,
+                    improper.type.phase,
+                    improper.type.per,
+                    (atomtype_namegetter(improper)),
```
daico007, line 596: Do you think we need to sort the members of improper (last 3 members)? since we are doing it for angle and dihedral.

gmso/external/convert_parmed.py
```diff
@@ -339,146 +386,296 @@ def _angle_types_from_pmd(structure, angle_types_member_map=None):
...
+    pmd_top_impropertypes = dict()
+    if not structure.improper_types:
+        return pmd_top_impropertypes
+    unique_improper_types = list(
+        dict.fromkeys(
+            [
+                (
+                    improper.type.psi_k,
+                    improper.type.psi_eq,
+                    (atomtype_namegetter(improper)),
```
daico007, line 654: same comment as above regarding sorting

gmso/external/convert_parmed.py
CalCraven: Yep that's right. In GMSO/Foyer we define it as the central atom in the middle. I'm guessing that no error arises here because the angle_type is is still associated with the angle (edge-point-1, middle-point, edge-point-2). So the angle itself is defined correctly, and if you printed out:
```python
# case 1: angle with reversed member_types
print(list(map(lambda x: x.atom_type.name, angle.connection_members)) # -> [A, B, C]
print(angle.angle_type.member_types) # -> [B, A, C]

# case 2: for some angle_mirror that is a mirror to angle
print(list(map(lambda x: x.atom_type.name, angle_mirror.connection_members)) # -> [C, B, A]
print(angle_mirror.angle_type.member_types) # -> [B, A, C]

# case 3: for some angle and sorts member types by GMSO addition
print(list(map(lambda x: x.atom_type.name, angle_gmso.connection_members)) # -> [A, B, C]
print(angle_mirror.angle_type.member_types) # -> [C, B, A]
```

So I guess it just depends on what we exactly what member types to entail. I think the central atom should still always go in the middle. However, I think the outer and inner atoms should reflect the sorted ordering of the type, not the actual ordering if the atoms in angle.connection_members:

Fixed code to generate this behavior:
```python
        top_angletype = gmso.AngleType(
            name=name,
            parameters=angle_params,
            expression=expression,
            independent_variables=variables,
            member_types=[angle_type[2][i] for i in [1,0,2]
        )

gmso/external/convert_parmed.py
CalCraven: I would say yes, but I think I need to clearly understand what the way to do that is. We have so few cases for real impropers, and I'm not certain pmd.Impropers is sorted the same as pmd.dihedrals with improper=True. I think I need to figure those cases out.

gmso/external/convert_parmed.py
CalCraven: Based on conversation with @daico007, we want to sort all member_types by the connection.connection_member indices. In this case, types can be made that are mirrors of each other, and only filtered into identical types if the correct filter is passed to the PotentialViews object. This change will be reflected shortly and applied to all connections in the parmed structure.

CalCraven:
The latest commit has addressed the changes to parmed conversion. Required changes due to sorting of GMSO object from ParmEd by topology index:

- [x] Tests using parmed objects have different ordering
- [x] New potential filter UNIQUE_SORTED_NAMES
- [x] Comparison of unique types using different potential filters
- [x] ParmEd back and forth conversions will result in some connections that are flipped

gmso/external/convert_parmed.py
CalCraven: Addressed in [408ef1b](https://github.com/mosdef-hub/gmso/pull/721/commits/408ef1ba5b6b613ad3f9f541d395a4a99daaa1e5)

gmso/external/convert_parmed.py
CalCraven: Addressed in [408ef1b](https://github.com/mosdef-hub/gmso/pull/721/commits/408ef1ba5b6b613ad3f9f541d395a4a99daaa1e5). Impropers are sorted by the index of the 2-4 sites, and the central atom is always listed first.

gmso/external/convert_parmed.py
CalCraven: Addressed in [408ef1b](https://github.com/mosdef-hub/gmso/pull/721/commits/408ef1ba5b6b613ad3f9f541d395a4a99daaa1e5)

PR #723, opened 2023-05-05
modified to_mbuild function to improve performance

chrisiacovella:
This adds speed improvements to the to_mbuild function, whereby the code now passes lists of Compounds to the add function.


daico007 (approved):
LGTM!

daico007:
Note: this change probably will only come into effect when we have a new release of mbuild (to include this change https://github.com/mosdef-hub/mbuild/pull/1107)

Issue #724, opened 2023-05-05
AttributeError from `base_units` in `to_hoomd_forcefield` method


marjanAlbouye:
I tried the following script to get Hoomd forcefields from a topology and this error is raised when the `base_unit` argument is not passed to the function. 

**Script:**
```python
import foyer
import forcefield_utilities as ffutils
from mbuild.lib.molecules import Ethane
from gmso.external.convert_mbuild import from_mbuild
from gmso.parameterization import apply
from gmso.external import to_hoomd_forcefield

# Start with a mBuild compound
mb_ethane = Ethane()
oplsaa = ffutils.FoyerFFs().load('oplsaa').to_gmso_ff()
# atomtype the system with foyer, and convert the resulting structure to a topology
gmso_ethane = from_mbuild(mb_ethane)
apply(top=gmso_ethane,
      forcefields=oplsaa,
      identify_connections=True)

forces, base_units = to_hoomd_forcefield(gmso_ethane, r_cut=2)
```

**Error:**
```python
File ~/anaconda3/envs/gmso-dev/lib/python3.11/site-packages/gmso/external/convert_hoomd.py:1397, in _convert_params_units(potentials, expected_units_dim, base_units)
   1393     ind_units = re.sub("[^a-zA-Z]+", " ", unit_dim).split()
   1394     for unit in ind_units:
   1395         unit_dim = unit_dim.replace(
   1396             unit,
-> 1397             f"({str(base_units[unit].value)} * {str(base_units[unit].units)})",
   1398         )
   1400     converted_params[parameter] = potential.parameters[parameter].to(
   1401         unit_dim
   1402     )
   1403 potential.parameters = converted_params

AttributeError: 'Unit' object has no attribute 'value'
````

Note: This error is not raised when the `base_unit` is provided or when `auto_scale=True`.

Seems like the error is caused by a typing issue in  `_infer_units` method in `convert_hoomd.py` . 
When the `base_units` is `None`, this method creates a dictionary with proper units for mass, energy and length. However, the type of each unit in that dictionary is `unyt.unit_object.Unit` which does not have an attribute called `value`. 

```python
>>> type(base_units['energy'])

<class 'unyt.unit_object.Unit'>
```
On the other hand, when `auto_scale=True` or when the `base_units` is provided, the final type for units is `unyt.array.unyt_quantity`:
```python
>>> type(base_units['energy'])

<class 'unyt.array.unyt_quantity'>
```
More information:
[...]

PR #725, opened 2023-05-05
Fix HOOMD conversion bug

daico007:
Address this issue #724. Handle case when neither base_units or auto_scale is provided for HOOMD conversion.


chrisjonesBSU (approved):
LGTM. Thanks @marjanAlbouye and @daico007

PR #727, opened 2023-05-09
Fix geometric mean error in HOOMD forcefield

marjanAlbouye:
This PR fixes the error raised by `statistics.geometric_mean()` when sigma or epsilon parameters are zero. Statistics package uses the [logscale](https://en.wikipedia.org/wiki/Geometric_mean) equation to calculate geometric mean but that only works for non-zero positive values, because ln(0) is undefined. 
This PR uses the non logarithmic equation for calculating the geometric mean.

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -4,7 +4,6 @@ from __future__ import division
 import itertools
 import json
 import re
-import statistics
 import warnings
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


chrisjonesBSU:
Thanks for working on this @marjanAlbouye!  We should probably have a unit test where at least one of the sigmas and/or epsilons are zero.

daico007 (approved):
LGTM! Thanks for finding and fixing this bug!

chrisjonesBSU (approved):
LGTM. Thanks :)

PR #729, opened 2023-05-24
Add verbose options for mol2 and element

daico007:
Add verbose options to reduce the amount of warnings raised when parsing mol2 file and element.


chrisjonesBSU (approved):
LGTM. I like this addition, thanks!

PR #730, opened 2023-06-05
Add shift_coord for gro writer

daico007:
I bumped into some odd behavior when writing out `gro` file, i.e., realizing the coordinates are slightly shifted. I thought it's a bug but turn out to be a feature, i.e., a check to move all sites to ensure their coords are all positive. This PR add the option to preserve the original coordinate for the `gro` writer.

### DIFF
--- a/gmso/formats/gro.py
+++ b/gmso/formats/gro.py
@@ -103,7 +103,7 @@ def read_gro(filename):
 
 
 @saves_as(".gro")
-def write_gro(top, filename, precision=3):
+def write_gro(top, filename, precision=3, shift_coord=False):
     """Write a topology to a gro file.
 
     The Gromos87 (gro) format is a common plain text structure file used
@@ -121,6 +121,9 @@ def write_gro(top, filename, precision=3):
         The location and name of file to save to disk.
     precision : int, optional, default=3
         The number of sig fig to write out the position in.
+    shift_coord : bool, optional, default=False
+        If True, shift the coordinate to be all sites to ensure
+        all coordinates are non-negative.
 
     Notes
     -----
@@ -131,7 +134,8 @@ def write_gro(top, filename, precision=3):
 
     """
     pos_array = np.ndarray.copy(top.positions)
-    pos_array = _validate_positions(pos_array)
+    if shift_coord:
+        pos_array = _validate_positions(pos_array)
 
     with open(filename, "w") as out_file:
         out_file.write(


gmso/formats/gro.py
```diff
@@ -121,6 +121,9 @@ def write_gro(top, filename, precision=3):
         The location and name of file to save to disk.
     precision : int, optional, default=3
```
CalCraven, line 122: Random question, but if this is written as 
```x=1000; f"{x:.3f}"```
Then the output is 1000.000. Which is 7 significant figures, not 3. Would it be better to define this as the number of decimals? For the LAMMPS writer, I used `n_decimals` as the name of this parameter. Might be good to use something consistenly.

File: gmso/formats/gro.py [hunk 2]
```diff
@@ -121,6 +121,9 @@ def write_gro(top, filename, precision=3):
...
         The number of sig fig to write out the position in.
+    shift_coord : bool, optional, default=False
+        If True, shift the coordinate to be all sites to ensure
+        all coordinates are non-negative.
```
CalCraven, line 126:
```suggestion
    shift_coord : bool, optional, default=False
        If True, shift the coordinates of all sites by the minimum position 
        to ensure all sites have non-negative positions. This is not a requirement
        for GRO files, but can be useful for visualizing.
```

gmso/formats/gro.py
```diff
@@ -131,7 +134,8 @@ def write_gro(top, filename, precision=3):
 
     """
     pos_array = np.ndarray.copy(top.positions)
-    pos_array = _validate_positions(pos_array)
+    if shift_coord:
+        pos_array = _validate_positions(pos_array)
```
CalCraven, line 138: I was glancing at this function and feel like it should be rewritten to use vectorized operations instead of this a for loop.

```python
def _validate_positions(pos_array):
    """Modify coordinates, as necessary, to fit limitations of the GRO format."""
    if np.min(pos_array) < 0:
        warnings.warn(
            "Topology contains some negative positions. Translating "
            "in order to ensure all coordinates are non-negative."
        )
    min_xyz = np.min(x, axis=1) # smallest position of each coordinate
    min_xyz0 =  np.where(min_xyz<0, min_xyz, 0) #shift by 0 or minimum value of the positions in that coordinate
    pos_array -= min_xyz0
    return pos_array
```

gmso/tests/test_gro.py
```diff
@@ -48,6 +48,10 @@ def test_write_gro(self):
         top = from_parmed(pmd.load_file(get_fn("ethane.gro"), structure=True))
         top.save("out.gro")
 
+    def test_write_gro_with_shift_coord(self):
+        top = from_parmed(pmd.load_file(get_fn("ethane.gro"), structure=True))
+        top.save("out.gro", shift_coord=True)
```
CalCraven, line 53: read_top = Topology("out.gro")
assert np.all(list(map(lambda x: x.positions>=0, read_top.sites)))

CalCraven (commented):
Added one extra test, otherwise it's good to go.

gmso/formats/gro.py
daico007: Good idea, I have changed it to be `n_decimals` for better agreement between engines

gmso/tests/test_gro.py
CalCraven: Not sure how these comments got resolved, @daico007 did you address them somehow?

gmso/formats/gro.py
CalCraven: Same as below comment. I think it's worthwhile to make these changes to this function as well.

daico007: oh shoot, I somehow miss these, I will add these changes in

CalCraven (approved):
LGTM!

Issue #738, opened 2023-06-22
Breaking changes in gsd 3.0


chrisjonesBSU:
GSD 3.0 was recently released, and how some breaking changes in its API.  It moves away from the `Snapshot` terminology and instead uses `Frame`

Example:

```
>>> import gsd.hoomd
>>> snap = gsd.hoomd.Snapshot
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'gsd.hoomd' has no attribute 'Snapshot'
>>> frame = gsd.hoomd.Frame
>>>
```

This causes some issues in `convert_hoomd`.  I changed the version pinning for GSD in #737 as the tests were failing, but we'll still have to make some changes to support both gsd 2 and gsd 3.

chrisjonesBSU:
I have a PR ready to go that adds some checks to `io.py` and `convert_hoomd` that supports both gsd 2.0 and 3.0.  

I'm curious though what others think is the best way to proceed.  `gsd.hoomd.Frame` is introduced in `gsd 2.8` So, we could update `convert_hoomd` to use `Frame` and still technically support both major versions with `gsd >= 2.8`

daico007:
If they (`gsd.hoomd.Frame` and `gsd.hoomd.Snapshot`) is doing the same thing, my vote is to update all the syntax to work with `gsd.hoomd.Frame` and pin gsd to be >2.8

CalCraven:
I would second the adoption of gsd.hoomd.Frame, and pinning to >2.8.

PR #739, opened 2023-06-23
Support GSD 3

chrisjonesBSU:
This PR makes a few changes that supports using GSD 3.0 (#738 ).  I pinned the version to `gsd>=2.9` since the new API was introduced then. 

A couple of things to note:

- hoomd's internal snapshot functionality still uses `Snapshot`
- I decided to keep `snapshot` in most of the variable names we use. If it makes more sense to update these names to use frame as well I can.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -15,7 +15,7 @@ dependencies:
   - openbabel>=3.0.0
   - foyer>=0.11.3
   - forcefield-utilities>=0.2.1
-  - gsd>=2.0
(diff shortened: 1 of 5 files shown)


CalCraven:
Linked to an issue that should fix the ubuntu tests that are failing
https://github.com/yt-project/unyt/issues/431

CalCraven (approved):
LGTM! Nice work.

daico007 (approved):
LGTM!

PR #741, opened 2023-06-27
Speedup identify_connections

daico007:
The current `Topology.identify_connections` method is quite slow for big system, this PR made several changes to the `gmso/utils/connectivity.py` to  improve the performance of said method (change the comparison of site to just their index, use IndexedSet instead of list, etc).

### DIFF
--- a/gmso/utils/connectivity.py
+++ b/gmso/utils/connectivity.py
@@ -119,8 +119,7 @@ def _detect_connections(compound_line_graph, top, type_="angle"):
             else:
                 sorted_conn = match[::-1]
         elif type_ == "improper":
-            latter_sites = sorted(match[1:])
-            sorted_conn = [match[0]] + latter_sites
+            sorted_conn = [match[0]] + sorted(match[1:])
         sorted_conn_matches.append(sorted_conn)
 
     # Final sorting the whole list
--- a/gmso/utils/connectivity.py
+++ b/gmso/utils/connectivity.py
@@ -1,6 +1,7 @@
 """Module supporting various connectivity methods and operations."""
 import networkx as nx
 import numpy as np
+from boltons.setutils import IndexedSet
 from networkx.algorithms import shortest_path_length
 
 from gmso.core.angle import Angle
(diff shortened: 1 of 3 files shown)


chrisjonesBSU:
Speeding this up will be great. Thanks for doing this!

Have you bench marked it yet to get an idea of the performance improvement?

daico007:
I haven't but I sweep up something

CalCraven:
Ran this code locally on a system.

```python
import gmso
import sys
sys.path.append("..")
from basic_testing import gmso_water_graph, box_water, box_polymer
import timeit
import numpy as np
setupStr = """
import gmso;
import sys;
sys.path.append("..");
from basic_testing import gmso_water_graph, box_polymer;
n_mols=100;
poly_box = box_polymer(n_mols);
gmso_top = gmso_water_graph(poly_box);
"""
outList = timeit.repeat("gmso_top.identify_connections()", setup=setupStr, number=3, repeat=3)
print(f"Identify {gmso_top.n_angles + gmso_top.n_dihedrals} Connection in"
      + f" {np.mean(outList):.2f} +- {1.96*np.std(outList)/np.sqrt(3):.3f} seconds"
)
```

```Output with this PR
Identify 1491 Connection in 0.92 +- 0.002 seconds
```

```Output with current GMSO main
Identify 1491 Connection in 1.24 +- 0.002 seconds
```

Which is a a 25% speedup for this system

gmso/utils/connectivity.py
```diff
@@ -120,14 +114,12 @@ def _detect_connections(compound_line_graph, top, type_="angle"):
...
-            if top.get_index(match[0]) < top.get_index(match[-1]):
+            if match[0] < match[-1]:
                 sorted_conn = match
             else:
                 sorted_conn = match[::-1]
         elif type_ == "improper":
-            latter_sites = sorted(
-                match[1:], key=lambda site: top.get_index(site)
-            )
+            latter_sites = sorted(match[1:])
```
CalCraven, line 122: This can be condensed into sorted_conn = [match[0]] + sorted(match]
```suggestion
            sorted_conn = [match[0]] + sorted(match[1:])
```

CalCraven (approved):
LGTM. Did a little plotting as well to verify locally, and I think these changes seem even better at larger number of connections.
image

chrisjonesBSU (approved):
LGTM. The performance improvement looks awesome, thanks @daico007!

gmso/utils/connectivity.py
daico007: added in latest commit

daico007:
I will merge this after all test passes, thanks for doing the profiling Cal!

PR #744, opened 2023-07-10
Fix return statement in `_parse_coulombic`

chrisjonesBSU:
This is a quick fix I noticed when using `convert_hoomd.py` on a topology with no charges.

`_parse_coulombic()` returns `None` if none of the sites have a charge value. This causes an error since the function is called inside of an `extend` method:

```
 763     nbonded_forces = list()
 764     nbonded_forces.extend(
 765         _parse_coulombic(
 766             top=top,
 767             nlist=nlist,
 768             scaling_factors=coulombic_scalings,
 769             resolution=pppm_kwargs["resolution"],
 770             order=pppm_kwargs["order"],
 771             r_cut=r_cut,
 772         )
```

With lists, passing `None` to `extend()` gives `TypeError: 'NoneType' object is not iterable`

Returning an empty list instead of `None` seems like the simplest fix.  I added a unit test that should have caught this as well.

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -800,7 +800,8 @@ def _parse_coulombic(
     )
     if not charge_groups:
         print("No charged group detected, skipping electrostatics.")
-        return None
+        #return None
+        return []
     else:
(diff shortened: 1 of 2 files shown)


daico007 (approved):
LGTM! Thanks for the fix!

PR #745, opened 2023-07-10
Update GMSO to work with pydantic 2.0 

daico007:
[Pydantic v2](https://python.plainenglish.io/pydantic-will-become-even-better-bc6ba9f260d5) just recently comes out and seem to offer a lot of benefits, but also bring along many breaking changes to our current API. This PR will attempt to update GMSO classes to work with the latest version of pydantic.


gmso/lib/potential_templates.py
```diff
@@ -52,7 +48,7 @@ def _load_template_json(item, json_dir=JSON_DIR):
 class PotentialTemplate(AbstractPotential):
     """Template for potential objects to be re-used."""
 
-    expected_parameters_dimensions_: Dict[str, sympy.Expr] = Field(
+    expected_parameters_dimensions: Dict[str, sympy.Expr] = Field(
```
CalCraven, line 51: What's the reason this was changed from `expected_parameters_dimensions_` to `expected_parameters_dimensions`

gmso/lib/potential_templates.py
```diff
@@ -149,16 +145,11 @@ def assert_can_parameterize_with(
...
-        """Pydantic configuration for potential template."""
-
-        allow_mutation = False
-        fields = {
-            "expected_parameters_dimensions_": "expected_parameters_dimensions"
-        }
-        alias_to_fields = {
-            "expected_parameters_dimensions": "expected_parameters_dimensions_"
-        }
+    # TODO[pydantic]: The following keys were removed: `allow_mutation`, `fields`.
```
CalCraven, line 148: I think this TODO can get removed since the model_config seems to be the way to handle it in V2.

CalCraven (approved):
So what is the result of removing the `JSONHandler` class in gmso.abc.serialization_utils? Is that going to change the ability to use the top.save("top.json") method?

Otherwise, minus just some questions for my understanding this is good to go.

gmso/lib/potential_templates.py
daico007: agree

daico007:
The reason `JSONHandler` got removed is because pydantic decided to deprecate the ability to have our own Json handler altogether (we used it before to handle non-serializable property, like unyt and sympy expression). The new way they try to make people do is having the custom json writer determined for each class (this is part of the work in this PR). The JSON reader and writer still works (the unit tests for those are kept mostly intact).

CalCraven:
Excellent, good to know. I'm about to push fixes to these conflicts with the main branch, then we can get this merged.

gmso/lib/potential_templates.py
daico007: good catch. I converted all these to remove the `_` at the end, thinking that the new pydantic has an updated way to handle custom setter/getter. They don't, so I have to revert back to the old way, having a name with `_` at the end and an `alias` for each. I think I just forgot to revert this.

CalCraven: ah okay, then I'll let you handle that. It looks like the same issue pops up with the restraints in atom.py when I tried to merge them, so I can let you revert whatever commit removed the `_` and add the getter and setters.

Issue #748, opened 2023-07-18
Use Forcefield utilities for all backend loading


CalCraven:
We have some older utilities that lived in GMSO for converting gmso forcefields from foyer forcefields, namely for ease of testing. However, this utilities have not completely been updated with modern GMSO, such as handling arrays of parameters in the forcefield. Therefore, these should be deprecated and removed in favor of the much faster forcefield-utilities backend.

PR #749, opened 2023-07-18
Remove convert_foyer_xml and replace with using foyer as the backend

CalCraven:
Issue #748 discusses this, but there are components of forcefield conversion in from Foyer to GMSO that have not been updated in the gmso.external.convert_foyer_xml module, but are supported in forcefield-utilities. Therefore, I think it is worthwhile to move the tests done on this utility to test the same forcefields loaded through that path.

- [x] unit tests
- [x] ForceField.load("foyer.xml", backend="forcefield-utilities") is the default
- [x] Migrate tests to forcefields converted via forcefield-utilities
- [x] issue raised and addressed

### DIFF
--- a/gmso/core/forcefield.py
+++ b/gmso/core/forcefield.py
@@ -730,7 +730,7 @@ class ForceField(object):
         )
 
     @classmethod
-    @deprecate_function("The from XML through GMSO conversion is deprecated.")
+    @deprecate_function("The internal `from_xml` will be deprecated soon. Please load the XML with the `xml_from_forcefield_utilities`.")
     def from_xml(cls, xmls_or_etrees, strict=True, greedy=True):
         """Create a gmso.Forcefield object from XML File(s).
 
--- a/gmso/core/forcefield.py
+++ b/gmso/core/forcefield.py
@@ -573,7 +573,7 @@ class ForceField(object):
         try:
             loader = GMSOFFs()
             ff = loader.load(filename).to_gmso_ff()
-        except ForceFieldParseError:
+        except (ForceFieldParseError, FileNotFoundError):
             loader = FoyerFFs()
             ff = loader.load(filename).to_gmso_ff()
         return ff
--- a/gmso/core/forcefield.py
+++ b/gmso/core/forcefield.py
@@ -7,7 +7,11 @@ from pathlib import Path
 from typing import Iterable
 
 from lxml import etree
-from pydantic import ValidationError
+
+try:
+    from pydantic.v1 import ValidationError
+except:
+    from pydantic import ValidationError
 
 from gmso.core.element import element_by_symbol
 from gmso.exceptions import (
@@ -574,7 +578,7 @@ class ForceField(object):
         try:
             loader = GMSOFFs()
             ff = loader.load(filename).to_gmso_ff()
-        except (ForceFieldParseError, FileNotFoundError, ValidationError):
+        except (ForceFieldParseError, ValidationError):
             loader = FoyerFFs()
             ff = loader.load(filename).to_gmso_ff()
         return ff
--- a/gmso/core/forcefield.py
+++ b/gmso/core/forcefield.py
@@ -578,7 +578,7 @@ class ForceField(object):
         try:
             loader = GMSOFFs()
             ff = loader.load(filename).to_gmso_ff()
-        except (ForceFieldParseError, ValidationError):
+        except (ForceFieldParseError, FileNotFoundError, ValidationError):
             loader = FoyerFFs()
             ff = loader.load(filename).to_gmso_ff()
         return ff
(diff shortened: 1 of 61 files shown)


marjanAlbouye:
I tried this little script for gaff forcefield and looks like there is an issue with the unit of periodicity parameter `n` in periodic torsion potentials in `to_hoomd_forcefield`.

Here's the script:

```python
import mbuild as mb
from gmso.external import from_mbuild, to_hoomd_forcefield

smiles = "c1ccc(S)cc1"
monomer = mb.load(smiles, smiles=True)
topology = from_mbuild(monomer)
topology.identify_connections()
gmso_ff = ForceField("gaff.xml")
apply(topology, gmso_ff)
ff, refs = to_hoomd_forcefield(
    top=topology,
    r_cut=2.5
)
```

The error is:

```error
  File "gmso/gmso/ff_run.py", line 12, in <module>
    ff, refs = to_hoomd_forcefield(
               ^^^^^^^^^^^^^^^^^^^^
  File "gmso/gmso/external/convert_hoomd.py", line 656, in to_hoomd_forcefield
    "dihedrals": _parse_dihedral_forces(
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "/gmso/gmso/external/convert_hoomd.py", line 1109, in _parse_dihedral_forces
    groups[group] = _convert_params_units(
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/gmso/gmso/external/convert_hoomd.py", line 1407, in _convert_params_units
    f"({str(base_units[unit].value)} * {str(base_units[unit].units)})",
            ~~~~~~~~~~^^^^^^
KeyError: 'dimensionless'
```

gmso/core/forcefield.py
```diff
@@ -723,6 +730,7 @@ def _xml_from_gmso(self, filename, overwrite=False):
         )
 
     @classmethod
+    @deprecate_function("The from XML through GMSO conversion is deprecated.")
```
daico007, line 733:
```suggestion
    @deprecate_function("The internal `from_xml` will be deprecated soon. Please load the XML with the `xml_from_forcefield_utilities`.")
```

gmso/external/convert_foyer_xml.py
```diff
@@ -4,8 +4,12 @@
 from lxml import etree
 
 from gmso.exceptions import ForceFieldParseError
+from gmso.utils.decorators import deprecate_function
 
 
+@deprecate_function(
+    "Converting directly to GMSO XML from Foyer XML is deprecated."
+)
```
daico007, line 12:
```suggestion
@deprecate_function(
    "The `from_foyer_xml` method will be deprecated soon. Please use the <alternative method in ffutils>"
)
```

gmso/utils/decorators.py
```diff
@@ -82,3 +82,17 @@ def _inner(*args, **kwargs):
         return _inner
 
     return _function_wrapper
+
+
+def deprecate_function(msg, klass=PendingDeprecationWarning):
+    """Raise a warning that a given function is deprecated."""
```
daico007, line 88:
```suggestion
    """Raise a warning that a given function will be deprecated soon."""
```

gmso/utils/ff_utils.py
```diff
@@ -312,8 +312,12 @@ def _validate_schema(xml_path_or_etree, schema=None):
...
     if not isinstance(xml_path_or_etree, etree._ElementTree):
         ff_xml = etree.parse(xml_path_or_etree)
-
-    xml_schema.assertValid(ff_xml)
+    try:
+        xml_schema.assertValid(ff_xml)
+    except:
+        raise ForceFieldParseError(
+            f"xml scheme {xml_schema} has invalid elements. Check the sections of the file."
+        )
```
daico007, line 320: What error message would the `xml_schema.assertValid` raised? Because the alternative message on the `except` doesn't provide that much info either.

daico007 (commented):
Some comments about the deprecating messages.

gmso/utils/ff_utils.py
CalCraven: Added a more descriptive error, sourced from the error handling Foyer uses when the xml is not generated properly. Although having looked at it, ff_utils module is going to be totally out of use as well once `convert_from_foyer_xml` is deprecated soon. 

So some of these errors are going to have to be tested for on that end, not in GMSO. For now though it's here once this PR is merged.

gmso/utils/decorators.py
CalCraven: Fixed

gmso/external/convert_foyer_xml.py
CalCraven: Fixed in e19

gmso/utils/decorators.py
CalCraven: in e19

gmso/utils/ff_utils.py
CalCraven: Should raise the error handling that is raised by lxml, which will give the section in question that is not matching the force field schema.
```raise ForceFieldParseError((error_text, ex, line))```

daico007 (approved):

PR #752, opened 2023-08-01
Support Pydantic v1 and v2

mattwthompson:
Stopgap bridge prior to #745, see https://docs.pydantic.dev/2.0/migration/#continue-using-pydantic-v1-features

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -20,6 +20,7 @@ jobs:
       matrix:
         os: [macOS-latest, ubuntu-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
+        pydantic-version: ["1", "2"]
 
     defaults:
       run:
@@ -33,7 +34,9 @@ jobs:
         uses: mamba-org/setup-micromamba@v1
         with:
           environment-file: environment-dev.yml
-          python-version: python=${{ matrix.python-version }}
+          create-args: >-
+            python=${{ matrix.python-version }}
+            pydantic=${{ matrix.pydantic-version }}
 
       - name: Install Package
         run: python -m pip install -e .
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -9,7 +9,7 @@ dependencies:
   - unyt<=2.9.2
   - boltons
   - lxml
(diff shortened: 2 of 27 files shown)


mattwthompson:
It looks like this sort of change would need to be propagated through the entire stack all at once

daico007:
There was some additional change introduced in pydantic v1.10.12 that failed our set up, I haven't checked, but those changes probably also causing the failing tests here.

mattwthompson:
It originally launched with a broken build (https://github.com/conda-forge/admin-requests/pull/770 begins the paper trail) but was fixed. There could be other issues, but the one I ran into is not what I'm seeing in these logs

CalCraven:
Yeah the issue is just forcefield-utilities. I think we can quickly release get Pydantic 2 as a PR there and release it so these tests pass. 
Looks like pydantic is in gmso_xml.py and foyer_xml.py only, so that can be done.

mattwthompson:
I've done the absolute minimum work to get that going: https://github.com/mosdef-hub/forcefield-utilities/pull/36

CalCraven (approved):
Probably good as a stopgap here.

CalCraven:
@daico007 merged the PR into ffutils, so we can rerun the tests here and see if the bleeding edge tests pass.

mattwthompson:
Ah, nice, I forgot you all set up those tests. Should be useful!

mattwthompson:
Cool, looks like that run passed: https://github.com/mosdef-hub/gmso/actions/runs/5729710869/job/15560739904

daico007 (approved):
LGTM! Thanks for the fix Matt! I will probably release a patch for forcefield-utilities tomorrow, follow with a patch for GMSO to have this new fix in.

mattwthompson:
Happy to help - could you take on the responsibility of merging this when appropriate?

No rush on a release but will be happy to see it out

daico007:
Yes, for sure, I am planning to have this merged right before releasing (just so the tests on the main CI doesn't appear half-failed), but I could merge this now if you need this for other work.

mattwthompson:
Not urgent at all, this won't cause us major problems for a few months

PR #756, opened 2023-08-11
Cassandra gmso

emarinri:
The goal of this PR is to finish up the MCF writer of GMSO originally started by #560


gmso/formats/mcf.py
```diff
@@ -209,53 +216,54 @@ def _write_atom_information(mcf, top, in_ring):
...
-    types = [site.atom_type.name for site in top.sites]
+    # Based upon Cassandra; updated following 1.2.2 release
+    max_element_length = 6
+    max_atomtype_length = 20
+
+    # Sort to make sure the atom order matches the topology indexing
+    sorted_sites = [site for site in top.sites]
+    sorted_sites.sort(key=lambda site: top.get_index(site))
+    names = [site.name for site in sorted_sites]
+    types = [site.atom_type.name for site in sorted_sites]
```
CalCraven, line 228:
```suggestion
    # Sort to make sure the atom order matches the topology indexing
    sitesList, namesList, typesList = zip(*[(site, site.name, site.atom_type.name) for site in top.sites])
```

gmso/formats/mcf.py
```diff
@@ -209,53 +216,54 @@ def _write_atom_information(mcf, top, in_ring):
...
-            "reduced due to shortening the name to two "
-            "characters."
+            "The number of unique names has been reduced due"
+            f"to shortening the name to {max_element_length} characters."
         )
 
     n_unique_types = len(set(types))
-    for itype in types:
-        if len(itype) > 6:
+    for type_ in types:
```
CalCraven, line 249: Here's how you use the potential filter.
And also, I think you only need to iterate over unique types here, not all types.
```python
from gmso.core.views import PotentialFilters
pfilter = PotentialFilters.UNIQUE_SORTED_NAMES

unique_typesList = top.atom_types(filter_by=pfilter))
for type_ in unique_typesList:
```

I also wouldn't save this to something called `types`. Maybe something like `atypesList` is better. Or `formatted_types`.

gmso/tests/test_mcf.py
```diff
@@ -149,3 +157,78 @@ def test_scaling_factors(self, n_typed_ar_system):
         assert np.allclose(float(mcf_data[-4][1]), 0.4)
         assert np.allclose(float(mcf_data[-4][2]), 0.6)
         assert np.allclose(float(mcf_data[-4][3]), 1.0)
+
```
CalCraven, line 160: I would add a test for if the atomtype names get cut off or something. Otherwise, I think you're just missing 3-5 tests that errors get raised properly if something is done wrong. I think they should include:
1. Try to write out a non-parameterized topology
2. Try to write out something with an incompatible potential.
3. Try to write out something with too many atomtypes
4. Try to write out something with two nonbonded potentials
5. Write out something that isn't charge neutral
6. Write out something that is multiple molecules
- This one maybe wouldn't raise an error if we design the writer to just make that write out two separate files.

We also need a test for the ring structures since those are handled oddly;.

CalCraven (changes requested):
Looking good, thanks for getting on this Eliseo. I think a few changes to use more of the GMSO functionality that's there will help speed this up a lot. Also need a few tests for things we expect to fail.

gmso/formats/mcf.py
emarinri: Great! Thank you for that suggestion.

gmso/tests/test_mcf.py
emarinri: Thanks Cal. I added your suggestions to my todo list.

CalCraven:
In terms of just getting a single one of each molecule, here is something you can use. It could be a method just in the mcf writer, or something that's a method for the `Topology` class.
```python
subtopsList = []
for molecule in top.unique_site_labels(name_only=True):
        subtopsList.append(top.create_subtop("molecule", (molecule, 1)))
```

CalCraven:
Is there any way to add a test that tests that the mcf files can actually be read into Cassandra. Something like the tests for the hoomd simulation in `mBuild` (https://github.com/mosdef-hub/mbuild/blob/1d744186e0c92c6fb5e704aef2c76eaa22cee802/mbuild/tests/test_hoomd.py#L197).

We probably also need to test MCF for an array of forcefields or molecules types.

daico007:
I think since canssandra is conda installable, we should be able to run a short sim (through mosdef_cassandra probably) to confirm the file written out is correctly formatted.

daico007:
I am pulling this PR for local testing. At first test, things seem to write out as expected (through inspection of the mcf file), however, there might some performance, as in it takes 8s to write out a system of one ethanol. I am doing some profiling right now and going through the code to see if there is something can be sped up. I might try to add some test, i.e., running a cassandra simulation, but since I can't install cassandra locally (there isn't a ARM distribution), it might or might not work.

daico007:
Found the source of the slow down (the `sympy.simplify` used to compare equation, swapped out for `symengine.expand` which could do the same thing but faster.

daico007:
> Hi @emarinri, can you check and confirm the conversion of `k` when writing out `dihedrals` and `improper`. Currently, our json has the OPLS dihedral with the form of:
> 
> ```
> {
>   "name": "OPLSTorsionPotential",
>   "expression": "0.5 * k1 * (1 + cos(phi)) + 0.5 * k2 * (1 - cos(2*phi)) + 0.5 * k3 * (1 + cos(3*phi)) + 0.5 * k4 * (1 - cos(4*phi))",
>   "independent_variables": "phi",
>   "expected_parameters_dimensions": {
>     "k1": "energy",
>     "k2": "energy",
>     "k3": "energy",
>     "k4": "energy"
>   }
> }
> ```
> 
> while the Cassandra forcefield is using 𝐸𝜙=𝑎0+𝑎1(1+cos(𝜙))+𝑎2(1−cos(2𝜙))+𝑎3(1+cos(3𝜙)) (https://cassandra-mc.readthedocs.io/en/latest/guides/forcefield.html), so without the 1/2.
> 
> However, the `mcf.py` currently scale down the number by another 1/2 (line 490-515), shouldn't these `k` be multiplied by 2 to match with the cassandra form?

After review this issue a bit more (and writing it out), I believe the current implementation is correct. I think I just confused myself when looking at these earlier.

gmso/tests/test_mcf.py
```diff
@@ -133,19 +386,179 @@
...
+                    break
+
+        with open("nvt_mbuild.out.log", mode="r") as f:
+            lines = f.readlines()
+            for line in lines:
+                if "Total system energy" in line:
+                    energy_ref = float(line.split()[-1])
+                    break
+
+        assert np.isclose(energy, energy_ref, rtol=1e-3)
```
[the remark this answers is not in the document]
CalCraven: set `energy_ref = 0.0` 
before going into this loop.

CalCraven (approved):
Adding the tests to read it into cassandra itself look pretty good. I think there's probably more edge cases to add, but we may have to add them as they come. Some of these test are going to be covered more in the `mosdef_cassandra` package as well.

gmso/tests/test_mcf.py
```diff
@@ -133,19 +386,179 @@ def test_scaling_factors(self, n_typed_ar_system):
...
+        with open("nvt_gmso.out.log", mode="r") as f:
+            lines = f.readlines()
+            for line in lines:
+                if "Total system energy" in line:
+                    energy = float(line.split()[-1])
+                    break
+
+        with open("nvt_mbuild.out.log", mode="r") as f:
+            lines = f.readlines()
+            for line in lines:
```
daico007, line 492:
```suggestion
            energy_ref = 0.0
            for line in lines:
```

daico007 (approved):
I think this is a good wrapping up point for this PR unless you want to add in more stuff @emarinri.

gmso/tests/test_mcf.py
```diff
@@ -133,19 +386,179 @@ def test_scaling_factors(self, n_typed_ar_system):
...
+        # Run the simulation with the GMSO MCF file
+        code, out, err = run_cassandra(cassandra, inp_file)
+
+        assert code == 0
+        assert "complete" in out
+
+        # Parse log files
+        with open("nvt_gmso.out.log", mode="r") as f:
+            lines = f.readlines()
+            for line in lines:
```
daico007, line 485:
```suggestion
            energy = None
            for line in lines:
```

gmso/tests/test_mcf.py
```diff
@@ -381,11 +416,86 @@ def test_in_cassandra(self, typed_ethane):
...
+        )
+
+        py, fraglib_setup, cassandra = detect_cassandra_binaries()
+
+        # TODO: not sure why the cassandra MCF writer of mBuild
+        # outputs a different intramolecular exclusions relative
+        # to the GMSO writer. This is a temporary fix to make the
+        # test pass. We should investigate this further.
+        # Also, try to use the function top.set_lj_scale() and see how
+        # to update subtopologies.
```
daico007, line 432: Is this something we need to look into now?

emarinri: I do not think this is something we have to worry about for now.

emarinri:
@daico007 I agree this is a point where we can wrap this up for now. As you mentioned, there are a few more things to be tested for should be OK for the short term.

gmso/tests/test_mcf.py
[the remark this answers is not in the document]
emarinri: Thanks!

Issue #757, opened 2023-08-23
Performance significantly affected by "flatness" of the compounds


chrisjonesBSU:
@marjanAlbouye and I were experiencing some very slow performance with `gmso.parameterization.apply` and have been doing some digging into what the problem could be.

It doesn't seem to be an issue of number of particles or molecules, and given the examples below, the only thing I can think of is that flat, planar molecules, or in this case polymers, are causing issues.

I'm curious to see if anyone else can re-create this, or has any ideas of what the issue could be. Possibly something related to graphs and sub graphs as a function of topology size and structure?

Examples:

**1. Creating a 20mer poly(ethylene) from SMILES compared to a 20mer using the Polymer Builder.**

- Creating the chain from SMILES gives a very windy, twisty chain. Creating it from the polymer builder gives a flat, straight chain.
- Applying a forcefield to the topology for the molecules from SMILES takes a few seconds. I'm not sure how long it takes for the system using the Polymer Builder because after letting it run for about an hour, it still isn't done.

```
import mbuild as mb
from mbuild.lib.recipes import Polymer
import gmso
from gmso.external import from_mbuild
from gmso.parameterization import apply
import forcefield_utilities as ffutils

import time

oplsaa = ffutils.FoyerFFs().load('oplsaa').to_gmso_ff()

# Polyethylene from SMILES

pe = mb.load("C"*40, smiles=True)
pe_box = mb.fill_box(compound=pe, box=[20, 20, 20], n_compounds=11)
pe_top = from_mbuild(pe_box)
print(pe_box.n_particles)

start = time.time()
apply(top=pe_top, forcefields=oplsaa, identify_connections=True)
print(time.time() - start)

1342 # Num particles
5.7355194091796875 # Time (seconds)
```

![Screenshot from 2023-08-23 13-19-12](https://github.com/mosdef-hub/gmso/assets/50423140/2bbd9799-22f4-4036-9f69-916db9458e99)

```
# Polyethylene from Polymer Builder

monomer = mb.load("CC", smiles=True)

chain = Polymer()
chain.add_monomer(monomer, indices=[2, -2], separation=0.145)
chain.build(n=20)

chain_box = mb.fill_box(compound=chain, n_compounds=11, box=[20, 20, 20])
chain_top = from_mbuild(chain_box)
print(chain_box.n_particles)

start = time.time()
apply(top=chain_top, forcefields=oplsaa, identify_connections=True)
print(time.time() - start)
```

[...]

PR #760, opened 2023-08-29
Turn off identify_connected_components default

daico007:
Per discussion stemmed from #757, this PR turned off the default `identify_connected_components` option in the `parameterize.apply`.

### DIFF
--- a/gmso/parameterization/parameterize.py
+++ b/gmso/parameterization/parameterize.py
@@ -43,7 +43,7 @@ def apply(
         the topology's bonding graph to smaller sub-graphs that correspond to an angle,
         dihedral, improper etc
 
-    speedup_by_molgraph: bool, optional, default=True
+    speedup_by_molgraph: bool, optional, default=False
         A flag to determine whether or not to search the topology for repeated disconnected
         structures, otherwise known as molecules and type each molecule only once.
         This option will be usefult to handle systems with many repeated small molecules,
--- a/gmso/parameterization/parameterize.py
+++ b/gmso/parameterization/parameterize.py
@@ -12,12 +12,9 @@ def apply(
     forcefields,
     match_ff_by="molecule",
     identify_connections=False,
-    identify_connected_components=False,
-    use_molecule_info=False,
-    assert_bond_params=True,
-    assert_angle_params=True,
-    assert_dihedral_params=True,
-    assert_improper_params=False,
+    speedup_by_molgraph=False,
+    speedup_by_moltag=False,
+    ignore_params=["improper"],
     remove_untyped=False,
     fast_copy=True,
 ):
@@ -25,10 +22,10 @@ def apply(
 
     Parameters
     ----------
-    top: gmso.core.topology.Topology, required
+    top : gmso.core.topology.Topology, required
         The GMSO topology on which to apply forcefields
 
-    forcefields: ForceField or dict, required
+    forcefields : ForceField or dict, required
         The forcefield to apply. If a dictionary is used the keys are labels that match
         the molecule name (specified as a label of site), and the values are gmso ForceField objects that gets applied
         to the specified molecule.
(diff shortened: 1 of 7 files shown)


CalCraven (changes requested):
Per this discussion, a few other changes may as well be made here:
```python
    identify_connected_components: bool, optional, default=False
        A flag which will speed up atomtyping for systems with many repeated molecules in
        the system. Specifically useful if `top.unique_site_labels()` isn't properly identifying 
        repeated. However, if the molecules are identified, then it is recommended to use
        use_molecule_info.
        molecules.
```

Also, I would set use_molecule_info to True as default, since that shouldn't add any overhead, since it's just checking the molecule tags. The use_molecule_info text should also be changed. Maybe something like:
A flag to determine whether or not to look at site.molcule_name to look for repeated molecules, and copy the parameters out to the remaining molecules.

I was also thinking we could rename these parameters. There are three potential speedup methods:
identify_connected_components=True
use_molecule_info=True
fast_copy=True

What if we named them something something more cohesive so it was clear if things are running slow for you, you should try to use of of these flags?
Something: 
speedup_idmolecules
speedup_molecule_tag
speedup_copyexpressions

While we're here, what do we think about reformatting assert_x_parameters as well. What if we condensed it to assert_parameter_connections, and you pass a dict instead of 4 different bools? Might reduce how lengthy this function docstring is.

chrisjonesBSU:
I like the idea of using `speedup` in the parameter names to help convey the idea of what these parameters are doing.

daico007:
Good idea! I think we can go with `speedup_by_moltag` and `speedup_by_molgraph`? (short for molecule tag and molecule graph, i.e., connected components)?

daico007:
Regarding the `assert_x_parameters`, I agree that we should to put them in one place/parameter. But I am thinking we could reverse it a little bit, instead of the user assert a set of parameters exist, we could turn it into `ignore_pararameters`, which could take a list/tuple of `('bond', 'angle', 'dihedral', 'improper')`. So in the default case, the `ignore_parameters` will just be `('improper')`.

daico007:
I included the proposed changes in the latest commit, lmk what y'all think.

gmso/parameterization/parameterize.py
```diff
@@ -12,63 +12,47 @@ def apply(
...
-
-    assert_dihedral_params : bool, optional, default=True
-        If True, an error is raised if parameters are not found for all system
-        proper dihedrals.
-
-    assert_improper_params : bool, optional, default=False
-        If True, an error is raised if parameters are not found for all system
-        improper dihedrals.
+    ignore_params : list, optional, default=["impropers"]
+        Skipping the checks that make sure all connections (in the list) have a connection types.
```
chrisjonesBSU, line 55: I think this doc string should list out all possible options that can go in this list just to help avoid any confusion.

gmso/parameterization/parameterize.py
```diff
@@ -12,63 +12,47 @@ def apply(
...
         as a dict.
 
-    identify_connections: bool, optional, default=False
+    identify_connections : bool, optional, default=False
         If true, add connections identified using networkx graph matching to match
         the topology's bonding graph to smaller sub-graphs that correspond to an angle,
         dihedral, improper etc
 
-    identify_connected_components: bool, optional, default=True
+    speedup_by_molgraph: bool, optional, default=True
```
chrisjonesBSU, line 46: I like this parameter name more, and the doc strings do a better job of explaining what this does, but do you think we should add a bit more context about when this parameter might want to be used and when it shouldn't?

Also we should change the `default` here to `False` since the parameter was changed to `False`

chrisjonesBSU (commented):
I left a couple comments for some small changes.

gmso/parameterization/topology_parameterizer.py
```diff
@@ -170,16 +150,20 @@ def _parameterize_connections(
             impropers = top.impropers
 
         self._apply_connection_parameters(
-            bonds, ff, self.config.assert_bond_params
+            bonds, ff, False if "bond" in self.config.ignore_params else True
```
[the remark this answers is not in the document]
daico007: good call, I can add an additional step to clean up the input (forcing them all to be lower case and trimming)

gmso/parameterization/parameterize.py
daico007: I will elaborate the docstring more

CalCraven (approved):
Once those few changes are in, this looks good to me.

gmso/parameterization/topology_parameterizer.py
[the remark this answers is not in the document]
daico007: Added

daico007:
I should have addressed all the comments by @chrisjonesBSU and @CalCraven in the last commit.

chrisjonesBSU (approved):
LGTM! I made one commit to change True to False in the doc string for `speedup_by_molgraph`

gmso/parameterization/parameterize.py
```diff
@@ -12,63 +12,52 @@ def apply(
...
         A flag to determine whether or not to search the topology for repeated disconnected
         structures, otherwise known as molecules and type each molecule only once.
+        This option will be usefult to handle systems with many repeated small molecules,
+        but may slow down system with large molecule, e.g., monolayer.
 
-    use_molecule_info: bool, optional, default=False
-        A flag to determine whether or not to look at site.residue_name to look parameterize
-        each molecule only once. Currently unused.
+    speedup_by_moltag : bool, optional, default=False
+        A flag to determine whether or not to look at site.molecule_name to look parameterize
```
CalCraven, line 53:
```suggestion
        A flag to determine whether or not to look at site.molecule_name to try to parameterize
```

gmso/parameterization/topology_parameterizer.py
```diff
@@ -58,44 +58,24 @@ class TopologyParameterizationConfig(GMSOBase):
...
         description="A flag to determine whether or not to look at site.molecule "
         "to look parameterize each molecule only once. Will only be used if "
-        "identify_connected_components=False",
-    )  # Unused
-
-    assert_bond_params: bool = Field(
-        default=True,
-        description="If True, an error is raised if parameters are not found for "
-        "all system bonds.",
+        "speedup_by_molgraph=False",
```
CalCraven, line 72:
```suggestion
        "speedup_by_molgraph=True",
```

CalCraven:
Few more minor doc string changes and this is good to merge.

gmso/parameterization/topology_parameterizer.py
chrisjonesBSU: Shouldn't this be False? The parameter is set to False by default, which was kind of the original point of this PR. Or am I missing something?

CalCraven: Yes, but that's not what this bit of text is saying. Here's the whole text in context:
``` speedup_by_moltag: bool = Field(
        default=False,
        description="A flag to determine whether or not to look at site.molecule "
        "to look parameterize each molecule only once. Will only be used if "
        "speedup_by_molgraph=False",
```

So that last line should read speedup_by_molgraph=True, since that's when it's used.

chrisjonesBSU: ahh ok I see what you mean.

Issue #736, opened 2023-06-21
Is there a way to remove sites from a topology?


chrisjonesBSU:
Is there a way to remove/delete a particle from a topology? For example, if someone wanted to delete/remove all hydrogens after applying a forcefield?  Obviously, mBuild supports this, and Parmed does as well with the `.strip` method.  

It's a little tricky, we need hydrogens in place in order for the SMARTS matching to work.

Here is a simplified example of how we would do it when using mbuild, Foyer (via Parmed):

```
system = mb.Compound(...)
typed_system = foyer.Forcefield.apply(system) # Parmed Structure
typed_system.strip(hydrogens) # Remove hydrogens
snapshot, ff = mb.create_hoomd_forcefield(typed_system)
```

I'm curious if there is a way to do a similar process in GMSO as well, and what kind of complications would come up if we tried to add it.

daico007:
Hi Chris, I don't think we have that feature yet. There was an attempt by Umesh a while back but it was based on the old API (https://github.com/mosdef-hub/gmso/pull/370). It's probably time to revive that effort. In the mean time, I think you can try manually remove site from the top level (and all connection objects that the site is a part of), but I don't know how smooth that will be (probably gonna be a trial-error process, fixing error raised by writers, etc). I will start working on that `delete_site` method this week.

CalCraven:
There is not a method currently in place for this, but it wouldn't be too difficult to include. The topology stores everything in protected attributes with the `self._attribute` syntax. We can access and modify the topology by creating a @setter decorator that can modify the self._sites IndexedSet using the set `discard` method. The complexity would come where we would have to iterate over the rest of the topology and remove connections that contain that site as well. Not extremely difficult, we already have an [iter_sites method](https://github.com/mosdef-hub/gmso/blob/e488a038fc33146d300d56c9708bd31ddf612246/gmso/core/topology.py#L1286), so could be useful to put together an iter_connections method that would take as `key="connection_member", value=site_to_remove`.

chrisjonesBSU:
Ok, thanks for the help!  I'll try and take a look at this as well.

PR #761, opened 2023-09-01
Add functionality to remove sites and connections from a topology

chrisjonesBSU:
This is still a work in progress, but this PR does a couple things:

**1. Within the `Site` class, we can access all the connections a site is directly involved in.**
- The `Site.connections` property is updated anytime `Topology.add_connection` is used as well as any time `Topology.remove_connection` is used.

**EDIT: After some conversation; this PR instead adds a method to `Toplogy` that returns a dict of all the connections that belong to a site.**

- The idea here is to have more direct access to connections from a particle-up approach rather than topology-down approach.  This way we can avoid larger nested for-loops. 

**2. Adds methods for removing a site and a connection from a topology.**

This addresses #736 and #161 and should replace PR #370 

This is still a WIP, I need to add unit tests and doc strings, and definitely need some feedback in terms of syntax, functionality and to make sure I'm following good pydantic practices. I'm interested what everyone else thinks.

[Here is a gist](https://gist.github.com/chrisjonesBSU/cd2b5b8d4fda72b20e497a022195c95c) showing an example.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1433,6 +1433,12 @@ class Topology(object):
             connections = ["bonds", "angles", "dihedrals", "impropers"]
         else:
             connections = set([option.lower() for option in connections])
+            for option in connections:
+                if option not in ["bonds", "angles", "dihedrals", "impropers"]:
+                    raise ValueError(
+                        "Valid connection types are limited to: "
+                        '"bonds", "angles", "dihedrals", "impropers"'
+                    )
         for conn_str in connections:
             for conn in getattr(self, conn_str):
                 if site in conn.connection_members:
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -672,6 +672,11 @@ class Topology(object):
     def remove_connection(self, connection):
         """Remove a connection from the topology.
 
+        Parameters
+        ----------
+        connection : gmso.abc.abstract_conneciton.Connection
+            The connection to be removed from the topology
+
         Notes
         -----
         The sites that belong to this connection are
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -630,18 +630,6 @@ class Topology(object):
             set([ptype.expression for ptype in self._pairpotential_types])
         )
 
-    def get_connections_by_site(self, site, connections=None):
-        """ """
-        if connections is None:
-            connections = ["bonds", "angles", "dihedrals", "impropers"]
-        connection_dict = dict()
-        for conn_str in connections:
-            connection_dict[conn_str] = []
-            for conn in getattr(self, conn_str):
-                if site in conn.connection_members:
-                    connection_dict[conn_str].append(conn)
-        return connection_dict
-
     def get_lj_scale(self, *, molecule_id=None, interaction=None):
         """Return the selected lj_scales defined for this topology."""
         return self._get_scaling_factor(molecule_id, interaction, "lj_scale", 0)
@@ -661,6 +649,8 @@ class Topology(object):
 
     def remove_site(self, site):
         """"""
+        for connection in self.iter_connections_by_site(site):
+            self.remove_connection(connection)
         self._sites.remove(site)
 
     def remove_connection(self, connection):
@@ -1392,6 +1382,15 @@ class Topology(object):
                     yield site
         else:
             return self.iter_sites("molecule", molecule_tag)
+
+    def iter_connections_by_site(self, site, connections=None):
+        """"""
+        if connections is None:
+            connections = ["bonds", "angles", "dihedrals", "impropers"]
+        for conn_str in connections:
+            for conn in getattr(self, conn_str):
+                if site in conn.connection_members:
+                    yield conn
 
     def create_subtop(self, label_type, label):
         """Create a new Topology object from a molecule or graup of the current Topology.
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -665,6 +665,10 @@ class Topology(object):
         gmso.core.topology.Topology.iter_connections_by_site
             The method that shows all connections belonging to a specific site
         """
+        if site not in self._sites:
+            raise ValueError(
+                f"Site {site} is not currently part of this topology."
+            )
         for connection in self.iter_connections_by_site(site):
             self.remove_connection(connection)
         self._sites.remove(site)
@@ -682,6 +686,10 @@ class Topology(object):
         The sites that belong to this connection are
         not removed from the topology.
         """
+        if connection not in self.connections:
+            raise ValueError(
+                f"Connection {connection} is not currently part of this topology."
+            )
         if isinstance(connection, gmso.core.bond.Bond):
             self._bonds.remove(connection)
         elif isinstance(connection, gmso.core.angle.Angle):
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1412,7 +1412,7 @@ class Topology(object):
         Parameters
         ----------
         site : gmso.core.Site
-            Site to limit connections search to. 
+            Site to limit connections search to.
         connections : set or list or tuple, optional, default=None
             The connection types to include in the search.
             If None, iterates through all of a site's connections.
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -647,6 +647,26 @@ class Topology(object):
             ]
         )
 
+    def remove_site(self, site):
+        """"""
+        for connection in site.connections:
+            try:
+                self.remove_connection(connection)
+            except KeyError:
+                pass
+        self._sites.remove(site)
+
+    def remove_connection(self, connection):
+        """"""
+        if isinstance(connection, gmso.core.bond.Bond):
+            self._bonds.remove(connection)
+        elif isinstance(connection, gmso.core.angle.Angle):
+            self._angles.remove(connection)
+        elif isinstance(connection, gmso.core.dihedral.Dihedral):
+            self._dihedrals.remove(connection)
+        elif isinstance(connection, gmso.core.improper.Improper):
+            self._impropers.remove(connection)
+
     def set_scaling_factors(self, lj, electrostatics, *, molecule_id=None):
         """Set both lj and electrostatics scaling factors."""
         self.set_lj_scale(
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1437,6 +1437,10 @@ class Topology(object):
             Connection where site is in Connection.connection_members
 
         """
+        if site not in self._sites:
+            raise ValueError(
+                f"Site {site} is not currently part of this topology."
+            )
         if connections is None:
             connections = ["bonds", "angles", "dihedrals", "impropers"]
         else:
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -648,13 +648,35 @@ class Topology(object):
         )
 
     def remove_site(self, site):
-        """"""
+        """Remove a site from the topology.
+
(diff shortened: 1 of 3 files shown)


CalCraven:
Time testing:
With this PR
```
THIS PR TESTS for Simple Ethanol
It took 0.3415291670244187 seconds to convert to GMSO
It took 0.006205082987435162 seconds to identify_connections
It took 0.9126980409491807 seconds to load forcefield
It took 0.12859962496440858 seconds to apply forcefield
It took 0.23195625003427267 seconds to write top, gro, lammps

MAIN GMSO TESTS for Simple Ethanol
It took 0.3757550000445917 seconds to convert to GMSO
It took 0.006402791012078524 seconds to identify_connections
It took 0.9505396250169724 seconds to load forcefield
It took 0.13197216694243252 seconds to apply forcefield
It took 0.2722244169563055 seconds to write top, gro, lammps

################################################

THIS PR TESTS for 10000 Ethanols
It took 7.212720583076589 seconds to convert to GMSO
It took 58.01840241695754 seconds to identify_connections
It took 0.7380243750521913 seconds to load forcefield
It took 43.929414541926235 seconds to apply forcefield
It took 183.31577900005504 seconds to write top, gro, lammps

MAIN GMSO TESTS for 10000 Ethanols
It took 7.272310999920592 seconds to convert to GMSO
It took 56.75998824997805 seconds to identify_connections
It took 1.2427897499874234 seconds to load forcefield
It took 43.663405083003454 seconds to apply forcefield
It took 180.0288589169504 seconds to write top, gro, lammps

################################################

THIS PR TESTS for 300-mer Ethylene Glycol
It took 4.278543125023134 seconds to convert to GMSO
It took 1.785474207950756 seconds to identify_connections
It took 0.732234416063875 seconds to load forcefield
It took 19.215889790910296 seconds to apply forcefield
It took 10.554971750010736 seconds to write top, gro, lammps

MAIN GMSO TESTS for 300-mer Ethylene Glycol
It took 0.31096937495749444 seconds to convert to GMSO
It took 1.7851174169918522 seconds to identify_connections
It took 0.746961124939844 seconds to load forcefield
It took 19.24234045902267 seconds to apply forcefield
It took 10.819363666931167 seconds to write top, gro, lammps

```

So the big issue timewise seems to be a 10X reduction in speed for the polymer case when going from mBuild to GMSO. Although that may get worse because site.connections is not being updated after identify_connections is run.

chrisjonesBSU:
These are good points @CalCraven! Thanks for doing some performance analysis as well.

While I like the idea of having quick access to connections from a site's perspective, I'm not totally set on doing it the way the PR currently does (automatically doing the book-keeping behind the scenes). I also like your idea of  using something like the `get_connections_by_site()` method above.

> Although that may get worse because site.connections is not being updated after identify_connections is run.

It should be getting updated after running `identify_connections()`

For example in the below. 

```
>>> import mbuild as mb
>>> methane = mb.load("C", smiles=True)
>>> methane_gmso = from_mbuild(methane)

>>> len(methane_gmso.sites[1].connections)
1
>>> for conn in methane_gmso.sites[1].connections:
...     print(type(conn))
<class 'gmso.core.bond.Bond'>

>>> methane_gmso.identify_connections()
>>> len(methane_gmso.sites[1].connections)
7

>>> for conn in methane_gmso.sites[1].connections:
...     print(type(conn))
<class 'gmso.core.bond.Bond'>
<class 'gmso.core.angle.Angle'>
<class 'gmso.core.improper.Improper'>
<class 'gmso.core.angle.Angle'>
<class 'gmso.core.angle.Angle'>
<class 'gmso.core.improper.Improper'>
<class 'gmso.core.improper.Improper'>
```

chrisjonesBSU:
> I'm happy to hear more feedback, but I think as a whole it is much preferred to modify the topology on the mBuild side, where we don't have a ton of other overhead attached to the topology. Then once things are in GMSO, modifying the structure should be a rare occurrence.

I agree about avoiding modifying the topology on the GMSO side, but I think one common use case of doing this is removing hydrogen atoms. If we do that on the mBuild side before converting to a GMSO topology, then I believe SMARTS matching no longer works.

CalCraven:
Generally speaking this looks great.

> I agree about avoiding modifying the topology on the GMSO side, but I think one common use case of doing this is removing hydrogen atoms. If we do that on the mBuild side before converting to a GMSO topology, then I believe SMARTS matching no longer works.

I see this as a possible workflow, but maybe not a preferred one. A better way would be to use a forcefield with no hydrogens. SMARTS should work fine if there are no hydrogens. For instance, the SMARTS string for methane may go from
[C;X4](H)(H)(H)(H)
to
[C;X0]
But I think it should still work, as long as the topology matches. And obviously you have a force field with no hydrogens.

gmso/core/topology.py
```diff
@@ -647,6 +659,21 @@ def get_scaling_factors(self, *, molecule_id=None):
             ]
         )
 
+    def remove_site(self, site):
```
CalCraven, line 662: I think remove site should automatically iterate through the connections and remove them. That loop that's in the gist should be part of remove_site.
```suggestion
    def remove_site(self, site):
        site_connDict = self.get_connections_by_site(site)
        for conn_type in site_connDict:
            for conn in site_connDict[conn_type]:
                self.remove_connection(conn)
       self._sites.remove(site)
```

CalCraven (commented):
I think maybe you could put some tests together now. We definitely need a check that site is in self._sites, and raise an Error if it's already not there. I also think we could make some of these methods private, so users aren't accessing them all the time. I'd like to hear some feedback on that though. I'm generally of the opinion that `Topology` is too bulky in terms of methods and attributes, and probably could be cleaned up so only the most important user methods are exposed.

chrisjonesBSU:
> Generally speaking this looks great.
> 
> > I agree about avoiding modifying the topology on the GMSO side, but I think one common use case of doing this is removing hydrogen atoms. If we do that on the mBuild side before converting to a GMSO topology, then I believe SMARTS matching no longer works.
> 
> I see this as a possible workflow, but maybe not a preferred one. A better way would be to use a forcefield with no hydrogens. SMARTS should work fine if there are no hydrogens. For instance, the SMARTS string for methane may go from [C;X4](H)(H)(H)(H) to [C;X0] But I think it should still work, as long as the topology matches. And obviously you have a force field with no hydrogens.

That's a good point about changing the SMARTS strings to not include hydrogens, but this may be unwieldy for someone using a large built in forcefield like oplsaa, or gaff via GAFF-Foyer. I just don't think the barrier to running united atom simulations should be _that_ high.

I think the complete workflow as intended works out nicely, at least with Hoomd.  In the example of the gist, if I apply a forcefield, remove hydrogens, then write out the hoomd forcefield objects, I get exactly what I would expect. i.e. Only the carbon atom types are in the LJ parameters, and the angle and dihedral potentials that would have been there in the all-atom system are not written out after removing hydrogens.

CalCraven:
@chrisjonesBSU That's fair, I'm not very familiar with coarse-graining. I don't think I would trust an off the shelf force field developed WITH hydrogens, and expect it to still work when removing the hydrogens and setting all of the charges to 0. But maybe I'm paranoid and this has been well tested somewhere?

chrisjonesBSU:
> 

Well, I'm not trying to make the argument that using AA force fields with UA systems is generally a good practice, and tested.  But, I think having a workflow of starting with an AA forcefield and modifying parameters on the fly to fit whatever UA system and/or properties you're studying is a lot more approachable than first having to create a new xml file and SMARTS strings.  I've come across a few papers where they essentially do this, and it is a great example of how something like gmso can be used as opposed to editing parameters in text files.

chrisjonesBSU:
@CalCraven After re-reading your comment in the issue, I replaced the `get_connections_by_site` method with `iter_connections_by_site`. It seems a bit cleaner than building up and returning a dictionary. Let me know what you think.  

I'll start working on some unit tests and doc strings.

chrisjonesBSU:
This should be good to go. There is still possibly one thing to consider. How do we want to handle it when someone uses `remove_site` or `remove_connection` and passes in a site or connection that isn't in the topology?  We could not raise any errors, since the end result is the same, or should we check and raise an error?  If we don't want to raise errors I think I'll have to use `.discard()` instead of `.remove()` in these methods.

gmso/core/topology.py
```diff
@@ -647,6 +647,45 @@ def get_scaling_factors(self, *, molecule_id=None):
...
+        gmso.core.topology.Topology.iter_connections_by_site
+            The method that shows all connections belonging to a specific site
+        """
+        for connection in self.iter_connections_by_site(site):
+            self.remove_connection(connection)
+        self._sites.remove(site)
+
+    def remove_connection(self, connection):
+        """Remove a connection from the topology.
+
```
marjanAlbouye, line 674: missing docstring for parameters

gmso/core/topology.py
```diff
@@ -1367,6 +1405,34 @@ def iter_sites_by_molecule(self, molecule_tag):
         else:
             return self.iter_sites("molecule", molecule_tag)
 
+    def iter_connections_by_site(self, site, connections=None):
```
marjanAlbouye, line 1408: I have two questions about this method:

1- Is there a reason this is not a private method? 

2- Are there scenarios in which the value of `connections` is something other than the default options you provided (`["bonds", "angles", "dihedrals", "impropers"]`)? If yes, maybe it's worth adding unit tests for such cases.

marjanAlbouye (approved):
LGTM. Thank you Chris!

I added a minor comment and some questions.

gmso/core/topology.py
chrisjonesBSU: 1. I think there are enough use cases for wanting quick access to a site's specific connections to not make it private.

2. I don't think there would be cases where it could be something other than this list. It's really just a list of attribute names in the `Topology` class. Issues could arise if someone has a typo, or uses a singular version (e.g. `bond` instead of `bonds`).  We could throw an error if this happens.

chrisjonesBSU:
> I think maybe you could put some tests together now. We definitely need a check that site is in self._sites, and raise an Error if it's already not there. I also think we could make some of these methods private, so users aren't accessing them all the time. I'd like to hear some feedback on that though. I'm generally of the opinion that `Topology` is too bulky in terms of methods and attributes, and probably could be cleaned up so only the most important user methods are exposed.

I must have overlooked this (regarding my last comment about raising errors). I made it so that an error is raise by these `remove` methods.

I agree about not making the Topology class too bulky, but IMO it seems inconsistent to make `add_site` and `add_connection` methods not private while making their `remove` counterparts private.

daico007:
I have a question regarding expected behavior of the `remove_connection` method. If we remove a bond from the topology, should we also remove the downstream connections (i.e., angle, dihedral, and improper)?

daico007:
I am rerunning the failed bleeding test, need to see if that's something we need to address before merging this PR.

chrisjonesBSU:
> I have a question regarding expected behavior of the `remove_connection` method. If we remove a bond from the topology, should we also remove the downstream connections (i.e., angle, dihedral, and improper)?



> LGTM! Regarding my previous question, I think since the Bond is not a pre-requisite for the creation of Angle/Dihedral/Improper, I think the current behavior should be fine (but it's open for discussion if anyone have any opinion about it).

I think it's a good question. Really, the goal of these new methods isn't to encourage people to start modifying the topology in GMSO. My thought behind the `remove_connection` method was that it wraps the process up nicely to make sure the correct `_connection` attribute is being used.  These if and elif statements could also be moved inside of the `remove_site` method as well.

Ultimately, I agree with @CalCraven that maybe these methods should be private along with the `add` methods (and possibly more in `Topology`). But I still think that should be a separate issue/PR

PR #764, opened 2023-09-13
Attempt to unpin some of our dependencies

daico007:
Local testing indicates that our tests run fine with the later version of numpy and unyt (?). I am not sure what changes might have happened upstream, but I think it's probably good time to unpin those packages.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -6,7 +6,7 @@ dependencies:
   - boltons
   - numpy
   - sympy
-  - unyt
+  - unyt>=2.9.5
   - boltons
   - lxml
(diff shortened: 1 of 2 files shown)


chrisjonesBSU:
I was just about to make a PR to change the version of unyt to the latest :)

Should we unpin these, or re-pin them to the latest versions?  It would be great to incorporate something like dependabot, but from what I could tell, I don't think it currently works with anaconda .yml files?

daico007:
Pinning them to the latest versions (probably with `>=`) is  a good idea! I will also look into dependabot and see if can be used in our cases.

PR #765, opened 2023-09-17
Fix HOOMD issues with GAFF ForceField

CalCraven:
Allow HOOMD writer to write out parameters from a GAFF forcefield.


gmso/external/convert_hoomd.py
```diff
@@ -1370,6 +1417,8 @@
                 base_units[key] = 1 * base_units[key]
     # Add angle unit (since HOOMD will use radian across the board)
     base_units["angle"] = 1 * u.radian
+    # add dimensionless handling
+    base_units["dimensionless"] = 1 * u.dimensionless
```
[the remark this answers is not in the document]
CalCraven: This shouldn't be necessary since base_units isn't given a default argument value.

gmso/external/convert_hoomd.py
```diff
@@ -1126,55 +1128,115 @@ def _parse_dihedral_forces(
...
+        ):
+            dihedral_forces.extend(
+                dtype_group_map[group]["parser"](
+                    container=dtype_group_map[group]["container"](),
+                    dihedrals=groups[group],
+                    expected_units_dim=expected_unitsDict[group],
+                    base_units=base_units,
+                )
+            )
+            raise ValueError
```
daico007, line 1179: Is there supposed to be an `else` clause here to raise this `ValueError`?

gmso/external/convert_hoomd.py
```diff
@@ -1091,26 +1089,30 @@ def _parse_dihedral_forces(
...
-        groups[group] = _convert_params_units(
-            groups[group],
-            expected_units_dim,
-            base_units,
-        )
+        # groups[group] = _convert_connection_params_units(
+        #    groups[group],
+        #    expected_units_dim,
+        #    base_units,
+        # )
```
daico007, line 1115:
```suggestion
```

gmso/utils/schema/ff-gmso.xsd
```diff
@@ -250,22 +249,6 @@
...
-            <xsd:field xpath="@name"/>
-        </xsd:unique>
-        <xsd:unique name="dihedraltypes_names">
-            <xsd:selector xpath="DihedralTypes/DihedralType"/>
-            <xsd:field xpath="@name"/>
-        </xsd:unique>
-        <xsd:unique name="pairpotentialtypes_names">
-            <xsd:selector xpath="PairPotentialTypes/PairPotentialType"/>
-            <xsd:field xpath="@name"/>
-        </xsd:unique>
```
daico007, line 268: Why is this section removed?

daico007 (commented):
Some more comments. Mostly minor, but I have a question on the changes to the schema which I think needed to be clarify before merging.

gmso/utils/schema/ff-gmso.xsd
CalCraven: This section seems to be checking that there are no duplicate names for these potentials. But I don't thing we should be checking the names at all, because then we have like 50 dihedrals that have to be named PeriodicTorsionType-1...50, and if any of them are not labeled this way than it'll fail to load. However, the real key to checking for unique tyeps is checking that no combinations of the types and classes that make up the components of the connection type. I'm not certain why these were implemented this way at all, because in Foyer XML you don't need unique names for each type.

daico007:
Included in #765

gmso/utils/schema/ff-gmso.xsd
daico007: Yeah, I think I am ok with that. Let's comment that section out before we can find a way to validate the uniqueness of these types via atomtype/atomclass

daico007 (approved):
I think all my issues have been resolved, so this should be good to go once all tests pass.

Issue #766, opened 2023-09-18
Performance issue with `to_parmed`


daico007:
Currently, the `to_parmed` method is current convert every individual potential type, which is really slow for big system (this works before when we still store only unique potential type). With the new changes, we should modify the method to only convert the unique types to cut down on the time.

daico007:
Some profiling indicates unit conversion is the culprit

<img width="1316" alt="Screenshot 2023-09-19 at 15 50 10" src="https://github.com/mosdef-hub/gmso/assets/43968221/4fbd1800-1428-40ae-b088-bbd41408dd5c">

PR #768, opened 2023-09-21
Speedup `to_parmed` conversion

daico007:
Address #766. A few minor changes that help speed up the `to_parmed` conversion.


CalCraven (approved):
LGMT if this offers enough speedup! Do we want to do the same for connectiontypes, or is it not impacting things as much?

daico007:
all the connection types are actually using a different filter (`pfilter = PotentialFilters.UNIQUE_PARAMETERS`), so that might have already been included.

PR #769, opened 2023-09-21
Sort indices in `Frame.bond.group`

chrisjonesBSU:
This is a quick change that sorts the particle indices before appending to `Frame.bond.group`.  This should have no effect on bond types or typeids.  The `bond.group` is just a list of length 2 with the particle indices in the bond.

The reason why I think we should do this is because the locality and neighbor list modules in freud require that the bond arrays are sorted. Also, I think its cleaner and easier to read if you're ever having to read these arrays manually for some reason.

I'm not sure why this was never a problem for me when using the hoomd writers in mBuild, maybe its some how related to how Parmed indexes particles in bonds.


CalCraven:
This seems reasonable especially because getting things into a format for freud is good. Yeah it looks like things go to parmed systems. Could be a good test to make sure the mbuild snapshot writer handles this the same way as the gmso one.

daico007:
Would it be easier/faster to sneak this into https://github.com/mosdef-hub/gmso/pull/765? Or would these clash/out-of-scope?

daico007:
That #765 basically contains https://github.com/mosdef-hub/gmso/pull/762, which try to deal with all the sorting too

chrisjonesBSU:
We could definitely just include this in #765

daico007 (approved):

daico007:
I think the change is small enough, and that it has no impact on the other things (bond types and typeids), I think I gonna merge this and do a release of GMSO this morning. We could work on adding more tests later.

PR #770, opened 2023-10-07
Add position restraints for Site

daico007:
Add `position_restraints` for site and write it out in GROMACS top writer. Pending unit test.


CalCraven:
I think the tests should go in `test_atom.py` and `test_top.py`. One that you can add a restraint to a site and parse it, and one that a topology with restraints will properly be id'd in gromacs. I think the best way to do that would be to make a `restrained_water_topology` in `base_test.py` for other engines to use, and then call that.

daico007:
This PR includes the fix for the current failing CI tests.

daico007:
All unit tests has been added.

gmso/utils/files/spce.xml
```diff
@@ -0,0 +1,22 @@
+<ForceField version="0.0.1" name="SPC/E Water" combining_rule="geometric">
```
CalCraven, line 1: I think this file already exists in the `utils/files/gmso_xmls/test_ff_styles` directory. I think you can probably delete it and source from there.

gmso/tests/base_test.py
```diff
@@ -165,6 +165,24 @@ def _typed_topology(n_sites=100):
 
         return _typed_topology
 
+    @pytest.fixture
+    def spce_water(self):
+        spce_comp = mb.lib.molecules.water.WaterSPC()
+        spce_ff = ForceField(get_fn("spce.xml"))
```
CalCraven, line 171:
```suggestion
        spce_ff = ForceField(get_fn("gmso_xmls/test_ffstyles/spce.xml"))
```

CalCraven (changes requested):
Just remove the duplicate files and it should be good to go!

gmso/tests/base_test.py
daico007: Good catch! Done.

CalCraven (approved):
Once tests pass, lets merge it.

Issue #774, opened 2023-10-17
Bug with GRO reader


daico007:
Currently the our gro reader is separating atom name/id, residue name/id by splitting the line. This would be fine if the atom name/id, residue name/id is short enough that there is white space between them. However, this approach would fail for bigger system, where the atom/residue id get long and would "attach" to the atom name, leaving no white space. Hence, we should do away with this approach. 

Since the `gro` is a fixed format, these fields are only allocated a fixed space in the file, we should instead slice the line/string (and `strip` to clean up) to populate these fields.

PR #775, opened 2023-10-17
Fix bug in GRO reader

daico007:
Address #774. Slicing line/string to populate atom/residue name/id when parsing gro file. Pending unit test.

### DIFF
--- a/gmso/formats/gro.py
+++ b/gmso/formats/gro.py
@@ -42,8 +42,8 @@ def read_gro(filename):
     Gro files do not specify connections between atoms, the returned topology
     will not have connections between sites either.
 
-    Currently this implementation does not support a gro file with more than 1
-    frame.
+    Currently this implementation does not support parsing velocities from a gro file or gro file
+    with more than 1 frame.
 
     All residues and resid information from the gro file are currently lost
     when converting to `topology`.
@@ -82,6 +82,11 @@ def read_gro(filename):
             site.molecule = (res_name, res_id)
             site.residue = (res_name, res_id)
             top.add_site(site, update_types=False)
+
+        if len(positions) == 6:
+            warnings.warn(
+                "Velocity information presents but will not be parsed."
+            )
         top.update_topology()
 
         # Box information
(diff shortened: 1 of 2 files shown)


gmso/formats/gro.py
```diff
@@ -57,7 +57,12 @@ def read_gro(filename):
         coords = u.nm * np.zeros(shape=(n_atoms, 3))
         for row, _ in enumerate(coords):
             line = gro_file.readline()
-            content = line.split()
+            res_id = int(line[:5])
```
daico007, line 60:
```suggestion
            res_id = int(line[:5].strip())
```

gmso/formats/gro.py
```diff
@@ -57,31 +57,30 @@ def read_gro(filename):
...
                 raise ValueError(msg.format(n_atoms))
+            res_id = int(line[:5].strip())
+            res_name = line[5:10].strip()
+            atom_name = line[10:15].strip()
+            atom_id = line[15:20].strip()
 
-            res = content[0]
-            atom_name = content[1]
-            atom_id = content[2]
+            positions = line[20:].split()
```
chrisiacovella, line 72: The positions are still fixed width.  I think it would likely be better to use explicit character locations to be completely consistent with the original file format. Granted, this should produce the same exact result, since there is always a space between each component of the position, but I think better to follow the explicit documentation. 

I think it would be good to add a comment in the code here as well, with the file format info:

> residue number (5 positions, integer)
> residue name (5 characters)
> atom name (5 characters)
> atom number (5 positions, integer)
> position (in nm, x y z in 3 columns, each 8 positions with 3 decimal places)
> velocity (in nm/ps (or km/s), x y z in 3 columns, each 8 positions with 4 decimal places)
> 
> "%5d%-5s%5s%5d%8.3f%8.3f%8.3f%8.4f%8.4f%8.4f"

Might also want to update doctoring to note that velocity is ignored if not provided in the file.

chrisiacovella (changes requested):
Minor changes.  I think the code should doing the coordinate breakdown via fixed width, not split, to be consistent.

gmso/formats/gro.py
daico007: Yeah, I think I will need to double check the case with positions, since it seems like their could be both the scenario where there's only 3 decimal and full 8 decimals, so my fear is having fixed width will only work for either case and not both. (like the example in https://manual.gromacs.org/archive/5.0.3/online/gro.html is having the shorter version).
I will add a note about velocity being ignored.

CalCraven (approved):
LGTM!

Issue #778, opened 2023-10-23
Daily CI failing


daico007:
Currently the daily CI is failing for both Linux and Mac with py3.8, potentially due to some changes in our upstream (some dependencies may have stopped supporting py3.8). I will look into it and see if I can do a quick patch.

PR #781, opened 2023-10-27
Drop support/testing for py3.8

daico007:
Relate to #778. I think it's time to drop 3.8 from the testing cycle.


chrisjonesBSU (approved):
LGTM!

PR #782, opened 2023-10-30
Add `to_xml` method to Topology

chrisjonesBSU:
This is a small PR that simplifies the workflow of saving a topology specific xml file by adding a `to_xml` method to `Topology`.  This method just calls both `Toplogy.get_forcefield` and `gmso.core.forcefield.to_xml`, so the unit tests don't check for things like consistency of the new XML file since that is already tested in `test_forcefield.py`.

I think it it will be helpful to have this functionality work as a single step at the `Topology` level.

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1228,6 +1228,22 @@ class Topology(object):
 
         return index
 
+    def to_xml(self, filename, overwrite=False, backend="gmso"):
+        """Save an lxml ElementTree representation of the topology's parameters
+
+        Parameters
+        ----------
+        filename: Union[str, pathlib.Path], default=None
+            The filename to write the XML file to
+        overwrite: bool, default=False
+            If True, overwrite an existing file if it exists
+        backend: str, default="gmso"
+            Can be "gmso" or "forcefield-utilities". This will define the methods to
+            write the xml.
+        """
+        ff = self.get_forcefield()
+        ff.to_xml(filename=filename, overwrite=overwrite, backend=backend)
+
     def to_dataframe(self, parameter="sites", site_attrs=None, unyts_bool=True):
         """Return a pandas dataframe object for the sites in a topology
 
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1240,6 +1240,11 @@ class Topology(object):
         backend: str, default="gmso"
             Can be "gmso" or "forcefield-utilities". This will define the methods to
             write the xml.
+
+        Raises
+        ------
+        GMSOError
+            If the topology is untyped
         """
         ff = self.get_forcefield()
         ff.to_xml(filename=filename, overwrite=overwrite, backend=backend)
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -1250,7 +1250,7 @@ class Topology(object):
             If the topology is untyped
         """
         ff = self.get_forcefield()
-        ff.to_xml(filename=filename, overwrite=overwrite, backend=backend)
+        ff.to_xml(filename=filename, overwrite=overwrite)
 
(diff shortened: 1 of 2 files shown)


daico007 (approved):
LGTM!

gmso/core/topology.py
```diff
@@ -1228,6 +1228,27 @@ def get_index(self, member):
 
         return index
 
+    def to_xml(self, filename, overwrite=False, backend="gmso"):
```
CalCraven, line 1231: Can we rename this? I think it's a bit confusing to have two functions to_xml(). I would suggest `write_forcefield`. 

Also, you can remove the backend argument. The only backend will be ffutils, as gmso file handling is being deprecated.

gmso/core/topology.py
```diff
@@ -1228,6 +1228,27 @@ def get_index(self, member):
...
+        """Save an lxml ElementTree representation of the topology's parameters
```
CalCraven, line 1232:
```suggestion
        """Save an xml file for all parameters found in the topology
```

gmso/core/topology.py
```diff
@@ -1228,6 +1228,27 @@ def get_index(self, member):
...
+        overwrite: bool, default=False
+            If True, overwrite an existing file if it exists
+        backend: str, default="gmso"
+            Can be "gmso" or "forcefield-utilities". This will define the methods to
+            write the xml.
+
+        Raises
+        ------
+        GMSOError
+            If the topology is untyped
```
CalCraven, line 1247: I would add a notes section. Something like about it being useful to write a small forcefield from a large forcefield, such as oplsaa.xml.
```python
Notes
-------
ff = gmso.ForceField("oplsaa")
top = from_mbuild(mb.load("CC", smiles=True))
top = apply(top, ff)
top.write_forcefield("opls-ethane.xml")
```

CalCraven (changes requested):
Few minor suggestions.

chrisjonesBSU:
It looks like the tests in `test_gro.py` are failing with unyt v3. If I downgrade to `2.9.5` they pass for me

daico007:
Yeah, I fixed that unyt issue in this PR https://github.com/mosdef-hub/gmso/pull/770 (https://github.com/mosdef-hub/gmso/pull/770/commits/6a674cec4f8bda5466159f8dae7fa2d7b21bc72e). That PR is just merged, so if you update the branch to `main` that should be fixed.

chrisjonesBSU:
The tutorial looks good! Just one suggestion, I think we should try to use the same formatting we have in pre-commit in the tutorial code snippets.

For example:

```
graphene = Graphene(x_repeat=20, y_repeat=20, n_layers=1,
                    base_units=droplet_system.reference_values)
graphene.apply_forcefield(force_field=OPLS_AA(), remove_charges=True,
                          auto_scale=False, r_cut=2.5)
```

Should become

```
graphene = Graphene(
    x_repeat=20,
    y_repeat=20,
    n_layers=1,
    base_units=droplet_system.reference_values
)

graphene.apply_forcefield(
    force_field=OPLS_AA(),
    remove_charges=True,
    auto_scale=False,
    r_cut=2.5
)
```

I think it is easier to read and following along this way.

gmso/tests/test_topology.py
```diff
@@ -1010,3 +1011,7 @@ def test_units(self, typed_ethane):
             1.60217662e-19 * u.Coulomb,
             0.1 * u.Unit("test_charge", registry=reg.reg),
         )
+
+    def test_write_forcefield(self, typed_benzene_aa_system):
```
CalCraven, line 1015: I think you should merge the two `test_write_forcefield` functions into one.

chrisjonesBSU: Good catch, I didn't notice that. Should be done now.

Issue #792, opened 2023-12-19
Bug in Elementary charges in json loading


CalCraven:
Seems to be an issue in saving and loading a json GMSO object with charges that are elementary charges.

Code to reproduce issue:
```python
import gmso
top = gmso.Topology.load(f"top.lammps")
print(top.sites[0].charge
top.save("top.json", overwrite=True)
top = gmso.Topology.load(f"top.json")
```
Output
```
-0.834 elementary_charge
---------------------------------------------------------------------------
UnitParseError                            Traceback (most recent call last)
Cell In[71], line 5
      3 print(top.sites[0].charge)
      4 top.save("source_files/top.json", overwrite=True)
----> 5 top = gmso.Topology.load(f"source_files/top.json")
      6 for fn in ["gro", "lammps", "json", "xyz", "mcf"]:
      7     print(fn)

File ~/Dropbox/Mac/Documents/Vanderbilt/Research/MoSDeF/gmso/gmso/core/topology.py:1825, in Topology.load(cls, filename, **kwargs)
   1822 from gmso.formats import LoadersRegistry
   1824 loader = LoadersRegistry.get_callable(filename.suffix)
-> 1825 return loader(filename, **kwargs)

File ~/Dropbox/Mac/Documents/Vanderbilt/Research/MoSDeF/gmso/gmso/formats/json.py:326, in load_json(filename)
    324 with filename.open("r") as json_file:
    325     json_dict = json.load(json_file)
--> 326     top = _from_json(json_dict)
    327     return top

File ~/Dropbox/Mac/Documents/Vanderbilt/Research/MoSDeF/gmso/gmso/formats/json.py:184, in _from_json(json_dict)
    182 for atom_dict in json_dict["atoms"]:
    183     atom_type_id = atom_dict.pop("atom_type", None)
--> 184     atom = Atom.model_validate(atom_dict)
    185     top.add_site(atom)
    186     if atom_type_id:

File ~/Dropbox/Mac/Documents/Vanderbilt/Research/MoSDeF/gmso/gmso/abc/gmso_base.py:69, in GMSOBase.model_validate(cls, obj)
     67 @classmethod
     68 def model_validate(cls: Type["Model"], obj: Any) -> "Model":
---> 69     dict_to_unyt(obj)
     70     return super(GMSOBase, cls).model_validate(obj)

[...]

PR #794, opened 2023-12-20
Write charge in elementary charge unit (XML writer)

daico007:
Currently, the XML writer will write out coulomb charge by default. This is partially due to the missing `u.Unit("elementary_charge")` in the unit default registry (see #792 for related issue). This PR will change the default to write out in `elementary_charge`, by using the custom registry `GMSOUnitRegistry` to store the charge unit and write them out to ff xml.


CalCraven (approved):
Looks like all good changes to me! One minor suggestion with calling the unit registry

PR #796, opened 2024-01-11
Sorting of improper types

CalCraven:
Recently, some utilities have been added to gmso/utils/sorting.py. These helped to write out or reorder improper types in the forcefield in a way that was consistent. However, the improper sorting was allowing for the following conditions.

i, j, k, l == i, sorted(k, l, j)

In reality, we need to use the following equivalence:

i, j, k, l == i, sorted(k, l), j

This is because it matters in the improper which atom is considered as the last atom in the list, since the angle from the improper is the angle between planes i,j,k and j,k,l. So only j and k are interchangeable.

Now technically i and l are interchangeable as well, but for the sake of consistency we are defining that the central atom always must go first, so I believe it is useful to stay consistent and leave i and l in their positions, even if mathematically they're equivalent.

- [ ] issues raised/addressed
- [x] tests done
- [x] feature implemented

### DIFF
--- a/gmso/tests/test_dihedral.py
+++ b/gmso/tests/test_dihedral.py
@@ -153,3 +153,35 @@ class TestDihedral(BaseTest):
             tuple(dihedral.connection_members)
             in dihedral_not_eq.equivalent_members()
         )
+
+    def test_sort_dihedral_types(self):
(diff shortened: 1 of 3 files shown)


daico007 (approved):
LGTM!

PR #803, opened 2024-02-12
Update docs landing page

daico007:
Update the landing page of the documentation to include more information.


daico007:
We may also want to update all the landing pages (for mbuild and foyer), to have a same structure.

Issue #767, opened 2023-09-21
HOOMD Forcefield compatibility issues


CalCraven:
HOOMD writer in GMSO is incompatible with forcefields that parameterize connections by a mix of types and classes.

Bug Report:
------------
The majority of the forcefields we have in the ecosystem define the connection types by a set of classes that makes up their components. However, we technically do support defining the connection by types as well, to give users flexibility to define a dihedral specifically while generally defining other connections by their classes (see [this PR discussion](#762 ) for more details). This can cause confusion when identifying unique sets of connection_types by a string that makes up the 4 members; there needs to be some flexibility to write by either atomtype or atomclass. 

This issue will need to be tested by adding a forcefield that behaves this way. Something such as 
```
DihedralType = class1, class2, class3, class4 
AngleType = type1, type2, type3
```

I think we also need some more specificity of how to define a forcefield i.e., is it acceptable to write an bondtype as:
BondType = class1, type2

I believe the plan will be to address this in the coming months

TLDR; `to_hoomd_snapshot` and `to_hoomd_forces` will not work if your topology is defined in such a way as to be have both type and class definitions for connections.

PR #807, opened 2024-03-08
 Add support for `hoomd.md.improper.Periodic` in `to_hoomd_forcefield`

chrisjonesBSU:
This PR starts adding support for periodic improper potentials as they were recently added to Hoomd version 4.5.

I'm running into some issues, I figured I'd get the PR submitted so there are some other eyes on it as well. [Here is a gist of the notebook I'm testing it out with.](https://gist.github.com/chrisjonesBSU/3d5c251890712df7346a25dd5285dc4d). 

One note: I made some changes to the environment file because I was having issues with mBuild plugins using mBuild < 17. This is a separate issue, but the pinned versions in `environment-dev.yml` are changed here to fix that for now.

The contents of the notebook is basically the unit test we have for using GAFF, but with benzene instead of ethanol.

```
import mbuild as mb
import unyt as u
import gmso
from gmso.external import from_mbuild, to_hoomd_forcefield, to_hoomd_snapshot
from gmso.utils.io import get_fn



benzene = mb.load("c1ccccc1", smiles=True)
benzene.box = mb.Box([5, 5, 5,])
top = from_mbuild(benzene)

base_units = dict(mass=u.g/u.mol, length=u.nm, energy=u.kJ/u.mol)
gaff = ForceField(get_fn("gmso_xmls/test_ffstyles/gaff.xml"))
paramaterized_top = apply(top=top, forcefields=gaff, identify_connections=True)
snap, _ = to_hoomd_snapshot(top, base_units=base_units)
forces, _ = to_hoomd_forcefield(top, r_cut=1.4, base_units=base_units)
```

It looks like hoomd improper force params are being written with the wild cards?

```
>>> improper_force = forces["impropers"][0]
>>> for param in improper_force.params:
>>>    print(param)
>>>    print(improper_force.params[param])

ca-*-*-ha
_HOOMDDict{'k': 4.6024, 'n': 2, 'd': 1, 'chi0': 3.141592653589793}
```

This doesn't match the snapshot improper types

```
>>> print(snap.impropers.types)
['ca-ca-ca-ha']
```

I'll keep working on this, but wanted to get it out there for review; maybe I'm missing something obvious at the moment.


environment-dev.yml
```diff
@@ -2,7 +2,7 @@ name: gmso-dev
 channels:
   - conda-forge
 dependencies:
-  - python>=3.8
+  - python<3.12
```
CalCraven, line 5: We probably should also include greater than 3.8 right?

gmso/external/convert_hoomd.py
```diff
@@ -1338,6 +1353,21 @@ def _parse_harmonic_improper(
     return container
 
 
+def _parse_periodic_improper(
```
CalCraven, line 1356: So I think this is fine if we used the `periodicimproperpotential` currently in GMSO. However, I think due to the GAFF file we're using, we're fixing `d = 1` in this potential. I just want to make sure others are on board with this decision, because HOOMD can do simulations when d is not fixed to be 1. The other option would be to make a `hoomdperiodicimproperpotential` json in `gmso/lib/jsons`

CalCraven (approved):

environment-dev.yml
daico007: yes, it should

gmso/tests/test_hoomd.py
```diff
@@ -410,8 +410,9 @@ def test_gaff_sim(self, gaff_forcefield):
             "energy": u.kJ / u.mol,
         }
         ethanol = mb.load("CCO", smiles=True)
```
daico007, line 412: We should just remove this line then

gmso/external/convert_hoomd.py
daico007: I think we should opt for a more flexible version, i.e., when `d` is not fixed to `1`, and have it set to `1` specifically when parsing GAFF

chrisjonesBSU:
So, there is kind of a band-aid fix in here that I think is sort of related to #767, but when the forcefield uses wild cards, these carry over into `member_classes`, so the force entry to hoomd looks something like `[ca, *, *, ha]` while the snapshot's particle types are `[ca, ca, ca, ha]`. 

Right now, the snapshot writer uses `sort_connection_members` while the forcefield writer(s) use `sort_by_classes`

The band-aid fix is to check for `*` in the forcefield members, and if it's found, use `sort_by_members` instead. 

IMO, I think this is fine for now, but there is still the overall larger issue of having more flexibility in both giving instructions on what to use in the writers (types vs classes) as well as flexibility in combining them (as is discussed in #767)

gmso/external/convert_hoomd.py
```diff
@@ -1330,14 +1355,35 @@ def _parse_harmonic_improper(
...
+    for itype in itypes:
+        members = sort_by_classes(itype)
+        # If wild card in class, sort by types instead
+        if "*" in members:
+            members = sort_by_types(itype)
+        container.params["-".join(members)] = {
+            "k": itype.parameters["k"],
+            "chi0": itype.parameters["phi_eq"],
+            "n": itype.parameters["n"],
+            "d": 1.0,
```
CalCraven, line 1382:
```suggestion
            "d": itype.parameters.get("d", 1.0),
```

gmso/external/convert_hoomd.py
chrisjonesBSU: I think this is resolved for now using `.get("d", 1.0)`

chrisjonesBSU:
This should  be ready to merge, the only test failing is `test_charmm_improper_ff` which is unrelated to this PR.

CalCraven (approved):
Lets get #815 merged in to main and to this PR, then I think it's good to go.

CalCraven:
Raise and issue about python < 3.12 though once this is merged, since we'll want to fix that when we get to that issue.

daico007 (approved):
LGTM!

PR #808, opened 2024-03-11
Update connection members during parameterization

CalCraven:
Add tests and change behavior when applying types, update connection connection_members from types. 

This addressed a bug found by @bc118 using NAMD, which write the indices of impropers which must match with the types of the improper types listed. 

image
This screenshot identifies the issues most clearly with impropers. Order is only enforced from identifying the improper in the forcefield. As such, we have to do reordering once the improper_type is identified.

This is a potentially impactful issue if using improper in GMSO. Anyone doing so should check that the improper atoms in their output files are ordered as expected, with the central atom first, and the last atom last.


bc118 (approved):
LGTM.  I tested it on the problematic system, and it seems to run on MoSDeF-GOMC test also.

daico007 (approved):
LGTM! only minor stuff regarding docs and styles

PR #814, opened 2024-04-01
Move pandas dataframe handling to external convert_dataframe module

CalCraven:
This PR looks to improve the handling for converting a topology to a dataframe. This currently lives as a method for topology. It is now being moved to a convert_dataframe.py module. A few different formats are available which give some nice default ways to view a topology. Notably, we have the formats: 
-`publication` which gives all the parameter values you would want to have in a table for publication. This also removes duplicates so each parameter is only listed once.
-`default` some default values which are nice to have
-`remove_duplicates` which allows you to get a smaller dataframe with duplicate rows removed.
-`specific_columns` Allows the user to specify what they want in the dataframe.

There is also an added function that allows you to generate dataframes that cover the parameters for a set of topologies.

Finally, there will be some function that prints the dataframes with the rdkit mols which are labeled to match the dataframes.

TODO Checklist:
- [x] Error checking on arguments
- [x] Replace topology.py dataframe methods/tests
- [x] Doc strings
- [x] Handle units
- [x] Handle parameters that return lists
- [x] Handle parameters that return dictionaries
- [x] Return unique elements if style is publication
- [x] Handle parameter "all" better
- [x] Function to concatenate multiple topogies into one output
- [x] Remove replicate rows flag  -> similar to the publication style, but without the atom_indices section added
- [ ] Function to create a topology with all data from dataframe matching rdkit mol image
   - [x] Could just also make this a format unique_types


gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+        'default' will output default column values of ["name", "atom_type.name", "atom_type.parameters", "charge", "mass"],
+        and any additional attributes in the `columns` argument.
+        'specific_columns' will only output the attributes from the `columns` argument.
+        'publication' will use the default outputs, but remove duplicate values from the dataframes. It adds a column labeled
+        'Atom Indices' to the `sites` dataframe, which enumerates the indices that the atom_type is a part of.
+        `remove_duplicates` will use the labels in passed through the columns argument, and remove duplicates rows in the dataframe.
+        You must also set the `parameter` argument to be one of {"sites", "bonds", "angles", "dihedrals", "impropers"}, not {"all"}
+        See Notes for more details on what this looks like.
+    columns : list of str, optional, default=None
+            List of strings that are attributes of the topology site and can be included as entries in the pandas dataframe.
```
chrisjonesBSU, line 52: Looks like there is an extra tab here

gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+    format : str, optional, default='default'
+        The output formatting style for the dataframe.
+        Options are 'default', 'specific_columns', 'publication', `remove_duplicates`. Defaults to 'default'
+        'default' will output default column values of ["name", "atom_type.name", "atom_type.parameters", "charge", "mass"],
+        and any additional attributes in the `columns` argument.
+        'specific_columns' will only output the attributes from the `columns` argument.
+        'publication' will use the default outputs, but remove duplicate values from the dataframes. It adds a column labeled
+        'Atom Indices' to the `sites` dataframe, which enumerates the indices that the atom_type is a part of.
+        `remove_duplicates` will use the labels in passed through the columns argument, and remove duplicates rows in the dataframe.
+        You must also set the `parameter` argument to be one of {"sites", "bonds", "angles", "dihedrals", "impropers"}, not {"all"}
```
chrisjonesBSU, line 49: I'm kind of confused about this line, and I don't see anything in the notes about it. Is this saying that `remove_duplicates` specifically requires that you don't use `all`

gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+            "Otherwise use a format of default."
+        )
+        if not columns and parameter == "sites":  # default values
+            columns = ["atom_type.name"]
+        elif not columns and parameter in connectionsList:
+            columns = [f"{parameter[:-1]}_type.member_classes"]
+        columnsDict = {parameter: columns}
+    else:
+        raise ValueError(
+            f"Please provide formt=['default', 'specific_columns', 'publication']"
```
chrisjonesBSU, line 161: Small typo here, I think this should say format

chrisjonesBSU: Also, we should add `remove_duplicates` as an option here correct? I think this error message could be a little more clear. Something like `"Available options for format are 'default', 'specific_columns', 'publication', or `remove_duplicates`."` 

Also, the `f` string isn't doing anything here, do we want to print out what they passed to `format`?

gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+                outDict[param], param
+            )
+
+    if format == "remove_duplicates":
+        for df in outDict.values():  # remove duplicate values
+            df.drop("Atom Indices", errors="ignore")
+            df.drop_duplicates(inplace=True, ignore_index=True)
+
+    ###############
+    # END OF FUNCTION
```
chrisjonesBSU, line 203: We could take this out.

gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+            )
+
+    iteritems = getattr(topology, attr)
+    parseFunction = lambda x: functools.reduce(
+        _getattr, [x] + attr_attr.split(".")
+    )
+    return list(map(parseFunction, iteritems))
+
+
+def _pandas_from_parameters(
```
chrisjonesBSU, line 309: Is this function called or used anywhere? Looks like this is the old method from `Topology`, I'm guessing we can remove it here?

gmso/external/convert_dataframe.py
```diff
@@ -0,0 +1,552 @@
...
+            for connection in getattr(self, parameter)
+        )
+    for attr in site_attrs:
+        df = self._parse_dataframe_attrs(
+            df, attr, parameter, sites_per_connection, unyts_bool
+        )
+    return df
+
+
+def _parse_dataframe_attrs(
```
chrisjonesBSU, line 335: We could remove this one as well

chrisjonesBSU (changes requested):
I left some comments after scrolling through the changes. I still need to test this out in a notebook though. I'll be sure to leave another review and/or comment after I do.

Thanks for doing this Cal, this will be a very useful feature!

chrisjonesBSU:
Do you have an example of how to see and access the units when using `handle_unyts="in_headers"`? It doesn't seem to be placing the units in the column names from what I can tell. It does work as expected when using `with_data`. Also, its not clear what the expected behavior of `all_floats` is from the doc strings. Does that mean not to include units at all? Maybe this could be more clear by using a different option name (e.g. `no_units`) or adding an extra sentence to the doc strings.

chrisjonesBSU:
A couple more thoughts:

1. Do we want `parameters` to be passable as a list of parameters? Right now, it is either pick one, or get all of them.  Would we prefer something like this to also work?

```
top_df_dict = to_dataframeDict(
    topology=top,
    parameter=["bonds", "angles"],
    handle_unyts="with_data"
)
```

We could handle this similarly to what we do with `ignore_params` in [apply](https://github.com/mosdef-hub/gmso/blob/afda8cea85e22a1ffc1d7234f68a3a67a2d63111/gmso/parameterization/parameterize.py#L76)

2) This is a matter of preference and style, but these changes use a lot of camel case in variable names and the main function name (`to_dataframeDict`), and I think most of the gmso code base is snake case. Should we try to be consistent here as well?

gmso/external/convert_dataframe.py
CalCraven: Yeah it's not the best I'd agree. How is this?
```
`remove_duplicates` will remove duplicate rows from the dataframe. For sites, this column is `atom_types.name`. For connections, it is the `connection_types.connection_members`. For sites, an additional column will be added, labeled `Atom Indices` that includes the site indexes of members that are the given `atom_type.name`. Because these methods are specific to a given Topology element, the `parameter` argument must be one of {"sites", "bonds", "angles", "dihedrals", "impropers"}, not {"all"}
        See Notes for more details on what this looks like.
```
TODO: I still think an notes example would be helpful.

gmso/external/convert_dataframe.py
CalCraven: Agreed

gmso/external/convert_dataframe.py
CalCraven: Yeah I left in a lot of the old bits as reference, but it's time to remove them.

CalCraven:
> Do you have an example of how to see and access the units when using `handle_unyts="in_headers"`? It doesn't seem to be placing the units in the column names from what I can tell. It does work as expected when using `with_data`. 
There was a bug in the function `_parse_unyts_to_headers`, so fixed that.

> Also, its not clear what the expected behavior of `all_floats` is from the doc strings. Does that mean not to include units at all? Maybe this could be more clear by using a different option name (e.g. `no_units`) or adding an extra sentence to the doc strings.

Added a lot to the docs to fix that. I agree, all_floats is odd. Is `to_float` clearer? No units I feel like could be a little confusing, as it's not clear what the object is converted to, could be an int, could be a string, etc.

PR #815, opened 2024-04-05
Fix failing LAMMPS writer test

daico007:
Define the box size of compound loading from SMILES string --> fix the flaky CHARMM improper ff test.


chrisjonesBSU:
The only difference between the test file and the comparison file were small differences in the atomic positions. As we discussed on a recent dev call this might be due to loading the compound from SMILES in the test. I added a mol2 for the molecule used in this test to `test/files` and updated the atomic positions in the `charmm.lamps` file to use these coordinates now.

CalCraven:
I checked pre-commit locally on this branch, it seems to work fine. I would suggest we merge this in as it'll fix the flaky tests, then we can update pre-commit and nail down any issues there in #816

CalCraven (approved):
LGTM!

PR #819, opened 2024-05-02
Add Ruff to pre-commit 

CalCraven:
This is just a test PR to test the linting with [Ruff](https://github.com/astral-sh/ruff), which is recommended by pydocstyle as a replacement and used by other big python package.


CalCraven:
> If the only issue is the F401 errors in the __init__.py files then maybe we can just ignore that specific error in those files.

Yes that would be good. We may have to do that with a few of the different files this way because it's a bunch of files across the project.

chrisjonesBSU:
Reading the ruff documentation https://docs.astral.sh/ruff/linter/#error-suppression

We can add `# ruff: noqa: F401` to the top of any `__init__.py` files which may be easier than setting the paths to ignore in a config file.

CalCraven:
I would say we should ignore things for the `tests` module. Otherwise, I think it would be wise to perform the rest of the changes just to improper our PEP compliance.

CalCraven:
https://github.com/BrianPugh/python-template/blob/main/.pre-commit-config.yaml

Going to link to a cookie-cutter version of a ruff linter, check out the pyproject.toml as well for some additional formatting stuff we could consider implementing, such as type checking and isort.

.pre-commit-config.yaml
```diff
@@ -8,28 +8,26 @@ ci:
     skip: []
     submodules: false
 repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.4.2
+  hooks:
+    # Run the linter.
+    - id: ruff
+      args: [--line-length=80]
```
chrisjonesBSU, line 17: Should we add the fix flag back in, so that easy formatting fixes are automatically made and comitted on PRs?

chrisjonesBSU (approved):
LGTM, just left one comment about how we want to handle automatic fixes on PRs.

.pre-commit-config.yaml
daico007: yeah, sure! we can always turn it off later if it's become troublesome.

daico007 (approved):

PR #822, opened 2024-05-21
Add python 3.12 to CI workflow and environment files

chrisjonesBSU:
Now that mbuild 0.17.1 is released, I think we're ready to bump everything up to python 3.12 in GMSO.


daico007 (approved):
LGTM!

PR #828, opened 2024-06-21
Fix hoomd version handling in `convert_hoomd.py`

chrisjonesBSU:
I've noticed a bug that you will get `hoomd not defined` error when importing anything from `gmso.external` (even `from_mbuild` and `to_mbuild`) and don't have hoomd in your environment.

This is happening because the hoomd import in `convert_hoomd` is under an if statement, but a few lines later we call `hoomd.version`

@CalCraven @daico007 should we do a quick patch release to get this fix in the conda package asap?

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -30,6 +30,7 @@ if has_gsd:
     import gsd.hoomd
 if has_hoomd:
     import hoomd
+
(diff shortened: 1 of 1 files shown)


daico007:
Yeah, I think this warrants a patch release

CalCraven (approved):
Yea let’s release merge this and then this weekend do another patch release.

PR #834, opened 2024-07-17
Update equation_compare.py

bc118:
Changing 'sympy.symplify' to 'sympy.nsimplify' allows the less simple equations to be solved with the 0.5 decimals in it (example OPLS dihedral).  This was the only was to improve or actually solve for the scalers in the equations when using non-integers in the forms like the OPLS dihedral.

Added me to author list also


daico007 (approved):
LGTM!

PR #848, opened 2024-09-16
Charmm parameters writer

CalCraven:
Port over the old CHARMM/NAMD .prm file writer from mBuild to GMSO. Adding to the savers and allowing for basic charmm style potentials to be written only.

TODOS:
- [x] Check for specified parameter types when writing file
- [x] Add support for urey bradley potential
- [ ] Fix 1-4 scaling in charmm
- [x] Write out all units using unit system
- [ ] Improve iteration over unique type info
- [x] Handle Array of charmm K dihedrals
- [ ] Add test coverage
- [ ] Fix failing tests


gmso/formats/prm_writer.py
```diff
@@ -0,0 +1,124 @@
...
+    -----
+    Follows format according to
+    https://www.ks.uiuc.edu/Training/Tutorials/namd/namd-tutorial-unix-html/
+    node25.html
+    Furthermore, ParmEd can support writing CHARMM par, rtf, str files
+    by converting the parmed.Structure into parmed.CharmmParameterSet
+
+    Parmed stores rmin/2 in "rmin"
+    """
+    unit_system = LAMMPS_UnitSystems("real")  # ang, kcal/mol, amu
```
chrisjonesBSU, line 21: Do we need to make sure we're using this everywhere we write out parameters? I see we use it for the atom mass, but not anywhere else.

gmso/formats/prm_writer.py
```diff
@@ -0,0 +1,124 @@
...
+                    atom3,
+                    atom4,
+                    improper.parameters["phi_k"][0],
+                    improper.parameters["n"][0],
+                    improper.parameters["phi_eq"][0],
+                )
+            )
+
+        f.write("\nNONBONDED\n")
+        nonbonded14 = topology.scaling_factors[0][2]
```
chrisjonesBSU, line 109: This doesn't need to be fixed here, but I feel like this attribute should be more readable rather than knowing that 0th index is LJ and 1st index is electrostatics. Maybe a dict? I can open a separate issue.

chrisjonesBSU (approved):
This LGTM. It seems like we're just missing some type checks to ensure only supported potential formats are being used?

gmso/formats/prm_writer.py
CalCraven: Yep, for in order to handle any input unyts this will do the filtering. I was just too lazy to do it every time at first. Was wondering if you think we should make a system that isn't specific to LAMMPS for this one. It is the exact same as the real units, but maybe a touch confusing to copy them across to other engines. What do you think?

gmso/formats/prm_writer.py
CalCraven: Was thinking the same thing myself. I think it should also be {"electrostatics":{"1-4":0.5}}, instead of a fixed list. Could potentially be some odd 1-5 interactions at some point.

CalCraven:
Cool. Thanks @chrisjonesBSU, yeah I'll punch those in this week to get this writer off the ground. I'm not sure it gets a lot of use, but it'll be nice to have.

gmso/formats/prm_writer.py
chrisjonesBSU: I wonder if `unit_system` should be a parameter of `write_prm` that defaults to `None`. If nothing is passed, then all units are used as-is from the toplogy? If we set `unit_system = topology.unit_system` as default behavior, would that work as well?

```
def write_prm(topology, filename, unit_system=None):
    if not unit_system:
        unit_system = topology.unit_system
```

gmso/formats/prm_writer.py
chrisjonesBSU: Yeah, that's sort of what I was thinking as well.

gmso/formats/prm_writer.py
CalCraven: The PRM file looks to me like it has default assumed units of kCal/mol, angstrom, amu, radians, etc for it's units. So it can only convert to one defined system.

CalCraven:
So I'm going to close this since the charmm writer is somewhat more complex than the simple one we had in mBuild. If anyone wants to call for adding it, we can dig back up this code and implement it pretty easily. Im just hesitant to merge it as is, since the charmm files tend to be more complex than the few sections that are listed in some of the reference files, including have some parameters that would classically be more in line with a simulation control file.

Additionally, we have [MoSDeF-GOMC](https://github.com/GOMC-WSU/MoSDeF-GOMC), which is much more complex in handling all of the checks necessary for writing jointly compatible psf, pdb, .prm files, since a lot of these files are linked based on the naming syntax in these files. So we'll plan to write some soft dependency API through MoSDeF-GOMC for writing these files.

PR #850, opened 2024-09-17
Add support for writing rigid bodies when converting to HOOMD formats.

chrisjonesBSU:
rigid IDs are `ints` assigned to particles/atoms that signal which rigid body they belong to which is needed when recording rigid body information in HOOMD Frames and/or GSD files. The mBuild hoomd writers handled this, but we still need to implement them here.

In mBuild, rigid_id is a property at the particle level, so we can treat it the same here, at the site level.

This is a draft for now. Remaining TODOs are:

- [x] handle cases where some molecules are rigid and others aren't
- [x] throw an error if rigid particle type is found in the topology already


gmso/external/convert_hoomd.py
```diff
@@ -295,6 +293,39 @@ def _parse_particle_information(
             else 1 * base_units["mass"]
         )
         charges.append(site.charge if site.charge else 0 * u.elementary_charge)
+    # Check for rigid IDs
+    rigid_ids = [site.rigid_id for site in top.sites]
```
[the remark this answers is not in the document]
chrisjonesBSU: The thought here is that mBuild assigns rigid IDs at the particle level, so if an mBuild compound has that information, then I was thinking a `site` in GMSO would also, so we read it from `from_mbuild`, and vice versa, write it out in `to_mbuild`.

For hoomd, rigid IDs are assigned at the particle level as well, where the ID points to the molecule they belong to.

I think either way is technically possible. In `from_build` we could disregard the particle rigid IDs, and instead use that information to just adjust `Molecule.isrigid` based on how someone sets up their particle rigid IDs in mBuild. I think this would be sensitive to the mBuild compound correctly having its hierarchy set up? Then, in the GSD writers, we could parse each molecule in the topology, and build up a list of ints that correctly maps sites to rigid body numbers. But, in both of these cases, it seems like this is a roundabout way to having the data structure in place to begin with. 

Maybe mBuild doesn't need to have rigid body information if it no longer has writers, and all rigid body stuff should originate in GMSO?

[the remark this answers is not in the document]
CalCraven: > Maybe mBuild doesn't need to have rigid body information if it no longer has writers, and all rigid body stuff should originate in GMSO?

Yeah, I'd lean at some point to migrate towards that sort of behavior. That is almost a part of the forcefield, since it defines the treatments needed on the particles. 

For now, I think molecule.isrigid is good enough to process that information, since that's what @daico007 set up for the GROMACS writer for rigid water. The molecules are essentially tags that are at the site level, since each molecule is assigned to a site. The edge case of course being if only part of a molecule is "rigid". But I think that is a less common and more complicated case that should be handled differently.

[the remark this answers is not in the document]
chrisjonesBSU: I agree..removing rigid body stuff from mBuild makes sense, and then we can implement however we need to (or already is) in GMSO.

CalCraven:
A remaining question to figure out: What if only some molecules are rigid in a hoomd simulation? How does the list of rigid body indices change?
https://hoomd-blue.readthedocs.io/en/v3.6.0/tutorial/06-Modelling-Rigid-Bodies/00-index.html

CalCraven:
Was playing around with the rigid bodies in HOOMD. Got a working simulation with [some rigid particles and some non-rigid particles](https://gist.github.com/CalCraven/caeb64447b2f2f120c0c28d72f2e0654). Looks like the body argument, shown now in [line](https://github.com/mosdef-hub/gmso/blob/7c656243c273746c12f59bf42e8c964175df0b97/gmso/external/convert_hoomd.py#L348), get's a tag of -1 if the particle is not rigid. The rigid particles get the rigid ID for that particle, and the rigid COM particle, named "dimer" here, also get's that same ID.

chrisjonesBSU:
> Was playing around with the rigid bodies in HOOMD. Got a working simulation with [some rigid particles and some non-rigid particles](https://gist.github.com/CalCraven/caeb64447b2f2f120c0c28d72f2e0654). Looks like the body argument, shown now in [line](https://github.com/mosdef-hub/gmso/blob/7c656243c273746c12f59bf42e8c964175df0b97/gmso/external/convert_hoomd.py#L348), get's a tag of -1 if the particle is not rigid. The rigid particles get the rigid ID for that particle, and the rigid COM particle, named "dimer" here, also get's that same ID.


I believe this is correctly implemented now, but I still need to test it out.

chrisjonesBSU:
The simulations with rigid bodies seem to be working as expected. [Here is a Gist](https://gist.github.com/chrisjonesBSU/d37bf294c0f688f751f98bd67ceb15f1) of the notebook I'm using to run them if you want to try it.

A couple thoughts.

- `to_gsd_snapshot()` and `to_hoomd_snapshot()` now return a dictionary of rigid body info (along with the snapshot and reference values dictionary) that has information needed to create the `hoomd.md.constrain.Rigid` object. After thinking about it more, we could just create `hoomd.md.constrain.Rigid`, populate the dictionary and return it.

This technically breaks the old behavior, since these methods now return 3 objects instead of 2. Also, being required to create a variable or placeholder for this in workflows that don't use rigid bodies might be awkward.  We could add a bool parameter to these methods. Something like `create_hoomd_constraint` that if toggled `True` will also create and return the `hoomd.md.constrain.Rigid` object. Or, we could do this automatically, and the `return` statement used depends on if there are rigid bodies or not. I'm not sure which is preferred or best practice.  Ultimately, I think it would be very helpful if the `hoomd.md.constrain.Rigid` can be made at the same time as the snapshot.

- Should `to_hoomd_forcefield()` have some parameters/checks for rigid bodies as well?

the non-bonded force objects need to include parameters between the rigid particles and their constituent particles. Also, the neighbor list exclusions are different for rigid body simulations than those without.  The gist handles these manually after calling `to_hoomd_forcefield()`. Maybe this method should make these changes if the topology contains rigid bodies.

chrisjonesBSU:
[I've been testing this out a bit.](https://gist.github.com/chrisjonesBSU/e010c3ebd9672b709cefe8d695d9e034) It works with multiple types of rigid molecules (e.g., a system with rigid benzene and rigid ethane), and it works with a mix of rigid and non-rigid molecules (e.g., rigid benzene and flexible ethane), **BUT** the rigid molecules have to come first in the compound/topology hierarchy. 

We can discuss more on a dev call, but there are still some questions.

I think the gsd writers should also build the `hoomd.md.constrain.Rigid` object that is needed to run rigid simulations, since all of the information needed is right there when creating the snapshot, but:

Do we make it so the `gsd` writers always return 3 things now (breaking change) or have the objects returned depend on whether or not rigid bodies exist (not necessarily a breaking change)

So:

```
return snapshot, ref_values, rigid_constrain
```
where `rigid_constrain` is just `None` if there aren't rigid bodies

or 

```
if rigid_bodies:
    return snapshot, ref_values, rigid_constrain
else:
    return snapshot, ref_values
```

The other thing to think about is what behavior we want in `to_hoomd_forcefield`. If the user is using rigid bodies do we:

- Make the necessary changes to the neighbor list (update exclusions to include "body")
- Automatically handle the LJ params between rigid particles and constituent atoms (i.e., add parameters needed to ignore these interactions)
- For systems where every molecule is rigid, do we remove any intramolecular interactions (bonds, angles, dihedrals, impropers) from the forces returned, or just skip their parsing methods to begin with?

Or, do we just return the complete forcefield as is, and leave it to the user to handle these things? In this case, we could make sure to create a thorough tutorial in the documentation, and provide some examples in a notebook.

One idea is possibly having a wrapper around the hoomd writers specific to rigid body stuff, so that the existing writers stay true to form as much as possible. Maybe something like `from external.convert_hoomd import create_rigid_body_sim` that calls `to_gsd_snapshot` and `to_hoomd_forcefield` under the hood, and makes the above changes.

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
     Return
     ------
     gsd_snapshot : gsd.hoomd.Frame
-        Converted hoomd Snapshot.
+        Converted hoomd Snapshot. Always retruned.
     base_units : dict
-        Based units dictionary utilized during the conversion.
+        Based units dictionary utilized during the conversion. Always returned.
```
CalCraven, line 100:
```suggestion
        Base units dictionary utilized during the conversion. Always returned. 
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+    generated using `to_hoomd_forcefield()`.
+
+    If you are using mBuild and GMSO to initialize rigid body simulations
+    with Hoomd-Blue set the site's `Molecule.isrigid` property to `True`.
+    If your topology contains a mix of rigid and flexible molecules,
+    the rigid molecules must come first in the heirarchy of the GMSO toplogy
+    and therefore also in the mBuild compound heirarchy.
+    See https://hoomd-blue.readthedocs.io/en/latest/index.html
+    for more information about running rigid body simulations in Hoomd-Blue.
+
```
CalCraven, line 119:
```suggestion
    For more information on the units from `base_unites`, see https://hoomd-blue.readthedocs.io/en/latest/units.html#base-units
    
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
     Return
     ------
     gsd_snapshot : gsd.hoomd.Frame
-        Converted hoomd Snapshot.
+        Converted hoomd Snapshot. Always retruned.
     base_units : dict
-        Based units dictionary utilized during the conversion.
+        Based units dictionary utilized during the conversion. Always returned.
+    rigid_info : hoomd.md.constrain.Rigid
+        Hoomd object storing constituent particle information
```
CalCraven, line 102:
```suggestion
        A hoomd `Constraint` object storing constituent particle information
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+        needed to run rigid body simulations in Hoomd-Blue.
```
CalCraven, line 103:
```suggestion
        needed to run rigid body simulations in HOOMD-blue.
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+        This is only returned if Molecule's with `Molecule.isrigid = True`
```
CalCraven, line 104:
```suggestion
        This is only returned if sites with `site.molecule.isrigid = True`
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+        are found in the topology.
 
     Notes
     -----
-    Force field parameters are not written to the GSD file and must be included
-    manually in a HOOMD input script. Work on a HOOMD plugin is underway to
-    read force field parameters from a Foyer XML file.
+    Force field parameters are not written to the GSD file and must be
+    generated using `to_hoomd_forcefield()`.
```
CalCraven, line 110:
```suggestion
    generated using `gmso.external.convert_hoomd.to_hoomd_forcefield()`.
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+
+    If you are using mBuild and GMSO to initialize rigid body simulations
+    with Hoomd-Blue set the site's `Molecule.isrigid` property to `True`.
```
CalCraven, line 113:
```suggestion
    with HOOMD-blue set the site's `molecule.isrigid` property to `True`.
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+    If your topology contains a mix of rigid and flexible molecules,
+    the rigid molecules must come first in the heirarchy of the GMSO toplogy
+    and therefore also in the mBuild compound heirarchy.
+    See https://hoomd-blue.readthedocs.io/en/latest/index.html
+    for more information about running rigid body simulations in Hoomd-Blue.
```
CalCraven, line 118:
```suggestion
    for more information about running rigid body simulations in HOOMD-blue.
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
-    Force field parameters are not written to the GSD file and must be included
-    manually in a HOOMD input script. Work on a HOOMD plugin is underway to
-    read force field parameters from a Foyer XML file.
+    Force field parameters are not written to the GSD file and must be
+    generated using `to_hoomd_forcefield()`.
+
+    If you are using mBuild and GMSO to initialize rigid body simulations
+    with Hoomd-Blue set the site's `Molecule.isrigid` property to `True`.
+    If your topology contains a mix of rigid and flexible molecules,
+    the rigid molecules must come first in the heirarchy of the GMSO toplogy
```
CalCraven, line 115:
```suggestion
    the rigid molecules must come first in the hierarchy of the GMSO Topology
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+    and therefore also in the mBuild compound heirarchy.
```
CalCraven, line 116:
```suggestion
    and therefore also in the mBuild compound hierarchy.
```

gmso/external/convert_hoomd.py
```diff
@@ -139,17 +171,20 @@ def to_gsd_snapshot(
...
     if top.n_dihedrals > 0:
-        _parse_dihedral_information(gsd_snapshot, top)
+        _parse_dihedral_information(gsd_snapshot, top, n_rigid)
     if top.n_impropers > 0:
-        _parse_improper_information(gsd_snapshot, top)
+        _parse_improper_information(gsd_snapshot, top, n_rigid)
 
-    return gsd_snapshot, base_units
+    if rigid_info:
+        return gsd_snapshot, base_units, rigid_info
```
CalCraven, line 185: CodeQL Pep8 doesn't like this. Would this be preferred at all:
```suggestion
        return ((gsd_snapshot, rigid_info), base_units)
```
I'm still not opposed to always returning None, if rigid_info is None. Regardless, I think we can tackle this problem down the road, if we redo the HOOMD writers and put them into a single pacakge, instead of keeping them as separate functions that cause a lot of validation headaches between the snapshot and forcefield.

gmso/external/convert_hoomd.py
```diff
@@ -160,9 +195,9 @@ def to_hoomd_snapshot(
     parse_special_pairs=True,
     auto_scale=False,
 ):
-    """Create a gsd.snapshot object (HOOMD default data format).
+    """Create a hoomd.Snapshot objcet (HOOMD default data format).
 
-    The gsd snapshot is molecular structure of HOOMD-Blue. This file
+    The Hoomd snapshot defines the topology of HOOMD-Blue simulations. This file
```
CalCraven, line 200:
```suggestion
    The Hoomd snapshot defines the topology of HOOMD-blue simulations. This file
```

gmso/external/convert_hoomd.py
```diff
@@ -160,9 +195,9 @@ def to_hoomd_snapshot(
...
     can be used as a starting point for a HOOMD-Blue simulation, for analysis,
```
CalCraven, line 201:
```suggestion
    can be used as a starting point for a HOOMD-blue simulation, for analysis,
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
     Return
     ------
     hoomd_snapshot : hoomd.Snapshot
-        Converted hoomd Snapshot.
+        Converted hoomd Snapshot. Always retruned.
     base_units : dict
-        Based units dictionary utilized during the conversion.
+        Based units dictionary utilized during the conversion. Always returned.
+    rigid_info : hoomd.md.constrain.Rigid
+        Hoomd object storing constituent particle information
```
CalCraven, line 240:
```suggestion
        A hoomd `Constraint` object storing constituent particle information
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
+        needed to run rigid body simulations in Hoomd-Blue.
```
CalCraven, line 241:
```suggestion
        needed to run rigid body simulations in HOOMD-blue.
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
+        This is only returned if Molecule's with `Molecule.isrigid = True`
```
CalCraven, line 242:
```suggestion
        This is only returned if sites with `molecule.isrigid = True`
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
     Notes
     -----
-    Force field parameters are not written to the GSD file and must be included
-    manually in a HOOMD input script. Work on a HOOMD plugin is underway to
-    read force field parameters from a Foyer XML file.
+    Force field parameters are not written to the GSD file and must be
+    generated using `to_hoomd_forcefield()`.
+
+    If you are using mBuild and GMSO to initialize rigid body simulations
+    with Hoomd-Blue set the `Molecule.isrigid` property to `True`.
```
CalCraven, line 251:
```suggestion
    with HOOMD-blue set the `site.molecule.isrigid` property to `True`.
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
+    If your topology contains a mix of rigid and flexible molecules,
+    the rigid molecules must come first in the heirarchy of the GMSO toplogy
```
CalCraven, line 253:
```suggestion
    the rigid molecules must come first in the hierarchy of the GMSO Topology
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
+    and therefore also in the mBuild compound heirarchy.
```
CalCraven, line 254:
```suggestion
    and therefore also in the mBuild compound hierarchy.
```

gmso/external/convert_hoomd.py
```diff
@@ -198,15 +233,46 @@ def to_hoomd_snapshot(
...
+    See https://hoomd-blue.readthedocs.io/en/latest/index.html
+    for more information about running rigid body simulations.
+
+    Rigid Body Example
+    ------------------
+    In this example, a system with rigid benzene and flexible ethane
+    is initialized. Since benzene is rigid, it must be passed first
+    into `mb.fill_box` so that benzene molecules are at the top of
+    `box` heirarchy.
```
CalCraven, line 263:
```suggestion
    `box` hierarchy.
```

gmso/external/convert_hoomd.py
```diff
@@ -95,15 +95,47 @@ def to_gsd_snapshot(
...
+    and therefore also in the mBuild compound heirarchy.
+    See https://hoomd-blue.readthedocs.io/en/latest/index.html
+    for more information about running rigid body simulations in Hoomd-Blue.
+
+    Rigid Body Example
+    ------------------
+    In this example, a system with rigid benzene and flexible ethane
+    is initialized. Since benzene is rigid, it must be passed first
+    into `mb.fill_box` so that benzene molecules are at the top of
+    `box` heirarchy.
```
CalCraven, line 125:
```suggestion
    `box` hierarchy.
```

gmso/external/convert_hoomd.py
```diff
@@ -292,17 +361,93 @@ def _parse_particle_information(
...
             else 1 * base_units["mass"]
-        )
-        charges.append(site.charge if site.charge else 0 * u.elementary_charge)
+            for site in top.sites
+        ]
+    )
+    charges = u.unyt_array(
+        [site.charge if site.charge else 0 * u.elementary_charge for site in top.sites]
+    )
+    moits = [[1, 0, 0] for i in xyz]
```
CalCraven, line 375:
```suggestion
    moment_of_inertias = [[1, 0, 0] for i in xyz]
``` 
Just for clarify on what this variable is,

gmso/utils/geometry.py
```diff
@@ -29,3 +29,34 @@ def coord_shift(xyz, box_lengths):
         xyz += box_max
 
     return xyz
+
+
+def moit(xyz, masses, center=np.zeros(3)):
+    """Find the moment of intertia tensor given a set of
```
CalCraven, line 35:
```suggestion
    """Find the moment of inertia tensor given a set of
```

gmso/utils/geometry.py
```diff
@@ -29,3 +29,34 @@ def coord_shift(xyz, box_lengths):
         xyz += box_max
 
     return xyz
+
+
+def moit(xyz, masses, center=np.zeros(3)):
```
CalCraven, line 34:
```suggestion
def moment_of_inertia(xyz, masses, center=np.zeros(3)):
```

CalCraven: Or something a little more descriptive

gmso/utils/geometry.py
```diff
@@ -29,3 +29,34 @@ def coord_shift(xyz, box_lengths):
...
+    """Find the moment of intertia tensor given a set of
+    particle coordinates and their corresponding masses.
+
+    This method is used in setting rigid body moments
+    of intertia.
```
CalCraven, line 39:
```suggestion
    of inertia.
```

CalCraven (changes requested):
Mostly a bunch of doc string suggestions.

CalCraven:
We should add a test also checking for the error that comes up when an mBuild object is generated with rigid molecules at the back of the compound list, where the snapshot rigid stuff will fail. Maybe we can write a descriptive error for that case to catch.

gmso/external/convert_hoomd.py
chrisjonesBSU: I'm not sure what you mean here. This is part of a Note about how to correctly set up rigid simulations, and shouldn't need to refer to base units at all. I updated the link to go directly to one of the rigid body tutorials in the hoomd docs though.

CalCraven: Well the base_units is one of the main return objects that you'll need to run the simulations no? So just some extra reference in case you want to interpret this object. It's a separate note from the rigid bodies note.

chrisjonesBSU: Good idea. I didn't notice the change was just an addition.

gmso/external/convert_hoomd.py
chrisjonesBSU: I'm not opposed either, other than the fact that it would be a breaking change for any script/program that uses the methods in `convert_hoomd`. I say we save the breaking changes for when we implement a wrapper over these methods, like we've discussed before. We should probably make a PR soon with some deprecation warnings. Maybe this should fall into a larger discussion about a GMSO 1.0 release, similar to what we did with mBuild.

CalCraven: Let's be sure to raise this issue, maybe we could make a new GMSO issue for breaking changes that we should start collating.

CalCraven (approved):
This LGTM! Should be ready to merge

chrisjonesBSU:
I'm just going to add a check for the hierarchy ordering of rigid and non-rigid, then it should be good to go.

PR #851, opened 2024-09-18
Fix `__repr__` method in `abstract_site.Molecule` and `abstract_site.Residue`

chrisjonesBSU:
When poking around some things while working on #850 I got a strange error when trying to look at a `Molecule` instance in a notebook. There is an error in the `__repr__` method of both the `Molecule` and `Residue` classes in `abstract_site.py` where it tries to call an attribute that doesn't exist.

This PR fixes that, and adds some extremely simple tests that should have caught this.

Also, I added fields for Residue and Molecule to the `__repr__` method for `Site` as I think it provides helpful info.

```
>>> from gmso.abc.abstract_site import Molecule
>>> mol = Molecule()
>>> mol

  File "/home/chris/mambaforge/envs/flowermd-dev/lib/python3.11/site-packages/gmso/abc/abstract_site.py", line 27, in __repr__
    f"Molecule(name={self.name}, residue={self.residue}, isrigid={self.isrigid}"
                                          ^^^^^^^^^^^^
  File "/home/chris/mambaforge/envs/flowermd-dev/lib/python3.11/site-packages/pydantic/main.py", line 828, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Molecule' object has no attribute 'residue'
```

### DIFF
--- a/gmso/abc/abstract_site.py
+++ b/gmso/abc/abstract_site.py
@@ -24,7 +24,7 @@ PositionType = Union[Sequence[float], np.ndarray, u.unyt_array]
 class Molecule(GMSOBase):
     def __repr__(self):
         return (
-            f"Molecule(name={self.name}, residue={self.residue}, isrigid={self.isrigid}"
+            f"Molecule(name={self.name}, number={self.number}, isrigid={self.isrigid})"
         )
 
     __iterable_attributes__: ClassVar[set] = {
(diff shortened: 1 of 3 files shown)


chrisjonesBSU:
I'm not sure why all of these merge with upstream are showing up as commits here..I'll try and fix that after these recent PRs are done.

CalCraven (approved):
LGTM!

PR #855, opened 2024-10-01
Handlers for template potentials that are different forms of the same equation.

CalCraven:
Add support for modifying template expression via set expression, and testing for handling accepted_potentials as scaled versions of template potentials.
For example:
```
expression1 = "epsilon*((sigma/r)**12 - (sigma/r)**6)" # missing scalar 4
expression2 = "4*epsilon*((sigma/r)**12 - (sigma/r)**6)" # standard lj
```

If expression1 is the atomtype potential, and expression2 is the accepted potential, since it is the same form as the `LennardJonesPotential` in lib/jsons, then this was previously handled okay. 

However, if expression2 was the atomtype potential, and you were trying to write to a format for _accepted_potentials that is expression1, then this was not supported unless we added a .json file for this modified LJ equation. The current PR seeks to allow a list of `accepted_potentials` where the expression can be reset to and therefore the original LJ template form can be used, but then slightly tweaked for each individual writer. Necessary for PR #848.

This PR also adds the ability to automate scalar equation conversions by loading templates from the name, or directly passing a template object. This allows the user to load the template in the writer, modify the equation slightly, then pass that template for conversion. 

Thoughts on ways to regulate the overwriting of the expression? Should we throw errors if the expression cannot be validated with the independent variables? If so, we'll need to explicitly switch the templates over to `_parametric_potentials`, since they are currently treated as `non_parameteric_potentials`

#### TODO:
- [ ] Tests for all new functionality

### DIFF
--- a/gmso/abc/abstract_potential.py
+++ b/gmso/abc/abstract_potential.py
@@ -1,6 +1,5 @@
 """Abstract representation of a Potential object."""
 
-from abc import abstractmethod
 from typing import Any, Dict, Iterator, List
 
 import unyt as u
@@ -159,11 +158,6 @@ class AbstractPotential(GMSOBase):
         if isinstance(v, dict):
             v = PotentialExpression(**v)
         return v
-
-    @abstractmethod
-    def set_expression(self):
-        """Set the functional form of the expression."""
-        raise NotImplementedError
 
     def __setattr__(self, key: Any, value: Any) -> None:
         """Set attributes of the potential."""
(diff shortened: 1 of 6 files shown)


CalCraven:
I think this code is ready for final review @Zeerakkhan47 and @chrisjonesBSU, if you two want to have a look. The failing codecov test is for updating the hoomd version to 4.5 for the impropers testing, which reduces are coverage slightly, but we are okay with that since it's more accurate at checking the correct versions of hoomd.

chrisjonesBSU (approved):
I played around with this a bit with similar examples we talked about on the dev call a couple of weeks ago. Everything seems to work as expected.

LGTM! Thanks for adding this.

PR #861, opened 2024-11-01
Add templates for PR and Issues

CalCraven:
Just a QOL thing for new users, of which we have a few, for interacting with the packages.


CalCraven:
Getting some odd behavior here with codecov project checks, it's comparing against an older commit than main. Need to take a look at that.

PR #862, opened 2024-12-03
Remove all references to mbuild loaders so tests begin to work

CalCraven:


chrisjonesBSU (approved):
LGTM! Thanks

Since mBuild, foyer and GMSO are so interconnected, I wonder if part of the CI workflow of each repo should include 1 bleeding edge test for the other 2 repos. For example, an mBuild PR runs GMSO unit tests and foyer unit tests, a GMSO PR runs mBuild and foyer unit tests, etc..

CalCraven:
Hmmm, I'll have to think about this. Ideally, we're not making many changes that should break the cross-compatibility as those are pretty significant modifications. Now that we have these stable releases, I think we should limit any of these changes to the develop branches as well. In that case, this would just add a lot of extra testing to any patch changes to the main branches. The other concern is that if either package has failing tests due to some package issue, or something else that needs breaks, then both packages will automatically fail. Which might be annoying if we're trying to add features to one package while the other is failing, say with a numpy 2.0 type situation.

chrisjonesBSU:
Those are good points. Also, this kind of testing is something that can just be done manually if we suspect it needs it.

PR #863, opened 2025-01-01
mol2 writer

Zeerakkhan47:
creating a mol 2 writer function using to output following format: https://chemicbook.com/2021/02/20/mol2-file-format-explained-for-beginners-part-2.html


CalCraven:
Thanks for looking into this! You should be able to add tests to this file @Zeerakkhan47. Then the tests will be run automatically when you push any changes to your branch of this PR. It looks like they're not running automatically right now because of the merge conflicts with an old branch. 
https://github.com/mosdef-hub/gmso/blob/main/gmso/tests/test_mol2.py

You can run the tests for now locally using [pytest](https://docs.pytest.org/en/stable/reference/customize.html#command-line-options-and-configuration-file-settings)

PR #865, opened 2025-01-19
Replace `setup.py` with `pyproject.toml`, update actions to latest versions

chrisjonesBSU:
### PR Summary:

Same as the PR made over in mBuild https://github.com/mosdef-hub/mbuild/pull/1217

It's recommended to use `pyproject.toml` instead of `setup.py` and gets rid of the deprecation error when running `pip install .` inside the repo.


chrisjonesBSU:
The failing bleeding edge test is related to t[his issue](https://github.com/mosdef-hub/forcefield-utilities/issues/44). Haven't had a chance to take a look yet.

CalCraven (approved):
This looks ready to merge.

chrisjonesBSU:
Ok, the environment has `lark 1.2.2`, but still gets this attribute error.

With lark 1.2.2, I don't get an attribute error

```
>>> import lark
>>> lark.__version__
'1.2.2'
>>> lark.Lark
<class 'lark.lark.Lark'>
```

CalCraven:
This looks wrong in line 52 under the Bleeding Edge Stage (Check Environment)
`  focefield_utilities        0.3.0+12.ga23b12a  pypi_0                 pypi    `
Just wanted to point that out.

We also are getting [lark 1.2.2](https://github.com/mosdef-hub/gmso/actions/runs/12913350488/job/36010382740?pr=865#step:7:90), but wonder if the issues lies in this odd downgrade of mBuild and Foyer, then reinstalling from source. Essentially, since the packages depend on each other, installing the dependencies of each may cause some self-reference issue. There's some reference to some weird "already existing files" happening [during the linking step](https://github.com/mosdef-hub/gmso/actions/runs/12913350488/job/36010382740?pr=865#step:5:327)

chrisjonesBSU:
> Yeah, I noticed the bleeding edge test isn't using the latest version of mBuild. I wonder if we should have yet another `.yml` file for running bleeding edge tests? That way, it installs everything in one go. We can install mbuild, foyer and forcefield_utitilities via pip in the yml file. This might be a more manageable and accurate way to do bleeding edge tests rather than running `mamba update` 3 different times.

I added this out of curiosity, we can still talk about it and switch it back if we decide not to go this route.

CalCraven:
Current bleeding edge test gets Foyer 1.0.0, mBuild 1.0.0, and GMSO 0.0.0. 

Looks good, other than the pip install for GMSO now is naming the wheel with the wrong version.

chrisjonesBSU:
> Looks like it's working as expected now. I'd say in general this is better practice since it will decrease confusion when the environments get updated.

Agreed. Should we have a new `.yml` file for this, as it is now, or should this new one become `environment-dev.yml`? This is basically a mosdef dev environment, rather than GMSO specific one. Whatever we decide, I'll make the same change in current mbuild PR, then in foyer in a new PR.

CalCraven:
I think your point is reasonable, this can essentially be a single dev environment. If you're going to be developing the packages, there aren't that many extra dependencies across all of them.

chrisjonesBSU:
> Current bleeding edge test gets Foyer 1.0.0, mBuild 1.0.0, and GMSO 0.0.0.
> 
> Looks good, other than the pip install for GMSO now is naming the wheel with the wrong version.

This should be fixed now.

chrisjonesBSU:
Ok, so just to clarify, gmso, foyer and mbuild will essentially all have the same `environment-dev.yml` file, but each one will install the other 2 packages from source instead of from conda. This file can be used to run bleeding edge tests instead of cloning, updating, and installing from source in the `CI.yaml` workflows for each repoi.

CalCraven:
I'm still confused as to what this package is. It looks like a typo somewhere for `forcefield_utilities`.
https://github.com/mosdef-hub/gmso/actions/runs/12920432658/job/36032738911?pr=865#step:5:69

CalCraven:
Found the [culprit](https://github.com/mosdef-hub/forcefield-utilities/blob/a23b12aa4c300ad93e1423786b3550870084b0b5/pyproject.toml#L6).

chrisjonesBSU:
> Found the [culprit](https://github.com/mosdef-hub/forcefield-utilities/blob/a23b12aa4c300ad93e1423786b3550870084b0b5/pyproject.toml#L6).

I'll make a PR to fix that. I  need to update the versioning in the .toml file over there as well.

PR #872, opened 2025-02-21
Add settles_tag

CalCraven:
Users may or may not have water in their system, and want to use the special rigid treatment in GROMACS called settles. However, being required to call the molecule "water", meant that the GROMACS residue file also had to have residues name "water", but commonly, water is name "WAT" in many practical cases. 

### PR Summary:

This PR, aims to make an argument to `write_top`, that allows users to purposefully set what residues/molecules they want to match for defining the settles section in a GROMACS Topology file.

### PR Checklist
------------

### DIFF
--- a/gmso/formats/top.py
+++ b/gmso/formats/top.py
@@ -24,7 +24,7 @@ from gmso.utils.connectivity import generate_pairs_lists
 
 
 @saves_as(".top")
-def write_top(top, filename, top_vars=None):
+def write_top(top, filename, top_vars=None, settles_tag=None):
     """Write a gmso.core.Topology object to a GROMACS topology (.TOP) file.
 
     Parameters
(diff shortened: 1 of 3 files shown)


gmso/formats/top.py
```diff
@@ -151,11 +161,10 @@ def write_top(top, filename, top_vars=None):
...
                     )
 
-            # Special treatment for water, may ned to consider a better way to tag rigid water
+            # Special treatment for water, may ned to consider a better way to tag rigid SETTLES for 3-site water
+            # need a more general way to work for 4-site and 5-site models.
             # Built using this https://github.com/gromacs/gromacs/blob/main/share/top/oplsaa.ff/spce.itp as reference
-            if "water" in tag.lower() and all(
-                site.molecule.isrigid for site in unique_molecules[tag]["sites"]
-            ):
+            if settles_tag and settles_tag.lower() == tag.lower():
```
chrisjonesBSU, line 167: Before, we checked for the tag name, and `site.molecule.isrigid`. Do we not need to check `site.molecule.isrigid` anymore with these changes?

CalCraven: Good question. Per @dyukovsm and my discussion, we think isrigid is different than SETTLES. Any molecule can be treated with isrigid. However, SETTLES is exclusive to just water. So we think the manual check for SETTLES can be explicitly for 3-site water, whereas we probably can hope to incorporate a more general treatment if a molecule has `isrigid()`, which would be entirely different from the SETTLES treatment.

chrisjonesBSU (approved):
LGTM!

PR #879, opened 2025-03-22
Drop support for hoomd 3, include Hoomd 5 in environment

chrisjonesBSU:
### PR Summary:

Hoomd-Blue 5 was released a couple of months ago. We have a few checks in `convert_hoomd.py` that juggle between functionality of Hoomd 3 and 4. I think it's time to drop support for Hoomd 3. As far as the writers in `convert_hoomd` are concerned, there is really no difference between all versions of Hoomd >= 4.5, so supporting both versions is easy enough, but we'll still have one conditional in place for Hoomd 4 - 4.4 for the time being.

### PR Checklist
------------

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -13,7 +13,7 @@ import unyt as u
 from unyt.array import allclose_units
 
 from gmso.core.views import PotentialFilters
-from gmso.exceptions import NotYetImplementedWarning
+from gmso.exceptions import EngineIncompatibilityError, NotYetImplementedWarning
 from gmso.lib.potential_templates import PotentialTemplateLibrary
 from gmso.utils.connectivity import generate_pairs_lists
 from gmso.utils.conversions import convert_ryckaert_to_opls
@@ -106,7 +106,9 @@ def to_gsd_snapshot(
     read force field parameters from a Foyer XML file.
     """
     if int(hoomd_version[0]) < 4:
-        raise RuntimeError("GMSO is only compatible with Hoomd-Blue >= 4.0")
+        raise EngineIncompatibilityError(
+            "GMSO is only compatible with Hoomd-Blue >= 4.0"
+        )
     base_units = _validate_base_units(base_units, top, auto_scale)
     gsd_snapshot = gsd.hoomd.Frame()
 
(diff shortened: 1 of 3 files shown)


gmso/external/convert_hoomd.py
```diff
@@ -57,7 +57,7 @@ def to_gsd_snapshot(
     parse_special_pairs=True,
     auto_scale=False,
 ):
-    """Create a gsd.snapshot objcet (HOOMD v3 default data format).
+    """Create a gsd.snapshot objcet (HOOMD default data format).
```
CalCraven, line 60:
```suggestion
    """Create a gsd.snapshot object (HOOMD default data format).
```

gmso/external/convert_hoomd.py
```diff
@@ -156,7 +156,7 @@ def to_hoomd_snapshot(
...
```
CalCraven, line 159:
```suggestion
    """Create a gsd.snapshot object (HOOMD default data format).
```

gmso/external/convert_hoomd.py
```diff
@@ -1107,18 +1107,13 @@ def _parse_dihedral_forces(
...
+    if float(f"{hoomd_version[0]}.{hoomd_version[1]}") >= 4.5:
         dtype_group_map["PeriodicTorsionPotential"] = {
             "container": hoomd.md.dihedral.Periodic,
             "parser": _parse_periodic_dihedral,
         }
 
-    else:
-        v_hoomd = "lt3.8"
-        # Should this be periodic, deprecated starting from 3.8.0
+    else:  # Hoomd < 4.5
```
CalCraven, line 1116: Do we want to raise a GMSOError here if HOOMD < 4.0? Since essentially, as it's not a hard dependency, a user could install GMSO with HOOMD 3.0 since it's not in environment.yml, there wouldn't be any compatibility errors.

CalCraven (commented):
This looks like everything is passing. Just a minor comment about testing.

gmso/external/convert_hoomd.py
chrisjonesBSU: Yeah, I agree. I put an error message at the beginning of `to_gsd_snapsot`, `to_hoomd_snapshot` and `to_hoom_forcefield`

PR #881, opened 2025-03-27
Sympy Numpy expressions

CalCraven:
We have been running into the need to support more complex functional forms that have numpy matrix expressions, in them. To validate the use of this, I added a test forcefield and some more tests to parse that these are handled as expected.

### PR Summary:
This doesn't change any functionality, it just verifies that sympy will allow us to place these types of functions in the expression for a forcefield.

To note, we can use lambdify to actually evaluate these functions with numpy arrays, in case we need to do any of this during the writing steps or to evaluate the output of the function. Some example code is shown below.
```python
import numpy as np
from sympy import symbols, sympify
from sympy.utilities.lambdify import lambdify

# Define the expression
expression = "ri + sin(b)*(rj-ri+a*(rk-rj)) / norm(b)"

# Define the custom functions with their corresponding numpy functions
custom_functions = {'sin': np.sin, "norm": np.linalg.norm}

# Define the variables as symbols
ri, rj, rk, a, b = symbols('ri rj rk a b')

# Lambdify the expression to convert it into a callable function
# that evaluates the expression for given values for ri, rj, rk, a, and b
func = lambdify([ri, rj, rk, a, b], expression, modules={'numpy': np, **custom_functions})

# Now func can be called with NumPy arrays or any other compatible types
# For example, to evaluate the function for specific values
ri_value = np.array([1, 0, 0])
rj_value = np.array([0, 1, 0])
rk_value = np.array([0, 0, 1])
a_value = 1
b_value = np.pi / 4  # 45 degrees in radians.

# Call the function with the specified values
result = func(ri_value, rj_value, rk_value, a_value, b_value)
print(result)
```

### PR Checklist
------------

### DIFF
--- a/gmso/tests/files/ff-example2.xml
+++ b/gmso/tests/files/ff-example2.xml
@@ -1,0 +1,98 @@
+<ForceField version="0.0.2" name="NumpyExpressions">
+    <FFMetaData electrostatics14Scale="0.5" nonBonded14Scale="0.67" combiningRule="lorentz">
+        <Units energy="kb" distance="nm" mass="amu" charge="coulomb"/>
+    </FFMetaData>
+    <AtomTypes expression="(A*cos(-B/r) - C/r**6)+norm(C)">
+        <ParametersUnitDef parameter="A" unit="kcal/mol"/>
+        <ParametersUnitDef parameter="B" unit="nm"/>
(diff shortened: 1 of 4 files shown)


chrisjonesBSU (approved):
LGTM!

PR #895, opened 2025-06-03
Fix invalid options in the  docstrings for `ignore_params` in `apply()`

chrisjonesBSU:
### PR Summary:

The doc strings say options for `ignore_params` are bonds, angles, dihedrals, impropers, but this only works if they are in their singular form (bond, angle, dihedral, improper). This PR updates the doc strings to match actual behavior. We could also make it so either singular or plural forms work.

### PR Checklist
------------

### DIFF
--- a/gmso/parameterization/parameterize.py
+++ b/gmso/parameterization/parameterize.py
@@ -73,7 +73,7 @@ def apply(
         this should be changed to False if further modification of expressions are
         necessary post parameterization.
     """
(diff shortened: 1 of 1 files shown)


CalCraven (commented):
Good idea. Let's keep the behavior the same as it was in Foyer, so let's leave the plural form as the doc string, and accept either in the code.

Add to parameterize.py
```python
        ignore_params = set([option.lower().rstrip('s') for option in ignore_params])
```

PR #904, opened 2025-07-09
GMSOLogger Class

CalCraven:
Replace all default warnings with logger.info messages, and create GM…SOLogger class in __init__.py to handle Library logging

### PR Summary:
This PR hopes to readjust the warnings that are overabundant in the package. Notably, warnings such as repeated missing parameters warnings that crop up when loading default files such as the OPLS-AA forcefield, and other unyt based warnings.

### Proposed USAGE
```python
import gmso
# Load logger and modify library logging level such that info is spit out.
gmso_logger = gmso.GMSOLogger()
gmso_logger.library_logger.setLevel(logging.INFO) # defaults to logging.WARNING
top = gmso.Topology()
ff = gmso.ForceField("oplsaa")
```

### PR Checklist
------------
 - [ ] Can write info, debug, warning, error messages to gmso.log file
 - [x] Specific warning messages are classified correctly.
 - [x] Default formatting is set for printing out messages
 - [x] Repeated info is deduplicated
 - [x] All warning.warn messages are replaced with logger messages.


chrisjonesBSU (approved):
LGTM, this is makes the warnings so much easier to read!

I wonder if we should set a default logger instance in `__init__.py`?

Adding this from your example works:

```
gmso_logger = GMSOLogger()
gmso_logger.library_logger.setLevel(logging.INFO)
```

chrisjonesBSU:
I think we need to reconfigure any unit test that checks that a warning is raised, since we removed `warnigns.warn`.

Something like:

```
    with caplog.at_level(logging.WARNING):
        my_func()
    assert "warning message" in caplog.text
```

CalCraven:
> I think we need to reconfigure any unit test that checks that a warning is raised, since we removed `warnigns.warn`.
> 
> Something like:
> 
> ```
>     with caplog.at_level(logging.WARNING):
>         my_func()
>     assert "warning message" in caplog.text
> ```

Complete in ec7442c

gmso/external/convert_hoomd.py
```diff
@@ -1507,11 +1510,11 @@ def _parse_periodic_improper(
 def _validate_base_units(base_units, top, auto_scale, potential_types=None):
     """Validate the provided base units, infer units (based on top's positions and masses) if none is provided."""
     if base_units and auto_scale:
-        warnings.warn(
+        logger.info(
```
chrisjonesBSU, line 1513: We might consider making this one a warning. Really, I wonder if this should raise an error, but at the least it would be good to keepthe warning. Also, this isn't a warning that will come up tons of times.

CalCraven: Changed to error in 970afa9

gmso/core/topology.py
```diff
@@ -1374,9 +1376,16 @@ def get_forcefield(self):
             }
 
             for connection_type in self.connection_types:
+                # ff_conn_types[type(connection_type)][
```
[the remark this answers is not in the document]
CalCraven: Deleted in 970afa9

CalCraven:
Should be good, the main ways to modify your output level are:
```python
gmso_logger.print_level("info") # set what goes to sys.stdout
gmso_logger.debug_file("out.log") # write debug messages to a named file
```

PR #912, opened 2025-08-08
Update oplsaa.xml file in utils, fix failing CI.

chrisjonesBSU:
### PR Summary:

This PR updates the xml file in `gmso/utils/files/gmso_xmls/test_ffstyles` to match the latest XML in foyer after https://github.com/mosdef-hub/foyer/pull/629 and https://github.com/mosdef-hub/foyer/pull/625.

I also set `parametrize=False` in the parmed load method (which matches what we do in foyer's testing). I think some of the .itp files used for testing are missing parameters in the bonds/angles/dihedrals sections. They can still be used for testing atom typing.

The tests here should pass once https://github.com/mosdef-hub/foyer/pull/629 is merged. 

### PR Checklist
------------


CalCraven (approved):
Probably good for consistency, but I think the parametrize=False shouldn't matter if the .top files have the same atomtypes as what's in Foyer. Either it's grabbing the atomtypes from oplsaa.xml or from the top file, which had atomtypes generated from oplsaa.xml. Just wanted to note that behavior.

CalCraven:
Ah, okay I see the issue. The new top files are taken from external sources, which do not have parameters (those are actually stored in the ITP files I think). So that causes the error when you try to use the essentially empty top parameters. The old top files all had parameters, probably because they were written out from ParmEd with their parameters.

chrisjonesBSU:
> The old top files all had parameters, probably because they were written out from ParmEd with their parameters.

Since we added new params, and all these "new" molecules pass testing, it might be worth it to regenerate their .top files with bonds, angles and dihedrals..It can go somewhere on the to-do list.

PR #928, opened 2025-09-24
Bonds and Angles handle wildcards for bonds and sites.

CalCraven:
### PR Summary:
This PR should allow for matching of forcefield connection parameters for Bonds and Angles. This was already usable with Dihedrals, the functionality has just been extended to dihedrals using the "*" syntax. 

Additionally, the ability to match connections by the bond_order of the internal bonds of the connection is also included, using the "~" syntax to designate wildcard bonds, and the SMILES tokens for single, double, triple, and aromatic bonds.

### PR Checklist
------------

### DIFF
--- a/gmso/core/forcefield.py
+++ b/gmso/core/forcefield.py
@@ -26,7 +26,11 @@ from gmso.utils.ff_utils import (
     parse_ff_virtual_types,
     validate,
 )
-from gmso.utils.misc import reverse_string_identifier, validate_type
+from gmso.utils.misc import (
+    improper_equivalents_string_identifier,
+    reverse_string_identifier,
+    validate_type,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -448,9 +452,10 @@ class ForceField(object):
 
     def _get_improper_type(self, identifier, exact_match=False):
         """Get a particular improper_type between `identifier` from this ForceField."""
+        equiv_idx = [(0, i, j, k) for (i, j, k) in itertools.permutations((1, 2, 3), 3)]
         if isinstance(identifier, str):
             forward = identifier
-            reverse = reverse_string_identifier(forward, is_improper=True)
+            equivalent = improper_equivalents_string_identifier(identifier)
         else:
             if len(identifier) == 4:  # add wildcard bonds
                 identifier.append("~")
@@ -463,9 +468,6 @@ class ForceField(object):
                 )
 
             forward = connection_identifier_to_string(identifier)
-            equiv_idx = [
-                (0, i, j, k) for (i, j, k) in itertools.permutations((1, 2, 3), 3)
-            ]
             equivalent = [
                 [
                     identifier[m],
--- a/gmso/utils/misc.py
+++ b/gmso/utils/misc.py
@@ -91,11 +91,21 @@ def get_xml_representation(value):
         return str(value)
 
 
-def reverse_string_identifier(identifier: str, is_improper=False):
+def reverse_string_identifier(identifier: str):
     """Change string identifier for a forcefield key."""
     tokens = r"([\=\~\-\#\:])"
     items = re.split(tokens, identifier)
-    if is_improper:  # only reverse middle two tokens and keep bonds
-        return "".join((items[:1] + items[3:5] + items[1:3] + items[5:]))
-    else:  # flip full
-        return "".join(items[::-1])
+    return "".join(items[::-1])
+
+
+def improper_equivalents_string_identifier(identifier: str):
+    # only reverse middle two tokens and keep bonds
+    tokens = r"([\=\~\-\#\:])"
+    items = re.split(tokens, identifier)
+    return [
+        "".join((items[:1] + items[3:5] + items[1:3] + items[5:])),
+        "".join((items[:1] + items[3:5] + items[1:3] + items[5:])),
+        "".join((items[:1] + items[5:] + items[3:5] + items[1:3])),
+        "".join((items[:1] + items[5:] + items[1:3] + items[3:5])),
+        "".join((items[:1] + items[1:3] + items[5:] + items[3:5])),
+    ]


gmso/core/angle.py
```diff
@@ -26,12 +27,24 @@ class Angle(Connection):
 
     __members_creator__: ClassVar[Callable] = Atom.model_validate
 
+    connectivity: ClassVar[Tuple[Tuple[int]]] = ((0, 1), (1, 2))
```
chrisjonesBSU, line 30: We should add a description here.

chrisjonesBSU: Also, do we need an alias?

gmso/core/angle.py
```diff
@@ -73,6 +87,24 @@ def restraint(self):
...
         return self.__dict__.get("restraint_")
 
+    @property
+    def bonds(self):
+        """Return the bond_order symbol of this bond."""
+        return self.__dict__.get("bonds_")
+
+    @bonds.setter
+    def bonds(self, bonds):
+        """Return the bonds that makeup this Improper.
```
chrisjonesBSU, line 97: I think these doc strings aren't correct for this setter method?

gmso/core/angle.py
```diff
@@ -99,3 +131,14 @@ def __setattr__(self, key, value):
             super(Angle, self).__setattr__("angle_type", value)
         else:
             super(Angle, self).__setattr__(key, value)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_dependent_value_default(cls, data):
```
chrisjonesBSU, line 137: I think a concise doc string for this method would be helpful.

gmso/core/bond.py
```diff
@@ -26,6 +26,8 @@ class Bond(Connection):
 
     __members_creator__: ClassVar[Callable] = Atom.model_validate
 
+    connectivity: ClassVar[Tuple[Tuple[int]]] = ((0, 1),)
```
chrisjonesBSU, line 29: Same as in `angle.py`. Do we need a description and alias here?

gmso/core/angle.py
```diff
@@ -73,6 +87,24 @@ def restraint(self):
         """Return the restraint of this angle."""
         return self.__dict__.get("restraint_")
 
+    @property
+    def bonds(self):
+        """Return the bond_order symbol of this bond."""
```
chrisjonesBSU, line 92: Update the doc string to say that it returns a tuple of `gmso.core.Bond` instances that make up this angle.

gmso/core/bond.py
chrisjonesBSU, line 1: We should probably have a `bond_order` setter in `Bond`

gmso/core/dihedral.py
```diff
@@ -28,12 +29,24 @@ class Dihedral(Connection):
 
     __members_creator__: ClassVar[Callable] = Atom.model_validate
 
+    connectivity: ClassVar[Tuple[Tuple[int]]] = ((0, 1), (1, 2), (2, 3))
```
chrisjonesBSU, line 32: Same, does this need a description and alias?

gmso/core/dihedral.py
```diff
@@ -74,6 +88,27 @@ def restraint(self):
...
+    def bonds(self):
+        """Return the bonds that makeup this dihedral.
+
+        Connectivity is ((0,1), (1,2), (2,3))
+        """
+        return self.__dict__.get("bonds_")
+
+    @bonds.setter
+    def bonds(self, bonds):
+        """Return the bonds that makeup this Improper.
```
chrisjonesBSU, line 101: Update doc strings here.

gmso/core/improper.py
```diff
@@ -94,3 +137,15 @@ def __setattr__(self, key, value):
             super(Improper, self).__setattr__("improper_type_", value)
         else:
             super(Improper, self).__setattr__(key, value)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_dependent_value_default(cls, data):
```
chrisjonesBSU, line 143: I think a concise doc string for this method would be helpful.

gmso/core/dihedral.py
```diff
@@ -99,3 +134,15 @@ def __setattr__(self, key, value):
             super(Dihedral, self).__setattr__("dihedral_type_", value)
         else:
             super(Dihedral, self).__setattr__(key, value)
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_dependent_value_default(cls, data):
```
chrisjonesBSU, line 140: I think a concise doc string for this method would be helpful.

gmso/core/improper.py
```diff
@@ -63,6 +85,27 @@ def connection_type(self):
...
+    def bonds(self):
+        """Return the bonds that makeup this Improper.
+
+        Connectivity is ((0,1), (0,2), (0,3))
+        """
+        return self.__dict__.get("bonds_")
+
+    @bonds.setter
+    def bonds(self, new_bonds):
+        """Return the bonds that makeup this Improper.
```
chrisjonesBSU, line 98: Update the doc strings here.

gmso/utils/connectivity.py
```diff
@@ -452,3 +457,13 @@ def _graph_from_vtype(vtype):
         virtual_type_graph.add_edge(i, i + 1)
 
     return virtual_type_graph
+
+
+def create_pattern(combination):
```
chrisjonesBSU, line 462: A quick doc string here would be helpful.

chrisjonesBSU (changes requested):
I left some comments, mostly related to doc strings. So far this is looking great. I'll try and get some minimal working examples going to further test it out.

gmso/core/angle.py
CalCraven: I specified connectivity as a ClassVar, so it's not modifiable across classes, and doesn't get treated the same way as other pydantic variables, which are defined under with pydantic Field.

CalCraven: So also no alias here.

gmso/core/dihedral.py
CalCraven: See above.

gmso/core/improper.py
CalCraven: Done.

gmso/core/angle.py
CalCraven: Done.

gmso/core/angle.py
CalCraven: remove setter.

PR #934, opened 2025-10-23
Fill out virtual_sites position functionality

CalCraven:
### PR Checklist
------------

### DIFF
--- a/gmso/tests/test_expression.py
+++ b/gmso/tests/test_expression.py
@@ -355,7 +355,6 @@ class TestExpression(BaseTest):
             parameters={"a": [3, 4, 0] * u.nm},  # norm is 5
             independent_variables="r",
         )
-        r = sympy.symbols("r")
         x, y, z = sympy.symbols("x y z")
         independent_namespace["r"] = sympy.Matrix([x, y, z])
         independent_parameters = {"x": 0 * u.nm, "y": 1 * u.nm, "z": 2 * u.nm}


gmso/core/virtual_site.py
```diff
@@ -1,11 +1,28 @@
...
 from pydantic import ConfigDict, Field
+from sympy import Function, Matrix, symbols
 
 from gmso.abc.abstract_site import Site
 from gmso.core.virtual_type import VirtualType
-from gmso.exceptions import MissingPotentialError, NotYetImplementedWarning
+from gmso.exceptions import MissingPotentialError
+
+
+class norm(Function):
```
chrisjonesBSU: This is defined here but also in `expression.py`.

gmso/core/virtual_site.py
```diff
@@ -68,10 +85,23 @@ def position(self) -> str:
...
+        for symbol, site in zip(string.ascii_lowercase[8:], self.parent_sites):
+            for i, pos in enumerate(site.position):
+                independent_parameters[f"r{symbol}{i + 1}"] = float(pos.value)
+
+        # perform expression evaluation
+        return (
+            self.virtual_type.virtual_position.potential_expression.evaluate(
+                independent_namespace, independent_parameters
+            ).T
+            * u.nm
```
chrisjonesBSU: We should get the units from the parent site positions and use that here instead of hard-coding nanometers. Or maybe in this case it would come from the units in the virtual site type.

chrisjonesBSU (changes requested):

gmso/core/virtual_site.py
CalCraven: Completed in 7f936662

gmso/core/virtual_site.py
```diff
@@ -1,11 +1,28 @@
...
+class norm(Function):
+    """Sympy functions for use in lambdify"""
+
+    @classmethod
+    def eval(cls, arg):
+        return None
+
+
+# Evaluate vector norm
+def norm_evaluation(matrix_arg):
```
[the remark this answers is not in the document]
CalCraven: Fixed in 7f936662

chrisjonesBSU (approved):

PR #948, opened 2025-12-23
Add python 3.14 to testing, drop python 3.10

chrisjonesBSU:
### PR Summary:

Just wanted to get this going to see where dependencies and testing passes or fails for python 3.14

### PR Checklist
------------


CalCraven:
Looks like parmed isn't updated to 3.14

chrisjonesBSU:
The environment builds now with python 3.14, but there are a couple tests failing from more recent versions of OpenMM (I think), debugging that now.

chrisjonesBSU:
The issue lives in foyer. OpenMM 8.4 added a new `setAtomClasses` method and foyer's own implementation of the `createSystem` method (overriding the same method in OpenMM) doesn't call it, so the atom classes list is empty, and there is an index error. I'll make a PR in foyer that fixes this, bumps python to 3.14 and numpy up to <2.6, so it is consistent with this environment.

PR #954, opened 2026-01-29
Add ndecimals to xyz file for higher precision

CalCraven:
### PR Summary:
This just adds an optional argument to add more precision to the xyz writer, which is important for some of the packing.py features in mBuild that interface with packmol via xyz files.

### PR Checklist
------------


chrisjonesBSU (approved):
LGTM! Does this approach add zeros if decimals is larger than what's available in the particle positions?

PR #957, opened 2026-02-05
Building forcefield XMLs and JSON templates for CG simulations

StephMcCallum:
### PR Summary:

[DRAFT] This PR is motivated by work in the MuPT project. The goal is to enable force-field parameterization of coarse-grained systems for LAMMPS and HOOMD simulation engines. For each simulation engine, I have included two models: Kremer-Grest polymers with FENE bonds and LJ potential, and also a bead-spring model with LJ potential and harmonic bonds, angles, and dihedrals.

### PR Checklist
------------

### DIFF
--- a/gmso/lib/jsons/LAMMPSFENEBondPotential.json
+++ b/gmso/lib/jsons/LAMMPSFENEBondPotential.json
@@ -1,0 +1,11 @@
+{
+  "name": "LAMMPSFENEBondPotential",
+  "expression": "-0.5 * K * R0**2 * log(1-(r/R0)**2) + 4 * epsilon * ((sigma/r)**12 - (sigma/r)**6) + epsilon",
+  "independent_variables": "r",
+  "expected_parameters_dimensions": {
+    "K": "energy/length**2",
+    "R0": "length",
+    "epsilon": "energy",
+    "sigma": "length"
+  }
+}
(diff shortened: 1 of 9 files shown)


gmso/lib/jsons/LAMMPSFENEBondPotential.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "LAMMPSFENEBondPotential",
```
CalCraven: I put some LAMMPS support in #958, However I just used a FENE potential with no sigma and epsilon parameters. Is it possible to set sigma and epsilon as a typical pair interaction, and then use the FENE bond potential as one with just K and R0 params? Do we want both potentials to be handled in the writers? You should be able to review the code I pushed in that PR to see where to store the LAMMPS functions we would use to parse the different potentials from our templates. I also noticed in lammps-lj-fene.xml you set both a pair potential and the sigma and epsilon interaction in the FENE potential. Should both of those sections then be written to the LAMMPS input file?

I'll sneak in a few changes to the hoomd writer as well into that PR so you can test using some HOOMD simulations. Let me know what isn't working if you do get to run some test simulations.

chrisjonesBSU (approved):
LGTM we'll be able to test these out in the PR Cal just opened for the writers.

PR #958, opened 2026-02-06
Add FENEBondPotential and writers to gromacs, hoomd and lammps

CalCraven:
### PR Checklist
------------
 - [x] Added to LAMMOPS
 - [x] Added to GROMACS
 - [x] Added to Hoomd

### DIFF
--- a/gmso/formats/lammpsdata.py
+++ b/gmso/formats/lammpsdata.py
@@ -260,8 +260,8 @@ def read_lammpsdata(
     Currently supporting the following unit styles: 'real', "real", "lj", "metal", "si", "cgs",
     "electron", "micro", "nano".
 
-    Currently supporting the following potential styles: 'lj'
-    Currently supporting the following bond styles: 'harmonic'
+    Currently supporting the following pair potential styles: 'lj'
+    Currently supporting the following bond styles: 'harmonic', 'fene'
     Currently supporting the following angle styles: 'harmonic'
     Currently supporting the following dihedral styles: 'opls'
     Currently supporting the following improper styles: 'harmonic'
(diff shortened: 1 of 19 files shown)


gmso/formats/lammpsdata.py
```diff
@@ -608,13 +608,15 @@ def _accepted_potentials():
     templates = PotentialTemplateLibrary()
     lennard_jones_potential = templates["LennardJonesPotential"]
     harmonic_bond_potential = templates["LAMMPSHarmonicBondPotential"]
+    fene_bond_potential = templates["FENEBondPotential"]
```
chrisjonesBSU: Should this be using the template `LAMMPSFENEBondPotential.json` instead?

gmso/formats/lammpsdata.py
```diff
@@ -860,6 +871,40 @@ def _write_bondtypes(out_file, top, base_unyts, cfactorsDict):
     return bond_types
 
 
+def _write_fene_bonds(out_file, top, first_bondtype, base_unyts, cfactorsDict):
+    """Setup `bond_style fene` Bond Type with 0 for sigma and epsilon parameters."""
```
chrisjonesBSU: I'm kind of confused. This should be writing everything needed for the `fene` bond style in lammps, right? In LAMMPS, sigma and epsilon are parameters that need to be set and not always zero. I think HOOMD and LAMMPS both only have FENE+WCA options (where sigma and epsilon are parameters) and GROMACS only does FENE (sigma and epsilon aren't parameters)

chrisjonesBSU (changes requested):
This is looking good, but I left a couple comments. Also, we should probably merge #960 first since it changes some of the parameter values in the templates.

gmso/formats/lammpsdata.py
CalCraven: Yes, I added it before the LAMMPS one was merged to main, good catch.

chrisjonesBSU (approved):

PR #962, opened 2026-02-18
DPD Potential XML and JSON

StephMcCallum:
### PR Summary:
This PR is meant to add the forcefield xml and json template for [HOOMD's DPD potential](https://hoomd-blue.readthedocs.io/en/latest/hoomd/md/pair/dpd.html). The DPD potential is a three part forcefield which includes an interaction potential, pairwise drag force, and pairwise random force. These added files only include the interaction potential. The interaction potential is a pairwise function. I have only included the non-zero case where r < r_cut. 

### PR Checklist
------------

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.14
+  rev: v0.15.2
   hooks:
     # Run the linter.
     - id: ruff
@@ -26,7 +26,7 @@ repos:
       exclude: 'setup.cfg|gmso/tests/files/.*'
 
 - repo: https://github.com/pycqa/isort
-  rev: 7.0.0
+  rev: 8.0.0
   hooks:
     - id: isort
       name: isort (python)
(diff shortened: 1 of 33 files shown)


mrshirts:
@CalCraven let us know if this looks ready, or if there are other things that are pending (like a DPD interaction builder).

chrisjonesBSU:
This looks good to go. We'll still need the corresponding DPD parser in `gmso.convert.convert_hoomd.py` like we added for the [FENE Bonds](https://github.com/mosdef-hub/gmso/blob/debed9d380e2bc1096b57807beff14f3e50c4b6f/gmso/external/convert_hoomd.py#L1214)

Also, @CalCraven we might need to have some discussion about how to handle `r_cut`. In HOOMD, `r_cut` can be set at the atom-type level for many of its pair potentials, BUT we've never treated it like a parameter in the xml file. Instead, its defined once in the hoomd writer, and treated more like a system-wide parameter.  

@StephMcCallum @mrshirts for the DPD potential, is `r_cut` something you all are going to vary with bead-type frequently? 

We could support both approaches, if its defined in the xml, then use that at the atom-type level, if its not, then use the value in the writer as the r_cut for everything, but now might be the time to decide on this. How do we handle an optional parameter in the templates?

CalCraven (approved):
LGTM. We'll have to write a test that actually checks properties of the dpd file

PR #965, opened 2026-03-04
Add pairpotential parsing to hoomd, and parse dpd forces

CalCraven:
### PR Checklist
------------


gmso/lib/jsons/HOOMDDPDConservativeForce.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "HOOMDDPDConservativeForce",
+  "expression": "A * (1-(r/r_cut)) - γ",
+  "independent_variables": "r",
+  "expected_parameters_dimensions": {
+    "A": "force",
+    "r_cut": "length",
+    "γ": "mass / length / time"
```
StephMcCallum: On this line, why has "/ length" been added? Does this need to be added in other places if it is for unit conversion?

gmso/lib/jsons/HOOMDDPDConservativeForce.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "HOOMDDPDConservativeForce",
+  "expression": "A * (1-(r/r_cut)) - γ",
```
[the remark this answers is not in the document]
CalCraven: Yeah, the reason it's not included is because it has terms which could be expressed in different ways. We have to figure out a reasonable way to express them in the equation. There will probably be just 4 or 5 independent variables. 

This was just my quick and easy way to get it in there so the form of the equation is usable in simulations. The actual expression is only used for converting one equation to the other. What's written to hoomd is just the force objects, so the representation is hoomd's native one. At that stage, the expression we used isn't a applied, which is why this works as a quick and dirty working approach. But like you said, we should probably include the whole dissipative force, for the sake of being clear what is here.

gmso/lib/jsons/HOOMDDPDForce.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "HOOMDDPDConservativeForce",
+  "expression": "A * (1-(r/r_cut)) - γ",
```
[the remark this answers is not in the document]
StephMcCallum: @CalCraven just wanted to follow up on this. How are the pair potential expressions being used in gmso? Based on the explanation above, can variables be added to the pair potential variable list to validate form but not be part of the expression? If we want to have the full form of the equation I can try to add that.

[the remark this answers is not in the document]
CalCraven: >
> How are the pair potential expressions being used in gmso? 


In general, pair potential expressions are used in gmso to write out pair forces to a given simulation engine. In most cases, we have a nonbonded force for something like LJ, where the force can be defined by a single site parameter, and the pair forces in the engine are then calculated based on mixing rules. The pair potential is the stand-in where you need to define both sites that are part of the pair interaction, but functionally they end up being the same.

For specifically the dpd pair force, [here's in this PR where it's being used to write a dpd force in hoomd](https://github.com/mosdef-hub/gmso/pull/965/changes#diff-88c376be9b0fb3ba22b51007b364c4b58603579f3a7cd70b9b91c2a00a4ef9ceR1079). As you can see, the kT, A, and gamma parameters are all being used to create the hoomd object that would be used in the simulation forces.

Now the actual form of the equation doesn't matter, as the only thing that matters is that the parameters are written "correctly" to the engine (in this case hoomd, so I'll stick with that). 
If I have equation 1: `y=mx+b`, and I'm trying to write to an engine that accepts the form of the equation, `y=1/2mx+b`, then I just need to multiple my `m` by 2 and then it'll be the correct value. That's what GMSO writers do, when the form of the equation in your forcefield doesn't match what's needed in the actual engine, it can handle some conversions, with the caveat that many conversions are not possible. So that's why it's important that we write the equation correctly, so those conversions can be made possible. However, the dpd force is somewhat complicated, as the actual terms are reported mostly as functionals of the parameters, and summed terms, since the actual mathematics are vectorized.


> If we want to have the full form of the equation I can try to add that.


Right now, there is no viable conversions besides unit conversions. Sot it's not an extremely urgent matter I would say. If you would like to write a form of the equation that includes the vecotrized math, we can absolutely handle that. See the position evaluation in this [PR for virtual sites](https://github.com/mosdef-hub/gmso/pull/934) and how the [virtual_site position potential](https://github.com/mosdef-hub/gmso/blob/38cb9d235ad03c25c2334fd1def3115fe7b28ade/gmso/utils/files/gmso_xmls/test_ffstyles/tip4p_2005.xml#L43) we recently implemented.

We can probably do this in a new PR, probably good to get this merged if it's in minimum viable working conditions.

gmso/lib/jsons/HOOMDDPDForce.json
```diff
@@ -0,0 +1,10 @@
...
+  "independent_variables": "r",
+  "expected_parameters_dimensions": {
+    "A": "force",
+    "r_cut": "length",
+    "γ": "mass / length / time"
```
StephMcCallum: Just to clarify, the "/ length" is needed here to verify the form with the GMSO library? So we will keep it?

gmso/external/convert_hoomd.py
```diff
@@ -885,6 +891,7 @@ def _validate_compatibility(top):
     harmonic_torsion_potential = templates["HarmonicTorsionPotential"]
     opls_torsion_potential = templates["OPLSTorsionPotential"]
     rb_torsion_potential = templates["RyckaertBellemansTorsionPotential"]
+    dpd_force = templates["HOOMDDPDConservativeForce"]
```
StephMcCallum: Now that we are representing the complete DPD force by adding gamma and kT parameters, I think it would be appropriate to change this and file names to "HOOMDDissipativeParticleDynamics".

StephMcCallum (approved):
This looks good as a minimal implementation of the DPD force from HOOMD. We may need to revisit a more complex expression for the potential in another PR.

gmso/lib/jsons/HOOMDDPDForce.json
CalCraven: Change from:
 `"γ": "mass / length / time"`
to 
`"γ": "mass / time"`

gmso/external/convert_hoomd.py
CalCraven: I think from a GMSO standpoint, that's a little confusing. All of the other templates are labeled `XPotential`. DPD is not a potential, it is a force. So it should be labeled as `XForce`. I thing spelling out DPD is maybe unnecessary as well. Maybe we should change to `"HOOMDDPDForce"`, and drop the "Conservative" portion.

StephMcCallum: That makes sense to me.

chrisjonesBSU: Hoomd has both md.pair.DPD and  md.pair.DPDConservative forces. In this PR, the writer is making a `hoomd.md.pair.DPD`object, so yes we should drop the "Conservative" from the template name, and anywhere else that it shows up. Then in a future PR if we add support for `hoomd.md.DPDConservative` the templates can include it.

Although, I think the only difference between them is that `DPDConservative` doesn't include the drag force (gamma) or the random force (kT), so technically, we could still use one XML template, and if gamma is zero everywhere in the xml, and `kT` parameter isn't defined, the writer creates the `md.pair.DPDConservative`

chrisjonesBSU:
> This looks good as a minimal implementation of the DPD force from HOOMD. We may need to revisit a more complex expression for the potential in another PR.

How are you thinking the expression would change for HOOMD's implementation of DPD and/or DPDConservative?

StephMcCallum:
> > This looks good as a minimal implementation of the DPD force from HOOMD. We may need to revisit a more complex expression for the potential in another PR.
> 
> How are you thinking the expression would change for HOOMD's implementation of DPD and/or DPDConservative?

This was discussed a bit above in a thread that I resolved. 

So I'm looking at the expression for the DPD potential in the XML file. The HOOMD DPD force has three parts, the conservative, random, and dissipative. I agree with the implementation for the conservative expression. 

This PR currently adds a " - gamma" term to the expression. The gamma parameter is part of the dissipative force and random force, but this isn't the complete expression. The dissipative is part of the integrator and depends on the velocity difference vector. 

The random force is missing from this expression. It's included in the writer by adding the kT parameter. The random force depends on a random number generator and also is part of the integrator because this expression depends on dt.

So, I think the correct, full expression for DPD would include these more complex parts for the dissipative and random force (random number generator, dot product, time step, vector calculation). Something like this:  "A * (1-(r/r_cut)) - gamma * (1-(r/r_cut))**2 * np.dot(r_hat, v_ij) - theta*sqrt(3)* (1-(r/r_cut))*  np.sqrt((2*gamma*kT)/dt)" @CalCraven suggested adding this vectorized math in another PR.

However, I don't think it's necessary to implement this equation. It's my understanding thats GMSO is verifying the existence of input parameters and their units before passing them to the HOOMD writer. An open question I have is, How is gmso using this expression? and Will we be incorrectly calculating the forces if the expression is not an exact match?

chrisjonesBSU:
>So, I think the correct, full expression for DPD would include these more complex parts for the dissipative and random force (random number generator, dot product, time step, vector calculation). Something like this: "A * (1-(r/r_cut)) - gamma * (1-(r/r_cut))**2 * np.dot(r_hat, v_ij) - thetasqrt(3) (1-(r/r_cut))* np.sqrt((2gammakT)/dt)" @CalCraven suggested adding this vectorized math in another PR. However, I don't think it's necessary to implement this equation. It's my understanding thats GMSO is verifying the existence of input parameters and their units before passing them to the HOOMD writer. An open question I have is, How is gmso using this expression? and Will we be incorrectly calculating the forces if the expression is not an exact match?

GMSO doesn't calculate forces, but Hoomd does. All we can do on the GMSO side is set values for whatever parameters Hoomd will let us. Which is A, gamma and kT and r cut. So far, we're treating A and gamma as atom type level parameters (expressed for each atom type in the xml file) and kT and r_cut as system-level parameters (set by parameters in the hoomd writer function).

The `v_i`, `v_j`, and `theta` parts of the expression are handled internally by hoomd, and can't be set from GMSO. I'm guessing the random number generator uses the seed you pass to the hoomd simulation class, and `dt` is set when you create the integrator in your hoomd script. I can't see a reason to add the complete expression on the GMSO side other than for documentation purposes, but maybe that's enough reason to add it? Would it technically be correct to list all of these as independent variables in the xml? If we did, I don't think it changes anything about the functionality or implementation of the force.

Really, the expressions are just used to make sure the parameters in the xml define everything that is expected by that potential (but really only in terms of what the simulation engines allow us to define). If a certain potential depends on things like particle velocities, random numbers, time step, these are probably never going to be handled on the writer side of things.

@CalCraven and I have talked a little about adding functionality of an internal energy calculator (like, getting a single energy value for a topology) which would use the expressions from the templates and the parameter values of the typed atom topology. But, that doesn't exist now, and even if it did, we would be limited on what potentials we could do that for. For example, GMSO has no way to set velocities in the topology, so even if the expression has placeholders for `v_i` and `v_j` that information isn't accessible.

gmso/external/convert_hoomd.py
CalCraven: yeah I'd lean towards using one template that works in both cases. So having the more general one that we've added is better.

CalCraven:
> GMSO doesn't calculate forces, but Hoomd does. All we can do on the GMSO side is set values for whatever parameters Hoomd will let us. Which is A, gamma and kT and r cut. So far, we're treating A and gamma as atom type level parameters (expressed for each atom type in the xml file) and kT and r_cut as system-level parameters (set by parameters in the hoomd writer function).

Correct.

> The `v_i`, `v_j`, and `theta` parts of the expression are handled internally by hoomd, and can't be set from GMSO. I'm guessing the random number generator uses the seed you pass to the hoomd simulation class, and `dt` is set when you create the integrator in your hoomd script. I can't see a reason to add the complete expression on the GMSO side other than for documentation purposes, but maybe that's enough reason to add it? Would it technically be correct to list all of these as independent variables in the xml? If we did, I don't think it changes anything about the functionality or implementation of the force.

No, no changes to the functionality other than just complete documentation so people know what actual expression they're using if they create a xml using these params/forces.

> Really, the expressions are just used to make sure the parameters in the xml define everything that is expected by that potential (but really only in terms of what the simulation engines allow us to define). If a certain potential depends on things like particle velocities, random numbers, time step, these are probably never going to be handled on the writer side of things.

Correct

> @CalCraven and I have talked a little about adding functionality of an internal energy calculator (like, getting a single energy value for a topology) which would use the expressions from the templates and the parameter values of the typed atom topology. But, that doesn't exist now, and even if it did, we would be limited on what potentials we could do that for. For example, GMSO has no way to set velocities in the topology, so even if the expression has placeholders for `v_i` and `v_j` that information isn't accessible.

Yeah, this would be a future case which we could handle down the line. We could easily allow v_i and v_j to be sampled from kinetic temperature distribution though if we wanted, but the obvious simple addition would be that that part of the force is just 0 for purposes of use in GMSO.

CalCraven:
@StephMcCallum Once that naming convention is updated and the branch is synced to main, I'll merge this.

StephMcCallum:
Thank you @CalCraven and @chrisjonesBSU for your work on this and your explanations!

PR #970, opened 2026-04-21
Speedup gsd snapshot and forcefield writing for convert_hoomd

CalCraven:
### PR Summary: 
Speedup cataloging of indexing for different objects when writing out to hoomd. Notably, replace all top.index(site) calls with a map of site_indexMap[id(site)] for indexing. Also remove separate iterations through top.sites for isrigid() checks, performing that during the main loop through Topology sites. Also perform some numpy improvements by initializing size arrays, and filling in xyz[i] as we iterate through the Topology.

### PR Checklist
------------

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -1065,7 +1065,7 @@ def _parse_coulombic(
         coulombic = hoomd.md.long_range.pppm.make_pppm_coulomb_forces(
             nlist=nlist, resolution=resolution, order=order, r_cut=r_cut
         )
-    if not np.all(scaling_factors):
+    if not np.any(scaling_factors):
         return [*coulombic]  # return early
     # Handle 1-2, 1-3, and 1-4 scaling
     # TODO: Figure out a more general way to do this and handle molecule scaling factors
@@ -1116,7 +1116,7 @@ def _parse_lj(top, atypes, combining_rule, r_cut, nlist, scaling_factors):
     # Handle 1-2, 1-3, and 1-4 scaling
     # TODO: Figure out a more general way to do this
     # and handle molecule scaling factors
-    if not np.all(scaling_factors):
+    if not np.any(scaling_factors):
         return [lj]
     special_lj = hoomd.md.special_pair.LJ()
 
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -185,6 +185,9 @@ def to_gsd_snapshot(
         "Special pairs are not currently written to GSD files",
     )
 
+    # Create IndexMap
+    site_indexMap = {id(site): i for i, site in enumerate(top.sites)}
+
     n_rigid, rigid_info = _parse_particle_information(
         gsd_snapshot,
         top,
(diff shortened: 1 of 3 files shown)


chrisjonesBSU:
While testing this out, I think I caught a bug from a previous PR that updated how special pairs are handled.

```
import mbuild as mb
import gmso
from gmso.parameterization.parameterize import apply

benzene = mb.load("c1ccccc1", smiles=True)
top = benzene.to_gmso()
oplsaa = gmso.core.forcefield.ForceField("oplsaa")
typed_top = apply(top, oplsaa)

print(typed_top.scaling_factors)

[[0.  0.  0.5]
 [0.  0.  0.5]]
```

So, we have scaling factors of 0.5 for both LJ and Coulomb pairs, but the special pair objects aren't created 

```
forces, _ = gmso.external.to_hoomd_forcefield(typed_top, r_cut=1.4)
print(forces["nonbonded"])

[<hoomd.md.pair.pair.Ewald at 0x314b66810>,
 <hoomd.md.long_range.pppm.Coulomb at 0x31c2e9f70>,
 <hoomd.md.pair.pair.LJ at 0x31c2ea810>]
```

I think the issue is the `not np.all()` used [here](https://github.com/mosdef-hub/gmso/blob/1a6017d2cb5658500f571b6bcea8fa3f2b059db2/gmso/external/convert_hoomd.py#L1019) and [here](https://github.com/mosdef-hub/gmso/blob/1a6017d2cb5658500f571b6bcea8fa3f2b059db2/gmso/external/convert_hoomd.py#L1070)

```
not np.all([[0, 0, 0.5], [0, 0, 0.5]])

True
```

I think it should use `if not np.any(scaling_factors):` instead. This is `True` if the scaling factors are all zeros. 

```
not np.any([[0, 0, 0.0], [0, 0, 0.0]])

True
```

Let me know if I'm missing something, but when there are non-zeros in the top.scaling_factors, we want to write out special pair forces, when it's all zeros, we don't. Right?

chrisjonesBSU:
The unit test wasn't catching this, because it appears manually changing `top.scaling_factors` doesn't work as intended.


```
benzene = mb.load("c1ccccc1", smiles=True)
top = benzene.to_gmso()
oplsaa = gmso.core.forcefield.ForceField("oplsaa")
typed_top = apply(top, oplsaa)
print(typed_top.scaling_factors)

[[0.  0.  0.5]
 [0.  0.  0.5]]


typed_top.scaling_factors[0][2] = 0
typed_top.scaling_factors[1][2] = 0
print(typed_top.scaling_factors)

[[0.  0.  0.5]
 [0.  0.  0.5]]
```

chrisjonesBSU (approved):
LGTM. 

Could there be any reason to have this `site_indexMap` be an attribute of `Topology`? It would have to be re-generated anytime a new site is added, but that isn't a common workflow in GMSO. I'm not saying we should, but it would be worth thinking about if this dict will be useful for other common workflows in GMSO, in other writers, conversions between formats (to_mbuild, to_parmed, etc...)

CalCraven:
> Let me know if I'm missing something, but when there are non-zeros in the top.scaling_factors, we want to write out special pair forces, when it's all zeros, we don't. Right?

Yes that's the intention there. We should make sure to have a test that covers that absolutely. Thanks for looking at that and getting that in!

CalCraven:
Getting a 35% speedup with 80k atoms. This is for an entirely untyped system, I suspect it would be even more for writing out force objects with a typed system or with rigid bodies.

```python
# Time test

import time
import mbuild as mb
import numpy as np

cpd = mb.load("CC", smiles=True)
box= mb.fill_box(cpd, n_compounds=10000, box=mb.Box([50,50,50]))

top = box.to_gmso()
top.save("out.gsd", overwrite=True)
times = np.arange(3)
for i in range(len(times)):
    start = time.perf_counter_ns()
    top.save("out.gsd", overwrite=True)
    times[i] = (time.perf_counter_ns() - start)/1e6

print(times.mean(), times.std())
# np.save("PR.npy", times)
np.save("MAIN.npy", times)

d1 = np.load("PR.npy").mean()
d2 = np.load("MAIN.npy").mean()
print(top.n_sites)

(d1 - d2)/d2

```
```
80000
np.float64(-0.35107410089307267)
```

chrisjonesBSU:
> > Let me know if I'm missing something, but when there are non-zeros in the top.scaling_factors, we want to write out special pair forces, when it's all zeros, we don't. Right?
> 
> Yes that's the intention there. We should make sure to have a test that covers that absolutely. Thanks for looking at that and getting that in!

I updated the previous unit test to make sure this is covered now.

PR #973, opened 2026-05-07
Revamp the documentation build.

chrisjonesBSU:
### PR Summary:

This PR completely redesigns the structure of the documentation and the readthedocs build. It adds an API section that references the doc strings. Type hints are added where possible to help improve readability of the docs. Most of the original information we had in the docs before remains.

@CalCraven do you want to look it over and see if we are missing anything? 

One possible section we could add (probably manually instead of inferred) is listing out all the currently supported potentials. EDIT: This has been added now.

I think the main thing still missing is documentation discussing the GMSO xml file format.

You can click the read the docs build action to see them, but for me it's not showing the changes from the latest commit, so you might have to build them locally.

### PR Checklist
------------


docs/api/parameterization.rst
```diff
@@ -0,0 +1,44 @@
...
+name) and writes atom types, bond types, angle types, dihedral types, and improper types
+onto the topology in-place.
+
+.. autofunction:: gmso.parameterization.parameterize.apply
+
+TopologyParameterizer
+---------------------
+
+:class:`~gmso.parameterization.topology_parameterizer.TopologyParameterizer` drives the
+parameterization workflow.  It is instantiated internally by :func:`apply` but may be used
```
CalCraven: Also missing the function arguments for apply. Apply is the main API point so this isn't a correct statement here.

CalCraven (changes requested):
Can we change from installation.rst:

Supported Python Versions[](https://gmso--973.org.readthedocs.build/en/973/installation.html#supported-python-versions)
Python 3.9-3.11 is the recommend version for users. It is the only version on which development and testing consistently takes place. Older (3.6-3.9) and newer (3.12+) versions of Python 3 are likely to work but no guarantee is made and, in addition, some dependencies may not be available for other versions. No effort is made to support Python 2 because it is considered obsolete as of early 2020.

To be:
3.10-3.13

Also: 

And then run the tests with the py.test executable:

Can be changed to the more modern pytest.

docs/api/formats.rst
```diff
@@ -0,0 +1,165 @@
...
+GROMACS
+-------
+
+GROMACS coordinate files (`.gro`) and topology/parameter files (`.top`, `.itp`) are
+supported for both reading and writing.
+
+.. autosummary::
+   :nosignatures:
+
+   gmso.formats.read_gro
```
CalCraven, line 20: Do we want the arguments supported here? Just the function names are showing for me.

chrisjonesBSU: If you build it locally the methods in the formats section should have the full API. I'm not sure why it isn't showing in the CI build.

docs/api/parameterization.rst
chrisjonesBSU: Check the build locally, just like with the formats.rst methods, all of the parameters and doc strings show up for `apply` as well.

chrisjonesBSU: Also, I removed the two non-public usage classes from this section of thed docs. The avoid confusion, `apply` is the only method that shows up in this section of the docs now.

chrisjonesBSU:
My local build is still different from what is showing up in the CI build for the docs. I'm just going to merge and we'll see what it looks like in the RTD build.

PR #979, opened 2026-06-03
Fix ordering and sorting to be dependent on presence of rigid bodies

chrisjonesBSU:
### PR Summary:

This uses different sorting methods in `convert_hoomd` depending on whether the topology contains rigid bodies. If rigid bodies are present, particle sorting changes, and is based on molecule (parent) grouping. This is required to correctly run rigid body simulations in hoomd. This also fixes an issue where we assumed that all molecules with the same label/name had the same number of sites per molecule.

### PR Checklist
------------

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -438,10 +438,6 @@ def _parse_particle_information(
     else:
         for site in itertools.chain(top.sites, top.virtual_sites):
             if isinstance(site, VirtualSite):
-                logger.warning(
-                    "This topology contains gmso.core.VirtualSite particles, but "
-                    "no rigid bodies were found. See Site.Molecule.isrigid to set rigid bodies."
-                )
                 xyz.append(site.position())
                 types.append(
                     site.name if site.virtual_type is None else site.virtual_type.name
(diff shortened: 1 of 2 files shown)


chrisjonesBSU:
Still need to add some warning messages about re-ordering, and unit tests that would have caught this.

gmso/external/convert_hoomd.py
```diff
@@ -401,34 +401,60 @@ def _parse_particle_information(
...
-                charges.append(site.charge if site.charge else 0 * u.elementary_charge)
-                rigid_ids.append(
-                    site.molecule.number + n_prior_molecules
-                    if site.molecule.isrigid
-                    else -1
+            n_prior_molecules += site.molecule.number + 1
+    else:
+        for site in itertools.chain(top.sites, top.virtual_sites):
+            if isinstance(site, VirtualSite):
+                logger.warning(
```
CalCraven: I don't think this warning is necessary. There are cases where you could want a virtual site. You could use them as a ghost particle force in some non-equilibrium md simulation, for tracking forces, or some other cases.

CalCraven (approved):

PR #985, opened 2026-06-28
Test with Numpy 2.5, harden hoomd and lammps improper writers

chrisjonesBSU:
### PR Summary: Fix improper parameters in HOOMD and LAMMPS writers

GAFF impropers come through as length-1 unyt_arrays, and neither writer's improper parser reduced them to a scalar like the dihedral parsers do.

In the HOOMD writer, the parser passed the array straight into HOOMD, which runs it through float(). That used to work (just a deprecation warning), but NumPy 2.4 made float() on a non-scalar array a hard error. While digging into this I noticed the LAMMPS improper parser had the same gap, it doesn't type-check the parameter the way parse_charmm_style_dihedral does, so convert_kelvin_to_energy_units rejected the array outright (this one fails regardless of NumPy version).

Both fixes just pull the scalar out with .flatten()[0]. This only handles single-force-constant impropers, but that covers everything GMSO supports today as none of the improper templates use more than one k.

Added test_gaff_improper_ff for the LAMMPS path (fails without the fix); the existing test_gaff_sim covers HOOMD.

This also bumps up the tested version of numpy up to 2.5.

### PR Checklist
------------


CalCraven (changes requested):
So the <Value>float<Value/> syntax allows us to write improper types with different sets of values. If you want a forcefield that reads in as floats, you just put value="1.0" in the [parameter line](https://github.com/mosdef-hub/gmso/blob/a4b00f6aec3f9511bc6739724ae42a969854f5a0/gmso/utils/files/gmso_xmls/test_ffstyles/FENE-alkane.xml#L51)
 Here's one that we have in [our GAFF](https://github.com/mosdef-hub/gmso/blob/a4b00f6aec3f9511bc6739724ae42a969854f5a0/gmso/utils/files/gmso_xmls/test_ffstyles/gaff.xml#L38302), for a dihedral but the improper and dihedral share the same functional form.

This is also really common in [charmm](https://github.com/mosdef-hub/gmso/blob/main/gmso/tests/files/trimmed_charmm.xml) for periodic dihedrals and impropers.

This behavior needs to be supported, so we can't just flatten and take the first instance in the array. We can probably check the length of the array and process from there, or just earlier in the process make sure everything is stored in array format. That might be more clear, but potentially more memory intensive.

CalCraven:
See this test that is testing that it is read in correctly too.

https://github.com/mosdef-hub/gmso/blob/a4b00f6aec3f9511bc6739724ae42a969854f5a0/gmso/tests/test_forcefield.py#L244

chrisjonesBSU:
Good catch.  The last commit reworked the improper parsers to mirror the dihedral parsers, so all terms get captured and layered correctly.

gmso/formats/lammpsdata.py
```diff
@@ -1091,48 +1091,58 @@ def _write_impropertypes(out_file, top, base_unyts, parser, cfactorsDict):
...
-        parametersList.append(improper_type.parameters[k])
-    return parametersList, namesList
+    """Take an improper type and list parameters as expected in lammps outputs."""
+    kArray = improper_type.parameters["k"].flatten()
+    nArray = improper_type.parameters["n"].flatten()
+    phi_eqArray = improper_type.parameters["phi_eq"].flatten()
+    allParamsList = []
+    for a, b, c in zip(kArray, nArray, phi_eqArray):
+        allParamsList.append([a, b, c])
+    return allParamsList, ["k", "n", "phi_eq"]
```
CalCraven, line 1135: I think this zipping of the parameters happens [here](https://github.com/chrisjonesBSU/gmso/blob/c34728b1861ed4554cff0a802ef87edabced6998/gmso/formats/lammpsdata.py#L1113), so these changes aren't necessary. The code is definitely unclear, but I think we do want to return them as parametersList = [[k1,k1],[phi1,phi2]], for instance, instead of [[k1, phi1], [k2,phi2]] as you've done here. Also, do we need the flatten calls?

gmso/tests/test_lammps.py
```diff
@@ -584,6 +584,34 @@ def test_charmm_improper_ff(self):
...
+        found_impropers = False
+        for i, line in enumerate(lines):
+            if "Improper Coeffs" in line:
+                found_impropers = True
+                coeffs = lines[i + 2].split("#")[0].split()
+                assert coeffs[0] == "1"
+                # k, n, phi_eq must be written as plain scalars
+                assert len(coeffs) == 4
+                for value in coeffs[1:]:
+                    float(value)
```
CalCraven, line 612: What is this line doing? Maybe we should have a test that the parameters are applied correctly?
How is this different than this [improper test](https://github.com/chrisjonesBSU/gmso/blob/c34728b1861ed4554cff0a802ef87edabced6998/gmso/tests/test_lammps.py#L304)

CalCraven (changes requested):
If we want to get this PR merged before CECAM, we can also skip adding the lammps changes and just make sure we're working with numpy 2.5 and working in hoomd-blue, I'll leave that up to you. The lammps writer could maybe use some cleanup too just for clarity.

gmso/tests/test_lammps.py
chrisjonesBSU: It is ensuring the parameters are extracted as unyt quantities rather than arrays as `convert_kelvin_to_energy_units` expects unyt quantities. 

That improper test looks like it applies the forcefield through foyer, and creates the topology with from_parmed. As a result, all the improper parameters are stored as unyt quantities. Using apply() directly on a GMSO topology stores improper params as unyt arrays.

CalCraven: Gotcha. Maybe instead of just float(value), we could do 
```suggestion
                    assert isinstance(value, unyt.array.unyt_quantity)
```

gmso/formats/lammpsdata.py
CalCraven: I think actually this is done correctly. I would still verify that the params are exact in that benzene test, but otherwise I think we're good.

gmso/tests/test_lammps.py
chrisjonesBSU: So this test is occuring at a point where we already strip the value from the unyt quantity, so this assertion will fail. I think we can keep this where calling float() passes whether its a float or an int, or a str.

chrisjonesBSU:
The HOOMD specific fixes have been patched over to #948 as they are just related to behavior changes in numpy. I think we should still merge these changes in the lammps writer once the review is finished.

PR #990, opened 2026-07-16
GCPM Water forcefield usability with LAMMPS

CalCraven:
Note to run these lammps files, you'll need to install lammps using the gcpm plugin:
https://github.com/ndtrung81/lammps/tree/gcpm

### PR Checklist
------------

### DIFF
--- a/gmso/core/topology.py
+++ b/gmso/core/topology.py
@@ -225,7 +225,7 @@ class Topology(object):
     @combining_rule.setter
     def combining_rule(self, rule):
         """Set the combining rule for the topology."""
-        if rule not in ["lorentz", "geometric"]:
+        if rule not in ["lorentz", "geometric", "None"]:
             raise GMSOError("Combining rule must be `lorentz` or `geometric`")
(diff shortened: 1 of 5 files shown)


chrisjonesBSU:
Regarding the GitHub security bot comments, Could we simplify this test by importing one of the water models from mBuild, or are these relative positions and angle specific to GCPM? If so, lets add a GCPM model to [water.py](https://github.com/mosdef-hub/mbuild/blob/main/mbuild/lib/molecules/water.py) in mBuild.

CalCraven:
> Regarding the GitHub security bot comments, Could we simplify this test by importing one of the water models from mBuild, or are these relative positions and angle specific to GCPM? If so, lets add a GCPM model to [water.py](https://github.com/mosdef-hub/mbuild/blob/main/mbuild/lib/molecules/water.py) in mBuild.

It should be TIP4P geometry, I just grabbed the exact positions from a structure file to verify that it was being implemented correctly, but this is a good idea now that we know it's being handled correctly.

PR #997, opened 2026-08-20
Remove empty special_pairs objects in HOOMD forcefield and code CI update

CalCraven:
### PR Summary:
Removal and better testing of different special pairs handling in the GMSO Forcefield to HOOMD conversion. Any special pairs that are empty, i.e. (value is 1), so they're not treated specially, can be removed from the special pairs handling.

Also, fix many suggestions made during pre-commit from #995 

### PR Checklist
------------

### DIFF
--- a/gmso/external/convert_hoomd.py
+++ b/gmso/external/convert_hoomd.py
@@ -1197,13 +1197,16 @@ def _parse_coulombic(
     # Use same method as to_hoomd_snapshot to generate pairs list
     pairs_dict = generate_pairs_lists(top)
     for i, pair_type in enumerate(pairs_dict):
-        if scaling_factors[i] and pairs_dict[pair_type]:
+        if not scaling_factors[i] in (0, 1) and pairs_dict[pair_type]:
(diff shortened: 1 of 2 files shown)


CalCraven:
Can get an Error:
```
**ERROR**: No pairs are defined.
```
When performing sim.run, and this is the source. It's not actually causing any simulation issues, i.e. the simulation will run as intended, but will be nice to remove this message so users can better diagnose real errors with their setup.
```python
    for force in forces:
        if isinstance(force, hoomd.md.special_pair.LJ):
            if len(list(force.params.items())) is 0:
                remove=True
                break
    if remove:
        forces.remove(force)
    return forces
```

CalCraven:
This PR is also going to update many of the Linting issues noticed in the latest pre-commit PR.

chrisjonesBSU (approved):
LGTM. I'm not sure what is going on with the CI, but all of the tests are passing for me locally.

chrisjonesBSU (approved):

